PR #1, opened 2020-10-02
feat: Set up the django project

frgfm:
This PR aims at setting up the django project by:
- settings up a basic django project structure
- adding a first app "user"
- adding CI jobs for lint checking, deps installation, doc building, doc deployment, docker container tests

Here is how the documentation renders:
api_docs

Also, once the django server is run, it can be easily accessed using postman:
api_postman

As of now, I only implemented the project structure with a user app and its model but no views yet. I'll open up another PR to set the other apps (models).

Any feedback is welcome!


fe51:
Thank you for all this set up and first feature !

I think having multiple apps is a good point to keep the project structured ! 
However, in the long term, to keep things clear, we also could have a urls.py file per app. To get started, it suits me !

fe51 (approved):

frgfm:
Replaced by #2 using a FastAPI implementation

PR #5, opened 2020-10-13
feat: Add Device auth

florianriche:
### Description
The pull request adds a device authentification along with scopes. A device identifies itself directly instead of 

It is heavily based on FastAPI tutorial and FastAPI fullstack.

* Add a password field to the device creation
* Add Authorization route
* Add a security file
* Add a heartbeat mechanism to update the last_ping field of a device

### Description
This PR shall be amended when the user-auth PR is merged as there are multiple redundancies.
In addition, one should probably design a mechanism where a device has a user_id where this user_id type is "device", allowing it only to access devices routes.

### DIFF
--- a/src/app/api/routes/authorization.py
+++ b/src/app/api/routes/authorization.py
@@ -26,5 +26,5 @@ async def login_device(form_data: OAuth2PasswordRequestForm = Depends()):
     if not device:
         raise HTTPException(status_code=400, detail="Incorrect username or password")
 
-    access_token = create_unlimited_access_token(data=device.id)
+    access_token = create_unlimited_access_token(data={"sub": device.id, "scopes": form_data.scopes} )
     return {"access_token": access_token, "token_type": "bearer"}
--- a/src/app/api/routes/authorization.py
+++ b/src/app/api/routes/authorization.py
@@ -1,0 +1,30 @@
+from typing import List
+from fastapi import APIRouter, Path, Depends, HTTPException, status
+from app.api import routing, crud
+from app.db import devices
+from app.security import get_password_hash, verify_password, create_unlimited_access_token
+from app.api.schemas import Device, Token
+from fastapi.security import OAuth2PasswordRequestForm
+
+router = APIRouter()
+
+
+async def authenticate_device(db, username: str, password: str):
+    device = await crud.get(int(username), devices)
+    if not device:
+        return False
+    device = Device(**device)
+    if not verify_password(password, device.hashed_password):
+        return False
+    return device
+
+
+@router.post("/token", response_model=Token)
+async def login_device(form_data: OAuth2PasswordRequestForm = Depends()):
+    device = await authenticate_device(devices, form_data.username, form_data.password)
+
+    if not device:
+        raise HTTPException(status_code=400, detail="Incorrect username or password")
+
+    access_token = create_unlimited_access_token(data=device.id)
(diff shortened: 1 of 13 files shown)


src/app/api/routes/authorization.py
```diff
@@ -0,0 +1,30 @@
+from typing import List
```
frgfm, line 1: According to codacy, this import isn't used

src/app/api/routes/authorization.py
```diff
@@ -0,0 +1,30 @@
...
+from fastapi import APIRouter, Path, Depends, HTTPException, status
```
frgfm, line 2: Same here for `Path`

src/app/api/routes/authorization.py
```diff
@@ -0,0 +1,30 @@
...
+from app.api import routing, crud
```
frgfm, line 3: Same her for routing

src/app/api/routes/authorization.py
```diff
@@ -0,0 +1,30 @@
...
+from app.db import devices
+from app.security import get_password_hash, verify_password, create_unlimited_access_token
```
frgfm, line 5: here as well for get_password_hash

src/app/api/routes/devices.py
```diff
@@ -1,16 +1,19 @@
 from typing import List
-from fastapi import APIRouter, Path
+from fastapi import APIRouter, Path, Depends, HTTPException
```
frgfm, line 2: HTTPException isn't used

src/app/api/routes/devices.py
```diff
@@ -1,16 +1,19 @@
...
 from app.api import routing
 from app.db import devices
-from app.api.schemas import DeviceOut, DeviceIn
-
+from app.api.schemas import BaseDevice, Device, DeviceOut, DeviceIn, DeviceDBIn, HeartbeatOut, HeartbeatIn
```
frgfm, line 5: HeartbeatIn isn't used

src/app/config.py
```diff
@@ -7,3 +7,6 @@
 DEBUG: bool = os.environ.get('DEBUG', '') != 'False'
 DATABASE_URL: str = os.getenv("DATABASE_URL")
 LOGO_URL: str = "https://github.com/pyronear/PyroNear/raw/master/docs/source/_static/img/pyronear-logo-dark.png"
+ACCESS_TOKEN_EXPIRE_MINUTES = 30
+ACCESS_TOKEN_UNLIMITED_MINUTES = 60 * 24 * 365 * 10
+SECRET_KEY = "09d25e094faa6ca2556c818166b7a9563b93f7099f6f0f4caa6cf63b88e8d3e7"
```
frgfm, line 12: We can't expose the secret key here, perhaps retrieving it from the OS environment variables would be an option?

src/app/security.py
```diff
@@ -0,0 +1,39 @@
+from datetime import datetime, timedelta
+from typing import Any, Union, Dict
```
frgfm, line 2: Any isn't used

frgfm (changes requested):
Thanks for the much needed PR!
Could you add references in the PR description to the resources you used so that I can dive into it please?

Also some unused imports were flagged by Codacy, and we need to find a solution for the secret key and for the unittests!

PR #9, opened 2020-10-21
feat: Device route

florianriche:
Based on the user-auth PR
* Create a new object to store the Authorization and Identifcation logic. Object is called "Access"
* Create new Routes for this Objects mainly for debugging purposes
* Add a device scope
* Add a heartbeat route to the device
* Add a updatelocation route to the device
* Added the missing unit Tests
* remove "last"  prefix in from devices location fields.
* Add multiple conditions filter to fetch methods


.gitignore
```diff
@@ -134,4 +135,4 @@ venv.bak/
 dmypy.json
 
 # Pyre type checker
-.pyre/
+.pyre/
```
frgfm, line 138: EOF missing over there ;)

README.md
```diff
@@ -47,6 +50,13 @@ PORT=8002 docker-compose up -d --build
 Once completed, you will notice that you have a docker container running on the port you selected, which can process requests just like any django server.
 
 
+## Run tests on Docker
+Once the project has been built with docker using the previous command above. One can launch the unit tests using the following instruction:
+
+```bash
+PORT=8002 docker-compose exec web pyttest .
+```
+
```
frgfm, line 59: You think we should add this on top of the section in Contributing? (cf. https://github.com/pyronear/pyronear-api/blob/master/CONTRIBUTING.md#running-ci-verifications-locally)

florianriche: I added this to  the README before your refacto. I removed it in a recent commit.

src/app/api/routes/devices.py
```diff
@@ -39,3 +37,15 @@ async def delete_device(device_id: int = Path(..., gt=0), _=Security(get_current
...
 async def fetch_my_devices(me: UserRead = Security(get_current_user, scopes=["me"])):
     return await routing.fetch_entries(devices, ("owner_id", me.id))
+
+
+@router.post("/heartbeat", response_model=HeartbeatOut)
+async def heartbeat(device: DeviceOut = Security(get_current_device, scopes=["device"])):
+    return await crud.DeviceCRUD.heartbeat(device)
+
+
+@router.post("/{device_id}/update_location", response_model=DeviceOut)
```
frgfm, line 47: happy to go for "_" separators rather than "-" in route naming, but perhaps it would be best to keep the same convention all of them?

src/app/api/routing.py
```diff
@@ -4,12 +4,14 @@
 from fastapi import HTTPException, Path
 
 from app.api import crud, security
-from app.api.schemas import UserAuth, UserCreation
+from app.api.schemas import UserAuth, UserCreation, UserRead
+from app.api.schemas import AccessCreation, AccessRead, AccessAuth
+from app.api.schemas import DeviceAuth, DeviceCreation, DeviceOut
```
frgfm, line 9: I'd only suggest to switch to:
```
from app.api.schemas import (UserAuth, ...)
```

src/app/db.py
```diff
@@ -19,7 +19,15 @@
...
     Column("username", String(50)),
+    Column("access_id", Integer, ForeignKey("access.id"), unique=True),
     Column("created_at", DateTime, default=func.now()),
+)
+
+access = Table(
+    "access",
+    metadata,
+    Column("id", Integer, primary_key=True),
+    Column("username", String(50)),
```
frgfm, line 30: If I understand correctly, this can be a user or a device so perhaps it would be best to rename it to "login", what do you think?

src/app/api/routing.py
```diff
@@ -4,12 +4,14 @@
...
+from app.db import access as access_table
```
frgfm, line 10: I think we can avoid this import by passing the `access_table` argument of `create_access` down to `_create_access` :)

src/app/api/routes/login.py
```diff
@@ -4,21 +4,25 @@
...
 from app import config as cfg
 
 router = APIRouter()
 
 
 @router.post("/access-token", response_model=Token)
 async def login_access_token(form_data: OAuth2PasswordRequestForm = Depends()):
-
-    entry = await routing.fetch_entry(users, ('username', form_data.username))
+    entry = await routing.fetch_entry(access, ('username', form_data.username))
```
frgfm, line 15: cf. my other comment about calling the field "username"

src/app/api/crud.py
```diff
@@ -40,3 +45,31 @@ async def put(entry_id: int, payload: BaseModel, table: Table):
...
+        if not user_owns_device:
+            raise HTTPException(
+                status_code=400,
+                detail="You don't own this device."
+            )
+        device = (await get(device_id, devices))
+        device.update(payload.dict())
+        device = DeviceOut(**device)
+        await put(device.id, device, devices)
+        return device
```
frgfm, line 75: Before moving on a proper refactoring, I would have 2 suggestions :
1. Specific tables (i.e. "devices" here) might be better off being imported in a single file to make later refactoring easier (for devices, I think it's only imported /passed in `routes/devices.py`). Maybe the `device_table` could be passed here?
2. Since CRUD is supposed to be basic table operations, perhaps we could move these three methods into `routing.py` directly? (since we will have to split either CRUD or routing, let's put everything in routing for now for easier refactoring later. Routing already has table-specific functions)

What do you think?

frgfm (commented):
Thanks a lot for the PR, happy to have avoided diving into access management myself!
Looks good to me, I only have a few comments about refactoring/naming :ok_hand:

src/app/api/routing.py
florianriche: With a line for every entity, I find it a bit clearer what we import for each. 
If you really want to wrap everything into one import, I will change it

src/app/api/routes/devices.py
florianriche: Done

src/app/db.py
florianriche: Yes I was unsure about the naming. I was also afraid that login would be confusing with the "login" route but now that I think of it, we have "user" route and we don't confuse it with the entities.

florianriche: Done

src/app/api/routing.py
florianriche: _create_access is called by create_user and create_device as well and I'd rather not add the access_table as input of those functions in the user and devices routes

src/app/api/crud.py
fe51: 2. I agree. I might keep CRUD as it is with basic operations and move methods to routing (to refactor later) because it is already specifc.

src/app/api/routes/devices.py
```diff
@@ -39,3 +37,15 @@ async def delete_device(device_id: int = Path(..., gt=0), _=Security(get_current
 @router.get("/my-devices", response_model=List[DeviceOut])
```
fe51, line 37:
```suggestion
@router.get("/my_devices", response_model=List[DeviceOut])
```

src/app/api/crud.py
florianriche: I moved the code from CRUD to routing. 
It implied a bit refactor for fetch_one and fetch_all to add the possibility to have multiple_conditions.

src/app/api/routes/devices.py
florianriche: Following FG comment, I think it makes more sense to have hypens rather than underscores

src/app/api/routing.py
frgfm: I agree that having a line for each entity is good. But keeping that, we can simply do :
```
from app.api.schema import (UserAuth, UserCreation, UserRead,
                            AccessCreation, AccessRead, AccessAuth,
                            DeviceAuth, DeviceCreation, DeviceOut)
```
to avoid multiple imports :)

src/app/api/routing.py
frgfm: I was thinking it would actually be helpful because, from the routing code, you know straight away which tables will be accessed by any route. But I understand your point, happy to move forward with this :+1:

src/app/api/routing.py
```diff
@@ -20,11 +23,11 @@ async def get_entry(table: Table, entry_id: int = Path(..., gt=0)):
     return entry
 
 
-async def fetch_entries(table: Table, query_filter: Optional[Tuple[str, Any]] = None):
+async def fetch_entries(table: Table, query_filter: Optional[List[Tuple[str, Any]]] = None):
```
frgfm, line 26: Good thinking, I hesitated to suggest you to do this for your location update suggestion :)

src/app/api/routing.py
```diff
@@ -42,18 +45,34 @@ async def delete_entry(table: Table, entry_id: int = Path(..., gt=0)):
...
-    pwd = await security.hash_password(payload.password)
-    payload = UserCreation(username=payload.username, hashed_password=pwd, scopes=payload.scopes)
-    return await create_entry(user_table, payload)
+    pwd = await security.hash_password(password)
+    access = AccessCreation(login=login, hashed_password=pwd, scopes=scopes)
+    access_entry = AccessRead(** await create_entry(access_table, access))
+    return access_entry
+
+
+async def create_access(access_table: Table, payload: AccessAuth) -> AccessRead:
```
frgfm, line 62: If we are to avoid passing down the table, then I believe the `access_table` argument can be removed here since it's not being used

src/app/api/routing.py
```diff
@@ -65,3 +84,23 @@ async def update_user_pwd(user_table: Table, payload: UserCred, entry_id: int =
...
+    await update_entry(device_table, device, device.id)
+    return device
+
+
+async def update_location(device_table: Table, payload: UpdatedLocation, device_id: int, user_id: int):
+    user_owns_device = bool(await fetch_entry(device_table, [("id", device_id), ("owner_id", user_id)]))
+    if not user_owns_device:
+        raise HTTPException(
+            status_code=400,
+            detail="You don't own this device."
```
frgfm, line 100: Small suggestion to get a more common exception message: "Permission denied"

src/app/api/routes/login.py
```diff
@@ -4,21 +4,25 @@
...
 from app import config as cfg
 
 router = APIRouter()
 
 
 @router.post("/access-token", response_model=Token)
 async def login_access_token(form_data: OAuth2PasswordRequestForm = Depends()):
-
-    entry = await routing.fetch_entry(users, ('username', form_data.username))
+    entry = await routing.fetch_entry(access, [('login', form_data.username)])
```
jeanpasquier75, line 15: @florianriche your idea is to use Access as a unified authentication table for both users and devices ? At this moment it seams not working on my side...

src/app/api/deps.py
```diff
@@ -54,4 +56,23 @@ async def get_current_user(security_scopes: SecurityScopes, token: str = Depends
...
-    return UserRead(**entry)
+    return AccessRead(**entry)
+
+
+async def get_current_user(access=Depends(get_current_access)):
+    user = await crud.fetch_one(users, [('access_id', access.id)])
+
+    if user is None:
+        # Could be a "permission denied error as well"
+        raise HTTPException(status_code=400, detail="No existing user")
```
frgfm, line 67: "Non existing user" or something like "User entry not found"

src/app/api/deps.py
```diff
@@ -54,4 +56,23 @@ async def get_current_user(security_scopes: SecurityScopes, token: str = Depends
...
+
+    return UserRead(**user)
+
+
+async def get_current_device(access=Depends(get_current_access)):
+    device = await crud.fetch_one(devices, [('access_id', access.id)])
+    if device is None:
+        # Could be a "permission denied error as well"
+        raise HTTPException(status_code=400, detail="No existing device")
```
frgfm, line 76: same here

src/app/api/routing.py
```diff
@@ -1,15 +1,18 @@
...
+from typing import Optional, Tuple, Any, List
 from fastapi import HTTPException, Path
-
+from datetime import datetime
 from app.api import crud, security
-from app.api.schemas import UserAuth, UserCreation, UserCred, UserCredHash
+from app.api.schemas import UserAuth, UserCreation, UserRead, UserCredHash, UserCred
+from app.api.schemas import AccessCreation, AccessRead, AccessAuth
+from app.api.schemas import DeviceAuth, DeviceCreation, DeviceOut
+from app.api.schemas import HeartbeatOut, UpdatedLocation
```
frgfm:
```
from app.api.schemas import (UserAuth, UserCreation, UserRead, UserCredHash, UserCred,
                             AccessCreation, AccessRead, AccessAuth,
                             DeviceAuth, DeviceCreation, DeviceOut,
                             HeartbeatOut, UpdatedLocation)
```
would be better I think (keeping the entity separation idea)

src/app/api/routing.py
florianriche: Ok I updated it

src/app/main.py
```diff
@@ -3,7 +3,7 @@
 from fastapi.openapi.utils import get_openapi
 
 from app import config as cfg
-from app.api.routes import login, users, sites, events, devices, media, installations, alerts
+from app.api.routes import login, users, sites, events, devices, media, installations, alerts, access
```
frgfm, line 6: Quick (annoying) thoughts: all the routes have plural names, should we make is "accesses" for consistency?

src/app/db.py
```diff
@@ -19,7 +19,15 @@
     metadata,
     Column("id", Integer, primary_key=True),
     Column("username", String(50)),
+    Column("access_id", Integer, ForeignKey("access.id"), unique=True),
     Column("created_at", DateTime, default=func.now()),
+)
+
+access = Table(
```
frgfm, line 26: same here, should we make it "accesses" to keep plural naming?

src/app/api/routes/login.py
florianriche: It was working before I started some refacto and I haven't tested since.
I will check if it is working on my side. 
Maybe you didn't restart the DB and the schema is still the same as before?

frgfm (commented):
Thanks for all the edits! A few comments left, but most of it is fine by me!

src/app/main.py
florianriche: I thought about it but thought it was weird naming. But will fix you are right

src/app/api/routing.py
florianriche: Indeed but I wanted the create_****  method to have the same signature in the access route as in device and users routes  But we can get rid of it

src/app/api/routing.py
florianriche: That's cleaner and more versatile this way :) 
But we will need something else if we end up using "or" clauses.

florianriche:
I still need to adapt the password reset to use the new "Access" object

src/app/api/routes/login.py
florianriche: Hmm it's working on my side.

src/app/api/routes/access.py
```diff
@@ -0,0 +1,33 @@
+from typing import List
```
frgfm, line 1: While you're at it, would you mind renaming the file as well please? :pray:

frgfm (approved):
Thanks a lot for all the edits!
Very useful PR :ok_hand:  If you feel like it, perhaps rename routes/access.py to routes/accesses.py but it's really nitpicking

src/app/api/routes/access.py
florianriche: Done

frgfm (approved):
Thanks a lot!

PR #13, opened 2020-10-30
style: Fixed codacy check issues

frgfm:
This PR fixes most Codacy issues by:
- removing unused `__init__` from the test folder
- switching validators to static methods
- renaming `id` arguments to something more explicit
- removing the unused `ping` route


src/app/api/routes/ping.py
```diff
@@ -5,6 +5,4 @@
 
 @router.get("/ping")
```
florianriche, line 6: Shouldn't we remove the ping route altogether ? If so, I don't think it is necessary to open another PR just for that.

florianriche (approved):

src/app/api/routes/ping.py
frgfm: Agreed!

florianriche (approved):

PR #15, opened 2020-11-06
feat: Added a password update route

frgfm:
This PR adds the possibility of password update by:
- renaming the existing `/update-me` route to `/update-info` for coherence
- adding an admin and personal password update route. Both of them check if the entry exists, then hash the password and updates the entry in the table. For security reasons, instead of returning the input payload, I returned the user name to avoid exposing the password.
- adding corresponding unittests
- refactored schemas by adding a template class `_CreatedAt` and using multi-inheritance to avoid unneeded code duplication

I should bring your attention to route naming: I made the choice mentioned above but perhaps we should clarify the naming of all user routes since it's a complicated table.

Any feedback is welcome!

### DIFF
--- a/src/app/api/routing.py
+++ b/src/app/api/routing.py
@@ -4,7 +4,7 @@ from typing import Optional, Tuple, Any
 from fastapi import HTTPException, Path
 
 from app.api import crud, security
-from app.api.schemas import UserAuth, UserCreation
+from app.api.schemas import UserAuth, UserCreation, UserCred, UserCredHash
 
 
 async def create_entry(table: Table, payload: BaseModel):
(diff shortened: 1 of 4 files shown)


fe51:
Hi there, thanks for the PR. (you are so fast I was willing to try this issue :D)

I review it soon and come back to you.

src/app/api/routing.py
```diff
@@ -54,3 +54,14 @@ async def create_user(user_table: Table, payload: UserAuth):
     pwd = await security.hash_password(payload.password)
     payload = UserCreation(username=payload.username, hashed_password=pwd, scopes=payload.scopes)
     return await create_entry(user_table, payload)
+
+
+async def update_user_pwd(user_table: Table, payload: UserCred, entry_id: int = Path(..., gt=0)):
```
florianriche, line 59: One should create another file to store those methods that do the interface between routing and crud. In my PR I added some lines in this file too. If this continue, the file will end up quite big.

florianriche (approved):
OK for me.

fe51 (approved):

PR #16, opened 2020-11-07
Fix user authentication dependency

jeanpasquier75:
During dome refactoring (`id` argument replaced by `entry_id`) and get_current_user (deps.py) might have been skipped


frgfm (approved):
Thanks a lot for catching this!
Looks good to me :ok_hand: 
Strange that our current unittests didn't catch this, do you see a way we could add a test to avoid this in the future?

florianriche:
I have fixed the issue in the device PR (hence the conflict). I added entry_id as keyword argument label. I think we can close this PR.

frgfm:
@jeanpasquier75 can you confirm this was solved and that we may close the PR?

jeanpasquier75:
Yes it is now fixed in master, we can close this PR

PR #19, opened 2020-11-08
feat: Added default example values to schemas

fe51:
This little PR adds default example values to schemas by adding example parameters in classes from schemas.py.

I don't know whether or not to add examples for the fields ids, dates and token. What do you think? 

Also, I am wondering if we should add description to  field. 

image

Feedbacks are welcome

### DIFF
--- a/src/app/api/schemas.py
+++ b/src/app/api/schemas.py
@@ -163,7 +163,7 @@ class AlertIn(BaseModel):
     lat: float = Field(..., gt=-90, lt=90, example=48.654321)
     lon: float = Field(..., gt=-180, lt=180, example=2.654321)
     type: AlertType = AlertType.start
-    is_acknowledged: bool = False
+    is_acknowledged: bool = Field(False, example=False)
 
 
 class AlertOut(AlertIn, _CreatedAt):
--- a/src/app/api/schemas.py
+++ b/src/app/api/schemas.py
@@ -62,7 +62,7 @@ class AccessRead(AccessBase):
 
 
 class Token(BaseModel):
-    access_token: str = Field(..., example="eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxIiwic2NvcGVzIjpbImFkbWluIl0sImV4cCI6MTYwNDg0NTAyNH0.423fgFGTfttrvU6D1k7vF92hH5Flcd9LkvaJHCGFYd8E")
+    access_token: str = Field(..., example="eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.423fgFGTfttrvU6D1k7vF92hH5vaJHCGFYd8E")
     token_type: str = Field(..., example="bearer")
 
 
@@ -99,7 +99,7 @@ class DeviceIn(BaseModel):
     owner_id: int = Field(..., gt=0)
     specs: str = Field(..., min_length=3, max_length=100)
     elevation: float = Field(None, gt=0., lt=10000, example=1582)
-    lat: float = Field(None, gt=-90, lt=90, example = 44.123456)
+    lat: float = Field(None, gt=-90, lt=90, example=44.123456)
     lon: float = Field(None, gt=-180, lt=180, example=4.123456)
     yaw: float = Field(None, gt=-180, lt=180, example=110)
     pitch: float = Field(None, gt=-90, lt=90, example=-5)
@@ -111,7 +111,7 @@ class DeviceAuth(DeviceIn):
     scopes: Optional[str] = "device"
 
     class Config:
-        schema_extra = { "example": {
+        schema_extra = {"example": {
             "password": "PickARobustOne",
             "scopes": "device"}}
 
--- a/src/app/api/schemas.py
+++ b/src/app/api/schemas.py
@@ -163,7 +163,7 @@ class AlertIn(BaseModel):
     lat: float = Field(..., gt=-90, lt=90, example=48.654321)
     lon: float = Field(..., gt=-180, lt=180, example=2.654321)
     type: AlertType = AlertType.start
-    is_acknowledged: bool = Field(False, example=False)
+    is_acknowledged: bool = Field(False)
 
 
 class AlertOut(AlertIn, _CreatedAt):
--- a/src/app/api/schemas.py
+++ b/src/app/api/schemas.py
@@ -17,12 +17,12 @@ class _CreatedAt(BaseModel):
 
 # Abstract information about a user
 class UserInfo(BaseModel):
-    username: str = Field(..., min_length=3, max_length=50)
+    username: str = Field(..., min_length=3, max_length=50, example="JohnDoe")
 
 
 # Sensitive information about the user
 class Cred(BaseModel):
-    password: str
+    password: str = Field(..., example="PickARobustOne")
 
 
 class CredHash(BaseModel):
@@ -45,12 +45,12 @@ class UserCreation(UserInfo):
 
 
 class AccessBase(BaseModel):
-    login: str = Field(..., min_length=3, max_length=50)
-    scopes: str
+    login: str = Field(..., min_length=3, max_length=50, example="JohnDoe")
+    scopes: str = Field(..., example="me")
 
 
 class AccessAuth(AccessBase):
-    password: str
+    password: str = Field(..., example="PickARobustOne")
 
 
 class AccessCreation(AccessBase):
@@ -62,8 +62,8 @@ class AccessRead(AccessBase):
 
 
 class Token(BaseModel):
-    access_token: str
-    token_type: str
+    access_token: str = Field(..., example="eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxIiwic2NvcGVzIjpbImFkbWluIl0sImV4cCI6MTYwNDg0NTAyNH0.423fgFGTfttrvU6D1k7vF92hH5Flcd9LkvaJHCGFYd8E")
+    token_type: str = Field(..., example="bearer")
 
 
 class TokenPayload(BaseModel):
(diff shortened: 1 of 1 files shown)


src/app/api/schemas.py
```diff
@@ -45,12 +45,12 @@ class UserCreation(UserInfo):
```
florianriche, line 45: Add example to "scope" in UserAuth

src/app/api/schemas.py
```diff
@@ -95,21 +95,26 @@ class EventOut(EventIn, _CreatedAt):
...
+    pitch: float = Field(None, gt=-90, lt=90, example=-5)
     last_ping: datetime = None
 
 
 class DeviceAuth(DeviceIn):
     password: str
     scopes: Optional[str] = "device"
 
+    class Config:
+        schema_extra = {"example": {
```
florianriche, line 114: Add fields from DeviceIn

jeanpasquier75 (approved):
Great feature for documentation ! Except for the few fields forgotten as Florian said it looks good to me 👍

src/app/api/schemas.py
```diff
@@ -95,21 +95,26 @@ class EventOut(EventIn, _CreatedAt):
 
 
 class DeviceIn(BaseModel):
-    name: str = Field(..., min_length=3, max_length=50)
+    name: str = Field(..., min_length=3, max_length=50, example="pyronearEngine51")
     owner_id: int = Field(..., gt=0)
     specs: str = Field(..., min_length=3, max_length=100)
```
frgfm, line 100: Perhaps some examples on ̀owner_id`and `specs` while you're at it?

src/app/api/schemas.py
```diff
@@ -83,8 +83,8 @@ class SiteOut(SiteIn, _CreatedAt):
 
 
 class EventIn(BaseModel):
-    lat: float = Field(..., gt=-90, lt=90)
-    lon: float = Field(..., gt=-180, lt=180)
+    lat: float = Field(..., gt=-90, lt=90, example=44.765181)
+    lon: float = Field(..., gt=-180, lt=180, exempl=4.514880)
     type: EventType = EventType.wildfire
     end_ts: datetime = None
     start_ts: datetime = None
```
frgfm, line 90: missing examples over here

src/app/api/schemas.py
```diff
@@ -95,21 +95,26 @@ class EventOut(EventIn, _CreatedAt):
...
+    elevation: float = Field(None, gt=0., lt=10000, example=1582)
+    lat: float = Field(None, gt=-90, lt=90, example=44.123456)
+    lon: float = Field(None, gt=-180, lt=180, example=4.123456)
+    yaw: float = Field(None, gt=-180, lt=180, example=110)
+    pitch: float = Field(None, gt=-90, lt=90, example=-5)
     last_ping: datetime = None
 
 
 class DeviceAuth(DeviceIn):
     password: str
```
frgfm, line 110: could you add the validator and an example over here please?

src/app/api/schemas.py
```diff
@@ -95,21 +95,26 @@ class EventOut(EventIn, _CreatedAt):
...
     scopes: Optional[str] = "device"
 
+    class Config:
+        schema_extra = {"example": {
+            "password": "PickARobustOne",
+            "scopes": "device"}}
+
 
 class DeviceCreation(DeviceIn):
     access_id: int = Field(..., gt=0)
```
frgfm, line 120: missing example here

src/app/api/schemas.py
```diff
@@ -143,11 +148,11 @@ class MediaOut(MediaIn, _CreatedAt):
 class InstallationIn(BaseModel):
     device_id: int = Field(..., gt=0)
     site_id: int = Field(..., gt=0)
```
frgfm, line 150: over here too

src/app/api/schemas.py
```diff
@@ -143,11 +148,11 @@ class MediaOut(MediaIn, _CreatedAt):
...
-    lon: float = Field(..., gt=-180, lt=180)
-    yaw: float = Field(..., gt=-180, lt=180)
-    pitch: float = Field(..., gt=-90, lt=90)
+    elevation: float = Field(..., gt=0., lt=10000, example=1110)
+    lat: float = Field(..., gt=-90, lt=90, example=48.8534)
+    lon: float = Field(..., gt=-180, lt=180, example=2.3488)
+    yaw: float = Field(..., gt=-180, lt=180, example=42)
+    pitch: float = Field(..., gt=-90, lt=90, example=16)
     start_ts: datetime = None
     end_ts: datetime = None
```
frgfm, line 157: over there

src/app/api/schemas.py
```diff
@@ -160,8 +165,8 @@ class AlertIn(BaseModel):
     device_id: int = Field(..., gt=0)
     event_id: int = Field(..., gt=0)
     media_id: int = Field(None, gt=0)
-    lat: float = Field(..., gt=-90, lt=90)
-    lon: float = Field(..., gt=-180, lt=180)
+    lat: float = Field(..., gt=-90, lt=90, example=48.654321)
+    lon: float = Field(..., gt=-180, lt=180, example=2.654321)
     type: AlertType = AlertType.start
     is_acknowledged: bool = False
```
frgfm, line 171: over here

src/app/api/schemas.py
```diff
@@ -160,8 +165,8 @@ class AlertIn(BaseModel):
     device_id: int = Field(..., gt=0)
     event_id: int = Field(..., gt=0)
     media_id: int = Field(None, gt=0)
```
frgfm, line 167: here too

frgfm (commented):
Thanks a lot for the PR!
Would be even better if you could an example to some fields that are missing it, but it is fine by me as is!

src/app/api/schemas.py
fe51: thanks, I have indeed forgotten "specs" !

For all ids, I was wondering if it is usefull to generate some because, there is a defaut id 0 when doc is generated.  I will add some examples then

src/app/api/schemas.py
fe51: same as for ids, because dates are filled with now() in docs, I have not add some at first. But I can add some examples dates, will make it clearer !

frgfm: Oh you're right, ignore my previous comment then!

fe51:
I think we are good now.

src/app/api/schemas.py
```diff
@@ -36,7 +36,7 @@ class UserRead(UserInfo, _CreatedAt):
 
 # Authentication request
 class UserAuth(UserInfo, Cred):
-    scopes: Optional[str] = "me"
+    scopes: str = Field("me", example="me")
```
frgfm, line 39: I think the `Optional[str]`should be preserved here?

src/app/api/schemas.py
```diff
@@ -95,20 +95,20 @@ class EventOut(EventIn, _CreatedAt):
...
-    lon: float = Field(None, gt=-180, lt=180)
-    yaw: float = Field(None, gt=-180, lt=180)
-    pitch: float = Field(None, gt=-90, lt=90)
+    specs: str = Field(..., min_length=3, max_length=100, example="systemV0.1")
+    elevation: float = Field(None, gt=0., lt=10000, example=1582)
+    lat: float = Field(None, gt=-90, lt=90, example=44.123456)
+    lon: float = Field(None, gt=-180, lt=180, example=4.123456)
+    yaw: float = Field(None, gt=-180, lt=180, example=110)
+    pitch: float = Field(None, gt=-90, lt=90, example=-5)
     last_ping: datetime = None
```
frgfm, line 106: Should we make this fields as `Optional` using typing?

src/app/api/schemas.py
```diff
@@ -95,20 +95,20 @@ class EventOut(EventIn, _CreatedAt):
...
 
 
 class DeviceAuth(DeviceIn):
-    password: str
-    scopes: Optional[str] = "device"
+    password: str = Field(..., example="PickARobustOne")
+    scopes: str = Field("device", example="device")
```
frgfm, line 111: `Optional[str]` should be kept here, I think

src/app/api/schemas.py
```diff
@@ -120,11 +120,11 @@ class DeviceOut(DeviceIn, _CreatedAt):
...
-    elevation: float = Field(None, ge=0., lt=10000)
-    lat: float = Field(None, gt=-90, lt=90)
-    lon: float = Field(None, gt=-180, lt=180)
-    yaw: float = Field(None, gt=-180, lt=180)
-    pitch: float = Field(None, gt=-90, lt=90)
+    elevation: float = Field(None, ge=0., lt=10000, example=855)
+    lat: float = Field(None, gt=-90, lt=90, example=48.8534)
+    lon: float = Field(None, gt=-180, lt=180, example=2.3488)
+    yaw: float = Field(None, gt=-180, lt=180, example=10)
+    pitch: float = Field(None, gt=-90, lt=90, example=-7)
```
frgfm, line 127: same question here, perhaps typing those as `Optional` could be good

src/app/api/schemas.py
```diff
@@ -160,10 +160,10 @@ class AlertIn(BaseModel):
     device_id: int = Field(..., gt=0)
     event_id: int = Field(..., gt=0)
     media_id: int = Field(None, gt=0)
-    lat: float = Field(..., gt=-90, lt=90)
-    lon: float = Field(..., gt=-180, lt=180)
+    lat: float = Field(..., gt=-90, lt=90, example=48.654321)
+    lon: float = Field(..., gt=-180, lt=180, example=2.654321)
     type: AlertType = AlertType.start
-    is_acknowledged: bool = False
+    is_acknowledged: bool = Field(False, example=False)
```
frgfm, line 166: I think I misspoke earlier : since it's typed as bool, I think the example is not required :thinking:

frgfm (commented):
Sorry for the troubles, I still have some comments :sweat_smile:

src/app/api/schemas.py
fe51: As far as I have understood, adding elements instead of ... in field make it optionnal.
https://pydantic-docs.helpmanual.io/usage/schema/

I have tested (I can join screens later), but you can test by yourself inserting some user without scope.

src/app/api/schemas.py
fe51: hum, indeed, I will check without example and remove it if necessary

src/app/api/schemas.py
frgfm: Oh my bad then!

florianriche (approved):

PR #20, opened 2020-11-08
feat: Add media upload capability

florianriche:
### PR description:
This PR handles the logic of uploading media on the API.

First, some routes have been created for the devices to upload medias and alerts. Indeed, to be created, an alert and a media need a device_id but the devices are not supposed to know their own idea. As a consequence we have created routes to create medias and alerts without having to specify the device id.

To upload a media (the real media, not the representation in the DB), we have decided to rely on a cloud service provider bucket service. To achieve that goal, we have added a `bucket_key` field to the Media Table. That field will store the location of the real media in the cloud service Provider. To interact with the cloud service provider, we have implemented an abstract base class and a dummy service provider that does nothing for now but mimics the behavior of a real service.

Finally,  there was no route to link a media to an existing alert. We fix that by adding a new route.

### As a list:
1. `alert/from-device` when the device is sending an alert. No need to specify the device in it
1. `media/from-device` when the device is sending a media. No need to specify the device in it
1. Add a `bucket_key` field to the media schema.
1. Created a `BaseBucketService` abstract class that gives the signature of a class that interact with a bucket service.
1. Created a  `DummyBucketService` that inherits from `BaseBucketService` and just return successful codes.
1. Added a route `media/{media_id}/upload`  to upload a file on the selected cloud providers.
1. `alert/{alert_id}/link-media` to link a media to an alert once a media has been created and /or uploaded.
1. Unit tests


src/app/api/routes/alerts.py
```diff
@@ -13,6 +13,18 @@ async def create_alert(payload: AlertIn):
     return await crud.create_entry(alerts, payload)
 
 
+@router.post("/created-by-device", response_model=AlertOut, status_code=201,
```
frgfm, line 16: Shorter suggestion "/from-device"

src/app/api/routes/alerts.py
```diff
@@ -31,3 +43,18 @@ async def update_alert(payload: AlertIn, alert_id: int = Path(..., gt=0)):
...
 async def delete_alert(alert_id: int = Path(..., gt=0)):
     return await crud.delete_entry(alerts, alert_id)
+
+
+@router.put("/{alert_id}/link-media", response_model=AlertOut)
+async def link_media(payload: AlertMediaId,
+                     alert_id: int = Path(..., gt=0),
+                     current_device: DeviceOut = Security(get_current_device, scopes=["device"])):
+    # Check that alert is linked to this device
+    existing_alert = await crud.fetch_one(alerts, [("id", alert_id), ("device_id", current_device.id)])
```
frgfm, line 53: you might need to merge master with the recent change to query filters ;)

requirements.txt
```diff
@@ -5,6 +5,7 @@ SQLAlchemy>=1.3.12
 python-jose>=3.2.0
 passlib[bcrypt]>=1.7.4
 python-multipart==0.0.5
+boto3
```
frgfm, line 8: for now, we don't need it?

src/app/api/routes/media.py
```diff
@@ -1,9 +1,12 @@
-from typing import List
-from fastapi import APIRouter, Path
+
```
frgfm, line 1: First line can be removed

src/app/api/routes/media.py
```diff
@@ -1,9 +1,12 @@
...
+from fastapi import APIRouter, Path, Security, File, UploadFile, HTTPException
 from app.api import crud
 from app.db import media
-from app.api.schemas import MediaOut, MediaIn
-
+from typing import List
+from app.api.schemas import MediaOut, MediaIn, MediaCreation, DeviceOut, BaseMedia
+from app.api.deps import get_current_device
+from app.services import bucket_service
+from datetime import datetime
```
frgfm, line 9: Better to separate imports of external libraries vs. current package. Perhaps fastapi, typing & datetime at the top?

src/app/api/routes/media.py
```diff
@@ -16,7 +19,21 @@ async def create_media(payload: MediaIn):
     Below, click on "Schema" for more detailed information about arguments
     or "Example Value" to get a concrete idea of arguments
     """
-    return await crud.create_entry(media, payload)
+    bucket_key = hash(datetime.utcnow())
+    return await crud.create_entry(media, MediaCreation(**payload.dict(), bucket_key=bucket_key))
+
+
+@router.post("/created-by-device", response_model=MediaOut, status_code=201,
```
frgfm, line 26: Same here, "/from-device" might be more self explanatory & shorter

src/app/api/routes/media.py
```diff
@@ -49,3 +66,26 @@ async def delete_media(media_id: int = Path(..., gt=0)):
     Based on a media_id, deletes the given media
     """
     return await crud.delete_entry(media, media_id)
+
+
+@router.post("/{media_id}/upload_file", response_model=MediaOut, status_code=200)
```
frgfm, line 71: "/upload" is already a long route name I think :)

src/app/api/routes/media.py
```diff
@@ -49,3 +66,26 @@ async def delete_media(media_id: int = Path(..., gt=0)):
...
+async def upload_media(media_id: int = Path(..., gt=0),
+                       file: UploadFile = File(...),
+                       current_device: DeviceOut = Security(get_current_device, scopes=["device"])):
+    existing_media = await crud.fetch_one(media, [("id", media_id), ("device_id", current_device.id)])
```
frgfm, line 75: query filters to update afeter merging master here

src/app/api/routes/media.py
```diff
@@ -49,3 +66,26 @@ async def delete_media(media_id: int = Path(..., gt=0)):
...
+    if existing_media is None:
+        raise HTTPException(
+            status_code=400,
+            detail="Permission denied"
+        )
+
+    bucket_key = existing_media["bucket_key"]
+    upload_success = await bucket_service.upload_file(bucket_name="mypyroneartest",
```
frgfm, line 83: shoud we put the bucketname in the config?

src/app/services/__init__.py
```diff
@@ -0,0 +1 @@
+from .services import *
```
frgfm, line 1: we might want to specify the `__all__` attribute in ̀services.py` if we do this

src/app/services/services.py
```diff
@@ -0,0 +1,5 @@
+from app.services.bucket.s3 import S3Service
+from app.services.bucket.baseBucketService import BaseBucketService
+
+bucket_service = S3Service()
+assert isinstance(bucket_service, BaseBucketService), "selected service is not an instance of BaseBucketService"
```
frgfm, line 5: Is this for debug?

frgfm (commented):
Thanks for the PR!
A few questions in the comments, also perhaps it's best to split the PR: this one about updating the media key, and doing a sterilized upload. And another, adding bucket service options?

requirements.txt
florianriche: It's is used by the S3 demonstrator. I would keep it and remove it later hen we won't rely on s3 anymore.

src/app/api/routes/media.py
florianriche: good point. Done

src/app/services/services.py
florianriche: It is to make sure the selected bucket service inherits from ̀BaseBucketService`. 
As python is not typed; I don't know how to do it in another way. But if you have a suggestion, I would be glad to hear it, I might have overlooked a very simple solution.

florianriche:
Is there an easy way in Github to split the PR besides reverting some of the commits and creating a new PR afterwards?

src/app/services/services.py
frgfm: But I'm not sure we need to :sweat_smile: 
Since you create your `bucket_service` with a class that inherits from BaseBucketService, I'm not seeing why we need to ?
If someone does create the bucket_service another way, then they won't be able to call its methods properly. But that makes sense. That check would be good to put in our tests, but I don't see it as something to put in this module

florianriche: Haha indeed. I moved it to a unitTest

frgfm:
> Is there an easy way in Github to split the PR besides reverting some of the commits and creating a new PR afterwards?

Unfortunately, to the best of my knowledge, no :cry: 

But if you don't feel like it, perhaps you can edit the PR description and clean the rest so that there is a double goal here:
- add key field + upload related routes (sterilize them by only updating the key, not actual upload for now)
- add a basic bucket management module

src/app/services/bucket/dummyBucketService.py
```diff
@@ -0,0 +1,19 @@
+from app.services.bucket.baseBucketService import BaseBucketService
+
+
+class DummyBucketService(BaseBucketService):
+    def __init__(self):
+        """Dummy class used for testing purpose.
+        """
+
+    async def create_bucket(self, bucket_name):
+        return True
```
frgfm, line 10: Didn't pay attention to that one before, perhaps it would be safer to remove this method?
I mean we can do this on the CSP administration platform, and adding this could mean that our monthly bill could multiply if someone calls the wrong method :sweat_smile:

src/app/services/bucket/baseBucketService.py
```diff
@@ -0,0 +1,21 @@
+from abc import ABC
+from abc import abstractmethod
+
+
+class BaseBucketService(ABC):
+    """Abstract BucketService class to setup an interface regardless of the actual service."""
+
+    def __init__(self):
+        super().__init__()
```
frgfm, line 9: I think you don't need to override this considering you're using it exactly as the inherited constructor!

src/app/services/bucket/dummyBucketService.py
```diff
@@ -0,0 +1,16 @@
+from app.services.bucket.baseBucketService import BaseBucketService
+
+
+class DummyBucketService(BaseBucketService):
+    def __init__(self):
+        """Dummy class used for testing purpose.
+        """
```
frgfm, line 7: Same here, no need to specify the constructor

And also, for class docstring, it's better to put it before (outside) the constructor. Automatic documentation system won't catch it if it's placed in the constructor

frgfm (commented):
Just a few nitpicking comments (sorry :sweat_smile: ) and it's all good!

frgfm (approved):
Thanks a lot!

PR #23, opened 2020-11-11
refactor: Refactored schemas 

frgfm:
This PR refactors slightly the schemas to better leverage class inheritance.

### DIFF
--- a/src/app/api/routing.py
+++ b/src/app/api/routing.py
@@ -7,7 +7,7 @@ from app.api import crud, security
 from app.api.schemas import (UserAuth, UserCreation, UserRead, CredHash, Cred,
                              AccessCreation, AccessRead, AccessAuth,
                              DeviceAuth, DeviceCreation, DeviceOut,
-                             HeartbeatOut, UpdatedLocation)
+                             HeartbeatOut, DefaultLocation)
 from app.db import accesses as access_table
 
 
@@ -98,7 +98,7 @@ async def heartbeat(device_table: Table, device: DeviceOut) -> HeartbeatOut:
     return device
 
 
-async def update_location(device_table: Table, payload: UpdatedLocation, device_id: int, user_id: int):
+async def update_location(device_table: Table, payload: DefaultLocation, device_id: int, user_id: int):
     user_owns_device = bool(await fetch_entry(device_table, [("id", device_id), ("owner_id", user_id)]))
     if not user_owns_device:
         raise HTTPException(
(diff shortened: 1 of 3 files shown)


src/app/api/schemas.py
```diff
@@ -71,38 +77,62 @@ class TokenPayload(BaseModel):
...
+class _FlatLocation(BaseModel):
+    lat: float = Field(..., gt=-90, lt=90, example=44.765181)
+    lon: float = Field(..., gt=-180, lt=180, example=4.514880)
+
+
+class Location(_FlatLocation):
+    elevation: float = Field(..., gt=0., lt=10000, example=1582)
+
+
+class DefaultLocation(BaseModel):
```
florianriche, line 90: inherit  Location? What is the diff between Location and DefaultLocation?

src/app/api/schemas.py
```diff
@@ -71,38 +77,62 @@ class TokenPayload(BaseModel):
...
+
+
+class _Rotation(BaseModel):
+    yaw: float = Field(..., gt=-180, lt=180, example=110)
+    pitch: float = Field(..., gt=-90, lt=90, example=-5)
+
+
+class _DefaultRotation(BaseModel):
+    yaw: float = Field(None, gt=-180, lt=180, example=110)
+    pitch: float = Field(None, gt=-90, lt=90, example=-5)
```
florianriche, line 103: Same question, why create a Default objet?

src/app/api/routing.py
```diff
@@ -98,7 +98,7 @@ async def heartbeat(device_table: Table, device: DeviceOut) -> HeartbeatOut:
     return device
 
 
-async def update_location(device_table: Table, payload: UpdatedLocation, device_id: int, user_id: int):
+async def update_location(device_table: Table, payload: DefaultLocation, device_id: int, user_id: int):
```
florianriche, line 101: Should be DefaultPostion from your modification in routes/devices.py

src/app/api/schemas.py
frgfm: the default values for class attributes

src/app/api/routing.py
frgfm: good catch! Strange how it wasn't catch by unittests this time again :thinking:

florianriche: It's the same JSON schema so I would assume it shouldn't catch an exception unless we look for the exact values?

frgfm: Well DefaultLocation does not have `yaw` or `pitch` as fields, and since we pass the full payload in the unittest, I was expecting it to return a validator error. But I might be missing something here

florianriche (dismissed):

florianriche (approved):

Issue #26, opened 2020-11-12
[scripts] Add script to create super user


fe51:
## Feature
Add script to create super user

## Motivation
After a possible deletion of the volume of the db, it would be convenient to have a script to create a super user.

## Pitch
We might add a new folder called scripts with a first one, adding a super user with admin scope.

frgfm:
I don't think there will but make sure to check #17 to avoid PR duplication with @florianriche ;)
Having one script to create a super user will be useful even in production, and an end-to-end script is more for example and testing purposes

florianriche:
I don't know if a script is the right way to init a superuser.
In the cookie-cutter of fastapi, there is a piece of code that is executed at launch time to create this superuser. If the script exists, isn't there a risk (albeit small) that the superuser is created multiple times?

```python
def init_db(db: Session) -> None:
    # Tables should be created with Alembic migrations
    # But if you don't want to use migrations, create
    # the tables un-commenting the next line
    # Base.metadata.create_all(bind=engine)
    user = crud.user.get_by_email(db, email=settings.FIRST_SUPERUSER)
    if not user:
        user_in = schemas.UserCreate(
            email=settings.FIRST_SUPERUSER,
            password=settings.FIRST_SUPERUSER_PASSWORD,
            is_superuser=True,
        )
        user = crud.user.create(db, obj_in=user_in)  # noqa: F841```

Up to discuss

fe51:
Thanks for the feed back.
Indeed, the possibility to create at launch time might be a better way ! We would run init_db, similar function to the one in the cookie-cutter, just after database connection in main.py (and having the init_db func somewhere)? or, should we have a separate script for inits, a bit like in cookie cutter (https://github.com/tiangolo/full-stack-fastapi-postgresql/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/backend/app/app/initial_data.py) and run it in adding a command in the docker-compose file ? 


I agree with you, having a script could led to have multiples super user.

frgfm:
Indeed, if it's possible to do this at initialization, that would be perfect!
Having a separate python module like in the cookie cutter would be good option I think :ok_hand:

PR #27, opened 2020-11-12
feat: Add super user

fe51:
This PR create a script create a super user with ease.

Any feedback is welcome


src/scripts/addSuperUser.py
```diff
@@ -0,0 +1,45 @@
...
+connection = engine.connect()
+metadata = MetaData()
+metadata.reflect(engine)
+
+
+def createSuperUser(login, password):
+
+    # create hasched password
+    # I think we might get the pwd_context from security.py !
+    # problems with imports, thus temporary
```
florianriche, line 15: I think it would be better to call the code that creates a user directly rather than reproducing this process in the script.

src/scripts/addSuperUser.py
```diff
@@ -0,0 +1,45 @@
...
+    # insert user according to access_id
+    query = insert(metadata.tables['users']).values(username="superuser", access_id=access_id, created_at=func.now())
+    res = connection.execute(query)
+
+
+if __name__ == "__main__":
+    import argparse
+    parser = argparse.ArgumentParser(description='Add a super user in db with admin scopes')
+    parser.add_argument('-l', '--login', default="superuser", help="login, default:superuser")
+    parser.add_argument('-p', '--password', default="superuser", help="password, default:superuser")
```
florianriche, line 40: I'd rather not have default values to force the user to think of something.

src/scripts/addSuperUser.py
florianriche: (And i wonder if we need to create a script or just embed this super usercreation in the API deployment process. See my comment in  #26 )

florianriche (dismissed):

src/app/db/init_db.py
```diff
@@ -0,0 +1,24 @@
...
+
+from app.api import crud
+from app.db import accesses, users
+from app.api.schemas import AccessCreation, UserCreation
+from app.api.security import hash_password
+
+
+async def init_db():
+
+    login = os.getenv("SUPERUSER_LOGIN")
```
frgfm, line 11: Just moving this and `os.getenv("SUPERUSER_PWD")` to a config variable, and I think we are good to go :ok_hand:

fe51: done !

frgfm:
It's still labelled as "draft PR" @fe51, would you mind converting that? (we won't be able to merge it otherwise)

fe51:
> It's still labelled as "draft PR" @fe51, would you mind converting that? (we won't be able to merge it otherwise)

Ho sorry, now you can approve :) @frgfm @florianriche

frgfm (approved):
Thanks for the PR! I had some issues first but after restarting the DB, everything looks fine!

src/app/db/init_db.py
```diff
@@ -0,0 +1,23 @@
...
+
+    login = cfg.SUPERUSER_LOGIN
+
+    # check if access login does not already exist
+    entry = await crud.fetch_one(accesses, [("login", login)])
+    if entry is None:
+
+        hashed_password = await hash_password(cfg.SUPERUSER_PWD)
+
+        access = AccessCreation(login=login, hashed_password=hashed_password, scopes="admin")
```
frgfm, line 18: @florianriche quick question: in the swagger, I managed to use admin routes, but it fails on self ID routes. Do we need to add  `"me"` in the scopes?

florianriche: @frgfm I think the proper way would be to add `admin` to the accepted scopes list in the routes.

frgfm: Also, I used the `login/...` route to get my token, but then how can I use GET on `/users/` without receiving a "Not authenticated" in postman? (Swagger is fine, but I don't know if/how the token should be passed to each query)

florianriche: You retrieve the token by querying the login route and then you have to add it in your query headers.
You have a python example  in the PR I just created to deploy the device
https://github.com/pyronear/pyro-api/pull/43/files

```python
class Client():
    def __init__(self):
        self.token = ACCESS_TOKEN
        self.api = API_URL
        self.headers = {"Authorization": f"Bearer {ACCESS_TOKEN}"}

    def hearbeat(self):
        return requests.put(self.api + "/devices/heartbeat", headers=self.headers)
```

Issue #17, opened 2020-11-07
[test] Some bugs aren't caught up with our unittests


frgfm:
In #13 a small issue was introduced, that wasn't caught up by unittests.
Fortunately, @jeanpasquier75 noticed it and it's being solved in #16. But this raises the question about how we should design our unittests

Any thoughts @florianriche @jeanpasquier75 @fe51 ?

- [x] Update current unittests to get a testing situation as close as possible from production (#33)
- [x] Adding super-user creation mean (#26)

florianriche:
I would assume we could create a dedicated test Class to test the CRUD behavior as well as the fact that the DB constraints are respected: e.g by inspiring ourselves from https://xvrdm.github.io/2017/07/03/testing-flask-sqlalchemy-database-with-pytest/ or https://stackoverflow.com/questions/14719507/unit-tests-for-query-in-sqlalchemy 

In addition, a E2E test (user creation, user_auth, device creation ....) would catch some issues too.

frgfm:
This is now a priority:
I just found out that we have a typo in the device route https://github.com/pyronear/pyro-api/blob/master/src/app/api/routes/devices.py#L51, that should raise an error whenever we call the route.
This was not caught by the unittests because of our lazy monkeypatching. I'll dive into improving the unittests

@florianriche you might be the one with the best knowledge about authentication. Would you mind drafting consecutive route + payload to get a full end-to-end script?

frgfm:
Changing the release target to 0.1.1 for the last part (end-to-end script)

PR #33, opened 2020-11-14
test: Refactored codebase and updated route unittests to better reflect production cases

frgfm:
Hey there :wave: 

First & foremost, my apologies for the massive PR :sweat_smile: 
I'll go through a detailed explanation of what I did below and why:

1. **Adding unittests for security module**: all testable components of the app (especially those overridden during unittests) need to be tested. So I added pytest async to the requirements so that the hashing and password verification functions can be tested.

2. **Mock CRUD in pytest fixtures**: whatever the table, we need to do minimum patching during our testing to be able to closely test the API in real-case situations. I rewrote all CRUD functions to interact with mock tables (expressed as list of dicts).

3. **Designing each route's unittests**: using the previous fixtures, I carefully patch the CRUD functions that are used in the route, and replace the table by a deepcopy of my mock table. The deepcopy is necessary for each scenario to be run with exactly the same table. 

4. **Typing when possible**: as rewriting require some heavy lifting, it became necessary to know the exact input/output format of each function. So I added/corrected typing when it was missing/wrong (among others, in `crud`, `deps`, `security`).

5. **Table/Schemas changes**: adding minimal length to password, removed HeartbeatOut schema since it wasn't used (the typing of the route was corrected)

6. **Routing functions refactoring**: the access creation and password update function needing only the access table, I moved it to the `api/routes/accesses` module for clearer interface. I deleted the routing module and moved general table operations to crud. Each table-specific operation is implemented in the route module whose table is used. I know we discussed keeping code out of the route module with @florianriche in previous PRs, but this allows keeping table access to its specific route, making things easier for understanding and testing.

This PR was created mostly to improved unittests as per #17. Please bear in mind that I would have split this up in multiple PRs but to design new unittests, I needed some clearer module separation and typing, hence the 4, 5, and 6.
Any feedback is welcome!


src/app/api/routes/users.py
```diff
@@ -18,27 +20,31 @@ async def get_my_user(me: UserRead = Security(get_current_user, scopes=["me"])):
...
 async def update_my_info(payload: UserInfo, me: UserRead = Security(get_current_user, scopes=["me"])):
-    return await routing.update_entry(users, payload, me.id)
+    return await crud.update_entry(users, payload, me.id)
 
 
 @router.put("/update-pwd", response_model=UserInfo)
 async def update_my_password(payload: Cred, me: UserRead = Security(get_current_user, scopes=["me"])):
-    return await routing.update_pwd(users, payload, me.id)
+    entry = await crud.get_entry(users, me.id)
+    await update_access_pwd(payload, entry["access_id"])
```
florianriche, line 29: same comments as in the device route. I would find it cleaner to have separate dedicated functions called by the route. Same for create_user and upate_user_password

src/app/api/schemas.py
```diff
@@ -137,7 +137,7 @@ class DeviceIn(DefaultPosition):
 
 
 class DeviceAuth(DeviceIn):
```
florianriche, line 139: Maybe remove the password field and inherit from Cred?

src/app/api/security.py
```diff
@@ -10,12 +10,12 @@
 pwd_context = CryptContext(schemes=["bcrypt"], deprecated="auto")
 
 
-async def create_unlimited_access_token(content: dict) -> str:
+async def create_unlimited_access_token(content: Dict[str, Any]) -> str:
```
florianriche, line 13: Same comment as in the Crud file. Maybe use/create a dedicated type.

src/tests/conftest.py
```diff
@@ -1,23 +1,93 @@
...
+
+async def mock_hash_password(password):
+    return f"{password}_hashed"
+
+
+async def mock_verify_password(plain_password, hashed_password):
+    return hashed_password == f"{plain_password}_hashed"
+
+
+def pytest_configure():
```
florianriche, line 75: Is it a reserved function name that will be called before launching all the tests?

florianriche (commented):
Overall I agree with your refacto. The Unit Tests are much cleaner this way.

I added some comments. Some are optional. However, I really think that you should wrap the specific codes in routes (devices and users) in separated function.

src/app/api/routes/accesses.py
```diff
@@ -1,33 +1,59 @@
...
+    payload = CredHash(hashed_password=pwd)
+    await crud.update_entry(accesses, payload, entry_id)
+    # Return non-sensitive information
+    return {"login": entry["login"]}
+
+
 @router.post("/", response_model=AccessRead, status_code=201)
 async def create_access(payload: AccessAuth):
-    return await routing.create_access(payload)
+    return await post_access(**payload.dict())
```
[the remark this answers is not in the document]
frgfm: done!

src/app/api/schemas.py
frgfm: done!

src/tests/conftest.py
frgfm: Correct, cf. https://docs.pytest.org/en/stable/writing_plugins.html#registering-custom-markers

src/app/api/crud.py
```diff
@@ -1,35 +1,36 @@
...
 
 
-async def post(payload: BaseModel, table: Table):
+async def post(payload: BaseModel, table: Table) -> int:
     query = table.insert().values(**payload.dict())
     return await database.execute(query=query)
 
 
-async def get(entry_id: int, table: Table):
+async def get(entry_id: int, table: Table) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
frgfm: I would agree if we had complex types, but this is really basic, so I don't really see how it would help. With the current suggestion, each typing is self-explanatory but if I were to see `JsonDict` I'd have to trace back where it's defined to understand!

But as per your next comment, it would be better to discuss this in another issue :ok_hand:

src/app/api/routes/devices.py
```diff
@@ -1,60 +1,82 @@
...
 
+from app.api.routes.accesses import post_access, update_access_pwd
+
 router = APIRouter()
 
 
 @router.post("/", response_model=DeviceOut, status_code=201)
 async def create_device(payload: DeviceAuth, _=Security(get_current_user, scopes=["admin"])):
-    return await routing.create_device(devices, payload)
+    access_entry = await post_access(payload.name, payload.password, scopes=payload.scopes)
```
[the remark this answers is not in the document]
frgfm: I used to agree, but I was thinking about best practices in django and other habits. However, here fastapi was created, among others, to overcome the limitations of other frameworks. The use of the router decorator does not suggest that it has to be a one liner cf. https://fastapi.tiangolo.com/tutorial/bigger-applications/#another-module-with-apirouter

What I did in the access route is only because both functions are called by different routes. For devices, and users, I no longer see the point of creating another function just to wrap it with the decorator. I only this as a slight computation overhead for passing this down to another function.

Writing other functions, at least for me on this project, should be made to avoid duplication when this operation is done several times in the codebase. But I'm biased here, what are the reasons you think we should wrap this?

jeanpasquier75 (commented):
Do we want an accesses route for the API, i.e that users perform crud operations on accesses ?

In my opinion `accesses` is an internal technical table that contain in a single way our users & devices credentials and have a 1:1 relation with these tables (via `access_id`). I think we should not decouple one access to its user/device associated. On the contrary we should add consistency by using transactions in our database writings. If we agree on this I can work on it in a dedicated PR

I think we should keep functions `post_access` & `update_access_pwd` but place them in a shared module (maybe a submodule of crud ?) in order not to import from a route (users/devices) functions from another route

Looking forward to your feedbacks

frgfm:
@jeanpasquier75 not sure to understand fully your suggestion

Are you suggesting to remove the accesses routes and only keep the crud operations since it's something that should only be exposed to admins? But happy to discuss this in another issue or a PR if it's already clear to you!

Regarding `post_access` & `update_access_pwd`, I moved them over there because they require operations on the access_table. And this way, the access_table is only imported by its route and no other module. Would you pass the access_table as argument if we were to move both of these elsewhere?

florianriche:
I agree with @jeanpasquier75 that access routes should not exist. I added it mainly for debugging purpose. As of now I think we should leave it. If we want to remove it,  we should do so in another review.

jeanpasquier75:
> @jeanpasquier75 not sure to understand fully your suggestion
> 
> Are you suggesting to remove the accesses routes and only keep the crud operations since it's something that should only be exposed to admins? But happy to discuss this in another issue or a PR if it's already clear to you!
> 
> Regarding `post_access` & `update_access_pwd`, I moved them over there because they require operations on the access_table. And this way, the access_table is only imported by its route and no other module. Would you pass the access_table as argument if we were to move both of these elsewhere?

Yes I was suggesting to remove the route, but as Florian mentioned we can leave the read operations for debugging purpose but we definitely need to remove CUD operations to maintain a consistent database. 

I understand your refactoring and I agree on your choice but I suggest to move the 2 functions in a dedicated module, e.g crud/accesses.py

I will open a PR soon to be more concrete, including db transaction

frgfm:
@jeanpasquier75 @florianriche Alright so we deal with the accesses routes in another PR, right?

Regarding moving the 2 access-related functions, do you want to do with along with the deletion of the routes mentioned above? or in this PR?

frgfm:
ping @jeanpasquier75 @florianriche :grin:

florianriche:
I would do it in another PR to clearly separate the PR topics :)

jeanpasquier75 (approved):
I approve the routing -> crud refactor and tests _patch_session standardization !
I am working on the auth interaction with accesses, I will open the PR by tomorrow

PR #35, opened 2020-11-14
add min requirements

### LINKED ISSUE #34: [requirements] Adding some lower bound to all libraries' versions
Currently, the requirements file does have some loose version constraints on a few libraries, namely `python-jose` and `passlib`. Perhaps we should specify those a bit further?

florianriche:
Micro PR: setting q minimum version for some packages using my own personal requirements.


frgfm (approved):
Thanks for the quick PR!

PR #38, opened 2020-11-16
[Docs] [sites] Add summary and description to sites routes

fe51:
This PR aims at adding summary and description to the sites routes.

- On one side, I might suggest to not specify if field are required and the accepted range of value, not only because I might be lazy, but because it is specified in the schema, and it avoids the need to update the docs if the schema evolves. On the other side, it could helpful for api clients,  so what do you think ? 
- Concerning the site creation route, I have specified "click on schema" because, at first, I did not realise schema was clickable (see below). Is it only me ? :D 

image

Feedback are welcome


frgfm:
Thanks for the PR @fe51 !
Could you upload a screenshot of how it renders for the admin post route of sites please? I'm unsure whether we should specify each argument considering the schema is already available.

Also, any idea about the docker test failure?

frgfm:
@fe51 The docker issue should be fixed with #39!

fe51:
I will quickly upload screenshots !

"I'm unsure whether we should specify each argument considering the schema is already available." ---> I completely agree with you and most of the fields are explicit, thus, the description of argument not pertinent.

However, in some case, could be interesting to add information especially a unit (not pertinent here, but as an example, altitude in meters) --> Eventually, I just discover that I can add a description in schemas and it renders in the docs (see the screen below) --> I suggest to add a description when needed (so far, units)
image

fe51:
Here is the screenshot :
I have deleted arguments description. 

image



Also @frgfm , I have not noticed the docker test failure yesterday when I have created the PR, and I think it was not in failure, thanks to #39  its is okay now :)

frgfm (approved):
Looks good to me, thanks!

PR #37, opened 2020-11-15
Changing query filters from list of tuples to dict

martin1tab:
Issue28's description:

While designing new unittests, I figured I wasn't sure about what to expect when we use query_filters with multiple filters on a single value. Since this does not include order relationships but strictly equality, I suggest changing this to a dictionary to prevent duplication of keys.

Here is an example on the fetch_all crud function:

what a call currently looks like

result_entries = crud.fetch_all(user_table, [('username', 'fg')]
my suggestion

result_entries = crud.fetch_all(user_table, {'username': 'fg'}]

PR #39, opened 2020-11-17
fix: Fixed package versioning of databases

frgfm:
In #38, tests were failing apparently because of a missing `psycopg2` dependency. So I checked the last previous PR successfully building which is #37:

The build of 37 (https://github.com/pyronear/pyro-api/pull/37/checks?check_run_id=1408361256) installs `databases` v0.4.0 while the other (https://github.com/pyronear/pyro-api/pull/38/checks?check_run_id=1411729940) installs  version 0.4.1. I checked locally and adding an upper bound constraint fixes the issue.

Any feedback is welcome!

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 fastapi==0.61.1
 uvicorn>=0.11.1
-databases[postgresql]>=0.2.6
+databases[postgresql]>=0.2.6,<=0.4.0
 SQLAlchemy>=1.3.12
 python-jose>=3.2.0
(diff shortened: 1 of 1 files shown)


florianriche (approved):
Do you know what is the root cause? What behavior/signature change; raises this error?

fe51 (approved):

frgfm:
@florianriche not specifically, they seem to have change psycopg2 dependency in https://github.com/encode/databases/pull/256 :man_shrugging:

PR #40, opened 2020-11-17
feat: Added a self update route for a device's location

### LINKED ISSUE #32: [devices] Add a route to update the location of the current device
Currently, the API supports the update of location information for a given device provided admin scope. However, a device with GPS capabilities should be able to update its own location.

This might be a good occasion to add a route to update the location of the current_device!

frgfm:
This PR adds a route for a connected device to update its own location, which will close #32.

Any feedback is welcome!


src/app/api/routes/devices.py
```diff
@@ -71,6 +71,18 @@ async def update_device_location(
...
 
+@router.put("/my-location", response_model=DeviceOut)
+async def update_my_location(
+    payload: DefaultPosition,
+    device: DeviceOut = Security(get_current_device, scopes=["device"])
+):
+    # Update only the position
+    for k, v in payload.dict().items():
+        setattr(device, k, v)
+    await crud.update_entry(devices, device, device.id)
```
florianriche, line 82: I think you don't need to recreate a full object.
`await crud.update_entry(devices, payload, device.id)`  should work if the field names match.

florianriche (commented):
I think there is a one-liner solution that makes It clearer. Ok otherwise.

src/app/api/routes/devices.py
frgfm: I like that option too, but then how should we return the updated device without doing so?
(if I understand well, it's best to avoid querying the table again with a GET)

florianriche: I thought `update_entry` was returning the updated object but I am mistaken, it returns payload + entry. 
Shouldn't we alter it to do what I thought it was doing? After all, there is already a "get_entry" that is called in it so it is not a big change.

florianriche: That would also simplify the other update_location.

frgfm: But how would you modify it then?
In `update_entry`, the `get_entry` is used to verify that the entry exists. We should avoid doing a GET twice if possible.
So what you suggest in `update_entry` would be to do the `put` without verifying existence, and doing `get_entry` afterwards?

florianriche: No I meant to basically move what you implemented in this feature into update_entry.

```python
async def update_entry(table: Table, payload: BaseModel, entry_id: int = Path(..., gt=0)) -> Dict[str, Any]:
    entry = await get_entry(table, entry_id)
    await put(entry_id, payload, table)

    for k, v in payload.dict().items():
            setattr(entry, k, v)
     # Alternatively: 
     # updated_entry = entry.dict()
     # updated_entry.update(payload.dict())
    # return updated_entry
    return entry
```

frgfm: Oh I see! Well I just realized there was unnecessary overhead in the calls to `update_entry`(checking for existence while it was already done in the route), and pushed the commit. Let me know if you see a way to integrate this, happy to do it!

florianriche (approved):

PR #43, opened 2020-11-18
Client side deployment

### LINKED ISSUE #24: Discussion: Client-side API
We need to develop on Client-side (Device) in order to send information to the API.

The devices will need to: 
* Ping the API
* Send alerts
* Send medias
* Send positions (if using GPS )
Those messages will rely on a token that should be stored on the client-side.

Now comes the question: how to develop that client-side capability?
- Do we alter the CV component to have a main that sends messages to the API when it detect something. This could end up being a bit of spaghetti code and make the purpose of the package less clear
- Do we create a totally new component whose goal is to deploy on a device. This client would communicate with the API and call the CV component to look for fires. This would increase the number of repos.

florianriche:
Create a separate folder to deploy a Client service that will query the API.

The deployment is performed by launching a shell script. 

That script will
1. Ask for the API URL
2. Ask for the credentials of the device upon which we want to deploy
3. Retrieve the matching token.
4. If docker is not install; install it.
5. launch a docker-compose using the environment variables that have been retrieved in the deployment script.

The docker-compose will launch the client service that will send queries to the API as expected.

Remaining to do:

- [ ] Discuss and validate this architecture
- [ ] Retrieve the pyroNEar model
- [ ] Create the missing routes.


frgfm:
Thanks for opening this! I was wondering: wouldn't it be simpler to make the client a python package?
And then we choose whether we prefer to install it directly or in a docker on the raspberry?

florianriche:
> Thanks for opening this! I was wondering: wouldn't it be simpler to make the client a python package?
> And then we choose whether we prefer to install it directly or in a docker on the raspberry?

@frgfm  To be fair I have never packaged a python package. 
Do you mean to create a virtualenv and a setup.py file?

frgfm:
> @frgfm To be fair I have never packaged a python package.
> Do you mean to create a virtualenv and a setup.py file?

That's quite alright! So are my thoughts about the **general usecase**:
- since the device will already have a python environment to run the detection part, it would be good to let it install the client (and provide usage instructions)
- since the client is always related to the API, I believe this should be done in this repo in another folder just like you suggest in this PR
- as a package, it should have a function or an object's method to authenticate
- once it's done, locally, the device will regularly use a client function to update its ping + location
- when the detection is positive, it will use the client function to create an alert + upload a media

**And how we could implement it:**
- either OOP design: the package let the user create a  `Client` instance by passing its cred, which raise an error in case the API rejects the authentication. After this, you can use the instance method to call the API.
- or functional: a general function for authentication and any other function will check that the current session is authenticated
_I personally think the OOP design is much clearer_
Either way to package this, in the `client` folder, we need to create:
- a `requirements.txt`
- a `pyroclient` folder: place an `__init__.py` file, and a  ̀cllient.py` file (which should define the `Client` class + its methods similarly to your current suggestion)
- a `setup.py` --> I can do this in a follow-up PR

frgfm:
@florianriche actually, something that might help: I'm happy opening a PR to create an "empty" package with what I mentioned up there. And you can then merge this into yours and modify it if you want?

florianriche:
While we still have a bit of time I would like to do it myself for learning purpose, if you think we are in a hurry we can do as you suggest though.
What I will need help on in particular is how to use the pyronear-vision (DL the model and call it in the package)

frgfm:
@florianriche sure no hurry!

In that case, you might want to take a look at pyro-risks : https://github.com/pyronear/pyro-risks/blob/master/setup.py

However, I'm not following, what do you mean by importing pyro-vision? We're only making a python library for API client, not the vision part. Or did you have something else in mind?

florianriche:
Well the client needs to call the vision model to detect wether there is a wildfire and to get a media isn't it ?

frgfm:
> Thanks for the link!
> Well the client needs to call the vision model to detect wether there is a wildfire and to get a media isn't it ?

In your last sentence, if you replace "client" with "raspberry pi" then yes, but client-wise we're talking about two different things. Here we only create a client for the API as a library. 

The raspberry Pi has its own device handler that calls the vision part, and **if** the detection output is positive, then it will use our API client to communicate with the API. But that's two separate pieces of software/repos

florianriche:
Hmm Ok I think my vision was not complete. I expected this client to be solely for the device. However, you want to use that client also for the webapp?

frgfm:
Yes potentially, for any "client" of the API per se. A raspberry pi device handling is much larger than the API so this current repo has no business handling everything taking place over there, only helping the poor souls trying to use the API :)

client/setup.py
```diff
@@ -0,0 +1,74 @@
...
+        "Intended Audience :: Science/Research",
+        "License :: OSI Approved :: GNU General Public License v3 (GPLv3)",
+        "Natural Language :: English",
+        "Operating System :: OS Independent",
+        "Programming Language :: Python :: 3",
+        "Programming Language :: Python :: 3.6",
+        "Programming Language :: Python :: 3.7",
+        "Topic :: Scientific/Engineering",
+        "Topic :: Scientific/Engineering :: Mathematics",
+        "Topic :: Scientific/Engineering :: Artificial Intelligence",
```
frgfm, line 62: Perhaps we should change those two in our case?

client/setup.py
```diff
@@ -0,0 +1,74 @@
...
+        "Topic :: Software Development",
+        "Topic :: Software Development :: Libraries",
+        "Topic :: Software Development :: Libraries :: Python Modules",
+    ],
+    keywords=["data science", "time series", "machine learning", "client", "api"],
```
frgfm, line 67: same here as we don't do any data science in the repo, here is a suggestion:
backend, api, client, wildfire

client/pyroclient/client.py
```diff
@@ -0,0 +1,65 @@
+import requests
+import logging
+from exceptions import HTTPRequestException
+from urllib.parse import urljoin
+
+logging.basicConfig()
+
+
+class Client():
```
frgfm, line 9: no need for the `()`

client/pyroclient/client.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def create_media(self):
+        return requests.post(self.routes["create-media"], headers=self.headers)
+
+    def upload_media(self):
+        return requests.post(self.routes["upload-media"], headers=self.headers)
+
+
+if __name__ == "__main__":
+    client = Client("http://localhost:8002", "superuser", "superuser")
```
frgfm, line 65: If we use the folder as a package, I think we should remove this

frgfm (changes requested):
Thanks a lot for the quick PR! I made a few comments, let me know what you think!

client/pyroclient/client.py
florianriche: Indeed, it's just for development purpose for myself. I will leave it while the PR is not yet ready to be merged.

client/pyroclient/client.py
```diff
@@ -0,0 +1,65 @@
...
+            return response.json()["access_token"]
+        else:
+            # Anyone has a better suggestion?
+            raise HTTPRequestException(response.status_code, response.text)
+
+    def hearbeat(self):
+        return requests.put(self.routes["heartbeat"], headers=self.headers)
+
+    def update_my_location(self):
+        return requests.put(self.routes["uppdate-my-location"], headers=self.headers)
```
frgfm, line 52: some typo over here

client/pyroclient/client.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def create_alert(self):
+        return requests.post(self.routes["create_alert"], headers=self.headers)
```
frgfm, line 55: "send-alert" ?

frgfm (changes requested):
Just a few things left and we'll be ready for merging!

fe51 (approved):

client/pyroclient/client.py
florianriche: What typo?

florianriche:
Is it ready for merge with the current function? We may add/fill them later

client/pyroclient/client.py
frgfm: "uppdate-my-location" --> "update-my-location" according to your `self.routes` keys' names

frgfm (approved):
Looks good thanks!

frgfm:
@florianriche fyi, once you merge the PR, you can delete the feature branch ;) (I just did, but for the next ones)

florianriche:
@frgfm good point thank you :)

PR #46, opened 2020-11-19
Access (auth) refactor & database transactions

jeanpasquier75:
Changes: 
* Move accesses functions from routes/accesses.py to crud/accesses.py
* Use database transaction for atomicity for user/device creation

1st point lead to split crud module into submodules base (legacy) & accesses (new). I would need some feedback on this part. I needed to change patching during unit tests, fetch_all does not seem to work.


src/app/api/routes/users.py
```diff
@@ -26,14 +24,15 @@ async def update_my_info(payload: UserInfo, me: UserRead = Security(get_current_
 @router.put("/update-pwd", response_model=UserInfo)
 async def update_my_password(payload: Cred, me: UserRead = Security(get_current_user, scopes=["me"])):
     entry = await crud.get_entry(users, me.id)
-    await update_access_pwd(payload, entry["access_id"])
+    await crud.update_access_pwd(payload, entry["access_id"])
```
frgfm, line 27: You forgot to do the same for the admin version further down below ;)

src/app/api/routes/users.py
```diff
@@ -26,14 +24,15 @@ async def update_my_info(payload: UserInfo, me: UserRead = Security(get_current_
...
     return entry
 
 
 @router.post("/", response_model=UserRead, status_code=201)
+@database.transaction()
```
frgfm, line 32: I'm curious, what is the decorator doing?

src/app/db/tables.py
```diff
@@ -28,7 +28,7 @@
     "accesses",
     metadata,
     Column("id", Integer, primary_key=True),
-    Column("login", String(50)),
+    Column("login", String(50), index=True, unique=True),  # index for fast lookup
```
frgfm, line 31: What is this doing? (sorry I might have not gone through the whole documentation :sweat_smile: )

src/app/db/tables.py
```diff
@@ -75,7 +75,7 @@ class EventType(str, enum.Enum):
     Column("id", Integer, primary_key=True),
     Column("login", String(50)),
     Column("owner_id", Integer, ForeignKey("users.id")),
-    Column("access_id", Integer, ForeignKey("accesses.id"), unique=True),
+    Column("access_id", Integer, ForeignKey("accesses.id", ondelete="CASCADE"), unique=True),
```
frgfm, line 78: So deleting the device will delete the access? If so, shouldn't we do the same for the user?

src/tests/test_accesses.py
```diff
@@ -16,12 +15,9 @@ def _patch_session(monkeypatch, mock_table):
...
     # Sterilize all DB interactions through CRUD override
-    monkeypatch.setattr(crud, "get", pytest.mock_get)
-    monkeypatch.setattr(crud, "fetch_all", pytest.mock_fetch_all)
-    monkeypatch.setattr(crud, "fetch_one", pytest.mock_fetch_one)
-    monkeypatch.setattr(crud, "post", pytest.mock_post)
-    monkeypatch.setattr(crud, "put", pytest.mock_put)
-    monkeypatch.setattr(crud, "delete", pytest.mock_delete)
+    monkeypatch.setattr(base, "get", pytest.mock_get)
+    monkeypatch.setattr(base, "fetch_all", pytest.mock_fetch_all)
+    monkeypatch.setattr(base, "fetch_one", pytest.mock_fetch_one)
```
frgfm, line 20: Since you made the crud __init__, you can keep the previous version (you have to patch the thin imported in the route, and does not import base, but crud)

src/tests/test_accesses.py
```diff
@@ -62,100 +58,3 @@ def test_fetch_accesses(test_app, monkeypatch):
     assert response.status_code == 200
     assert response.json() == [{k: v for k, v in entry.items() if k != "hashed_password"}
                                for entry in mock_access_table]
```
frgfm, line 60: If we remove the unittests for the routes, we need to adapt them to test the crud.accesses functions. I don't think it would be wise to decrease coverage by that much

src/tests/test_alerts.py
```diff
@@ -21,11 +21,11 @@ def _patch_session(monkeypatch, mock_table):
...
-    monkeypatch.setattr(crud, "get", pytest.mock_get)
-    monkeypatch.setattr(crud, "fetch_all", pytest.mock_fetch_all)
-    monkeypatch.setattr(crud, "post", pytest.mock_post)
-    monkeypatch.setattr(crud, "put", pytest.mock_put)
-    monkeypatch.setattr(crud, "delete", pytest.mock_delete)
+    monkeypatch.setattr(base, "get", pytest.mock_get)
+    monkeypatch.setattr(base, "fetch_all", pytest.mock_fetch_all)
+    monkeypatch.setattr(base, "post", pytest.mock_post)
+    monkeypatch.setattr(base, "put", pytest.mock_put)
+    monkeypatch.setattr(base, "delete", pytest.mock_delete)
```
frgfm, line 28: Same here

frgfm (changes requested):
Thanks for the PR! You can revert the changes on imports and crud patching in the unittests, that should solve most of the unittest issues!

src/app/api/routes/users.py
jeanpasquier75: It is a way to execute queries within a transaction, see: [databases documentation](https://www.encode.io/databases/connections_and_transactions/)

src/app/db/tables.py
jeanpasquier75: An index allows the database to fetch rows faster when filter clause contains the index column -> accesses table is always queried using the login. See: [postgresql documentation](https://www.postgresql.org/docs/12/indexes-intro.html)

src/app/api/routes/devices.py
```diff
@@ -3,17 +3,20 @@
...
 
-from app.api.routes.accesses import post_access, update_access_pwd
+from app.api.crud.accesses import post_access, update_access_pwd
+
 
 router = APIRouter()
 
 
 @router.post("/", response_model=DeviceOut, status_code=201)
+@database.transaction()
```
florianriche, line 17: Why add this decorator only for route that create objects?

src/app/api/routes/accesses.py
```diff
@@ -1,44 +1,15 @@
...
-    # Return non-sensitive information
-    return {"login": entry["login"]}
-
-
-@router.post("/", response_model=AccessRead, status_code=201)
-async def create_access(payload: AccessAuth):
-    return await post_access(payload.login, payload.password, payload.scopes)
-
-
 @router.get("/{access_id}/", response_model=AccessRead)
```
florianriche, line 13: If we want to remove the access routes (it is fine by me if we think we don't need it anymore for debugging purpose), why leave those routes?

src/app/db/tables.py
florianriche: And shall we do it also for the owner_id ?  If we delete the users, the devices will disappear?

src/app/api/routes/accesses.py
frgfm: I think he mentioned that we don't want to expose creation/modification of accesses but reading is fine

src/app/db/tables.py
jeanpasquier75: Yes I thought that I also had added the on delete cascade on user but I forgot.

I figured out that it is the opposite: deleting the access deletes the device/user. I am working on fixing the device/user deletion. 

I will add the mechanism owner id as well !

src/tests/test_accesses.py
jeanpasquier75: Actually it is not that simple (I think, I am not masting import mechanisms) because it seams that mocking crud.get will keep untouched the original crud.base.get function and if it imported afterward the original is used instead of the mocked version. So far the only way I could mock all crud function was to mock both crud._ and crud.base._ (_ beeing each function). Do you see ?

src/app/api/routes/devices.py
jeanpasquier75: I planned to add the decorator when transaction is needed (writing in several tables with atomicity), user/device creation was the first use case, did I miss other ones ?

I realised that route decorator was not the best practice: it makes it impossible to test. We may need to find a way to mock a database transaction or we'll have to rewrite operations into pytest mocks. @frgfm do you see what I mean ?

src/app/api/routes/accesses.py
jeanpasquier75: Yes reading is fine while it does not break database consistency

src/app/db/tables.py
jeanpasquier75: Another issue when we use the on cascade mechanism -> as it perform at db level we have to find a way to mock it in our unit tests

src/app/api/routes/devices.py
frgfm: I am no where close to be knowledgeable about database transactions since I just discovered that thanks to you @jeanpasquier75. What is the expected behaviour difference after decorating the route? reducing execution time? output type change?

If it's beneficial to the API, even if we can't test it per se, I would argue that we should still put it. If it doesn't bring much and we can't test it, then perhaps let's leave it out

src/tests/test_accesses.py
frgfm: Mmmh, what error did you get with my suggestion?

Because the import for the accesses is over here: https://github.com/pyronear/pyro-api/blob/master/src/app/api/routes/accesses.py#L3

So if we patch this very import, it should be working without any problem :thinking:

frgfm:
@jeanpasquier75 any update? I'm changing the release target as this isn't critical for the incoming demo

frgfm:
So @jeanpasquier75, what should we do with this PR?
Considering the (very favorable) pace of PRs/modif of the repo, it's better to process PR rather quickly :sweat_smile:

jeanpasquier75:
> So @jeanpasquier75, what should we do with this PR?
> Considering the (very favorable) pace of PRs/modif of the repo, it's better to process PR rather quickly 😅

Sorry for the delay 😅 I will dry to come up with something by the end of the week. As you mentioned in #76 this PR was supposed to fix access deletion with the SQL cascade mechanism as well

frgfm:
@jeanpasquier75 #76 was merged, let us know if you think we should keep this PR for th upcoming release or staging it for the next one!

jeanpasquier75:
Closing this outdated PR resolved in #90

PR #47, opened 2020-11-21
fix: Fixed DB initialization

frgfm:
Since query filters were recently changed in #37, the db initialization was still using the legacy interface. This PR fixes it

_To prevent this type of issue from happening again, perhaps we should have a github action checking the status of the Heroku deployment upon build completion (pinging the documentation for instance?). I noticed this after trying to access the documentation, and I looked at the logs to find out what was causing the crash_

### DIFF
--- a/src/app/db/init_db.py
+++ b/src/app/db/init_db.py
@@ -10,7 +10,7 @@ async def init_db():
     login = cfg.SUPERUSER_LOGIN
 
     # check if access login does not already exist
-    entry = await crud.fetch_one(accesses, [("login", login)])
+    entry = await crud.fetch_one(accesses, {"login": login})
     if entry is None:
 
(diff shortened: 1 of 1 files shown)


fe51 (approved):

fe51:
Hi, thanks for this PR.

(that could explain I could not reach docs from http://pyronear-api.herokuapp.com/redoc)

That is a good idea, I agree with you, pinging the docs would be okay. However, is heroku deployment linked to our PR merged ? Sorry I am not knowledgeable about that part.

jeanpasquier75 (approved):
Maybe a static type checker (mypy, pytype, etc.) could catch this type of error ?

florianriche (approved):

frgfm:
Good idea @jeanpasquier75 !
I tried on my end and strangely, mypy didn't throw any error. But let's investigate that option further

PR #51, opened 2020-11-22
[init][config][bug] Add a warning/error to verify that DB is not init with None credentials

fe51:
This PR aims at raising an error if Credentials are missing from environment variables.

Now, it raises an ValueError if SUPERUSER_PWD/SUPERUSER_LOGIN is None.

It will improve debugging.

Any feedback is welcome !

### DIFF
--- a/src/app/config.py
+++ b/src/app/config.py
@@ -22,4 +22,9 @@ ACCESS_TOKEN_UNLIMITED_MINUTES = 60 * 24 * 365 * 10
 JWT_ENCODING_ALGORITHM = "HS256"
 
 SUPERUSER_LOGIN: str = os.getenv("SUPERUSER_LOGIN")
+if SUPERUSER_LOGIN is None:
+    raise ValueError("SUPERUSER_LOGIN is missing from environment variables")
+
 SUPERUSER_PWD: str = os.getenv("SUPERUSER_PWD")
+if SUPERUSER_PWD is None:
+    raise ValueError("SUPERUSER_PWD is missing from environment variables")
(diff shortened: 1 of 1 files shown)


src/app/config.py
```diff
@@ -22,4 +22,9 @@
 JWT_ENCODING_ALGORITHM = "HS256"
 
 SUPERUSER_LOGIN: str = os.getenv("SUPERUSER_LOGIN")
+if SUPERUSER_LOGIN is None:
+    raise ValueError("SUPERUSER_LOGIN is missing from environment variables")
+
 SUPERUSER_PWD: str = os.getenv("SUPERUSER_PWD")
+if SUPERUSER_PWD is None:
+    raise ValueError("SUPERUSER_PWD is missing from environment variables")
```
frgfm, line 30: Since both of them are to be specified, perhaps we could use the minimal
`if SUPERUSER_LOGIN is None or SUPERUSER_PWD is None:`

And about the error, something more suggestive of how to solve this perhaps? Like "Please set 'SUPERUSER_LOGIN' and 'SUPERUSER_PWD' in your environment variables.

frgfm (commented):
Thanks for the PR! Strange that the CI didn't finish, could you merge master please?

src/app/config.py
fe51: Yes I was hesitating with the or statement. 

Okay !

fe51:
> Thanks for the PR! Strange that the CI didn't finish, could you merge master please?

hum weird, because I seem up to date !

src/app/config.py
```diff
@@ -23,3 +23,7 @@
 
 SUPERUSER_LOGIN: str = os.getenv("SUPERUSER_LOGIN")
 SUPERUSER_PWD: str = os.getenv("SUPERUSER_PWD")
```
florianriche, line 25: in which case could we not have the credentials ?

florianriche (approved):

src/app/config.py
frgfm: When the environment variable was not set, just like it happened earlier today on Heroku :upside_down_face:

frgfm (approved):
Looks good to me!

frgfm:
@fe51 don't forget to delete the branch afterwards :sweat_smile:

PR #56, opened 2020-11-23
feat: Adding a route to list of devices on a site at a given time 

### LINKED ISSUE #49: [installations] Adding a route to list of devices on a site at a given time
As per https://github.com/pyronear/pyro-platform/issues/19#issuecomment-731733564, we need to able to tell how many devices (list them would avoid creating another route later) are on a specific site at a given time.

Addind a route `/list-devices` in installations would be the best option in my opinion

martin1tab:
This PR adds a route to list all devices on an installation that is active at a given timestamp and located at a specific site.

Successfully merging this will close #49


fe51:
Hi there, thanks for the PR !
Could you state is as draft as long as the PR is not ready for reveiw ?

martin1tab:
> Hi there, thanks for the PR !
> Could you state is as draft as long as the PR is not ready for reveiw ?

Hi, yes sure!!

martin1tab:
> > Hi there, thanks for the PR !
> > Could you state is as draft as long as the PR is not ready for reveiw ?
> 
> Hi, yes sure!!

It is ready now

src/app/api/routes/installations.py
```diff
@@ -31,3 +32,13 @@ async def update_installation(payload: InstallationIn, installation_id: int = Pa
 @router.delete("/{installation_id}/", response_model=InstallationOut)
 async def delete_installation(installation_id: int = Path(..., gt=0)):
     return await crud.delete_entry(installations, installation_id)
+
+
+@router.get("/list_devices/{site_id}", response_model=List[int])
```
frgfm, line 37: For consistency, I'd recommend changing "list_devices" to "list-devices", and moving the site_id before "list-devices" 

And are we positive that using typing as response model is supported? (not schemas I mean) Have you tried the feature locally?

src/app/api/routes/installations.py
```diff
@@ -31,3 +32,13 @@ async def update_installation(payload: InstallationIn, installation_id: int = Pa
...
+async def get_all_at_given_ts_and_site(timestamp: datetime, site_id: int = Path(..., gt=0)):
+    query = installations.select(installations.c.device_id)
+    query = query.where(installations.c.site_id == site_id)
+    query = query.where(installations.c.start_ts <= timestamp)
+    query = query.where(installations.c.end_ts >= timestamp)
```
frgfm, line 42: Two things:
- rather than doing multiple where, we should use `and_` (cf. https://docs.sqlalchemy.org/en/13/core/sqlelement.html#sqlalchemy.sql.expression.and_ I'll open a PR to modify this in the rest of the codebase)
- we have to be careful, in the installation table so far, start_ts and end_ts can be None. I'm not entirely sure the query will work if that's the case?

src/app/api/routes/installations.py
```diff
@@ -31,3 +32,13 @@ async def update_installation(payload: InstallationIn, installation_id: int = Pa
...
+
+    return await crud.get_custom_entry(installations)
```
frgfm, line 44: I'm trying to think about a way to adapt the query filters so that we don't need that extra crud. Perhaps the others will have some insights

src/tests/test_installations.py
```diff
@@ -15,6 +15,11 @@
 ]
 
 
+async def mock_get_custom_entry(query):
+
+    return [INSTALLATION_TABLE[0]["device_id"]]
+
+
```
frgfm, line 22: That's one problem with this crud, we can't mock it properly. This override is completely static, which is a risk

src/tests/test_installations.py
```diff
@@ -159,3 +164,11 @@ def test_delete_installation_invalid(test_app, monkeypatch, installation_id, sta
     assert response.status_code == status_code, print(installation_id)
     if isinstance(status_details, str):
         assert response.json()["detail"] == status_details, print(installation_id)
+
+
+def test_list_devices(test_app, monkeypatch):
+    monkeypatch.setattr(crud, "get_custom_entry", mock_get_custom_entry)
+
+    response = test_app.get("/installations/list_devices/1?timestamp=2020-11-13T08:18:45.447773")
```
frgfm, line 172: "list_devices" --> "list-devices"

src/app/api/routes/installations.py
```diff
@@ -31,3 +32,13 @@ async def update_installation(payload: InstallationIn, installation_id: int = Pa
 @router.delete("/{installation_id}/", response_model=InstallationOut)
 async def delete_installation(installation_id: int = Path(..., gt=0)):
     return await crud.delete_entry(installations, installation_id)
+
+
+@router.get("/list_devices/{site_id}", response_model=List[int])
+async def get_all_at_given_ts_and_site(timestamp: datetime, site_id: int = Path(..., gt=0)):
```
frgfm, line 38: I think we need to incorporate timestamp in a payload and type it with a schema

src/tests/test_installations.py
```diff
@@ -159,3 +164,11 @@ def test_delete_installation_invalid(test_app, monkeypatch, installation_id, sta
...
+    assert response.status_code == 200
+    assert response.json() == [INSTALLATION_TABLE[0]["device_id"]]
```
frgfm, line 174: I really think the patching is too hard for this test to be effective. We should patch something doing the same thing dynamically, simply sterilizing all actual DB operations

frgfm (changes requested):
Thanks for the PR! Got a few comments, let me know what you think!

src/app/api/routes/installations.py
fe51: At first sight, in fetch_all, we could add logical operator as string in query filter dict add have conditions on query.where according to it, but maybe that complicates things too much. (and would be "==" by defautl)

```python
async def fetch_all(table: Table, query_filters: Optional[Dict[str, Any]] = None) -> List[Dict[str, Any]]:
    query = table.select()
    if isinstance(query_filters, dict):
        for query_filter_key, query_filter_value in query_filters.items():
            query = query.where(getattr(table.c, query_filter_key) == query_filter_value)
    return await database.fetch_all(query=query)

File: src/app/api/routes/installations.py (hunk 1 above)
Comment (florianriche) [reply src/app/api/routes/installations.py:37]: I  support @frgfm point of view. I would create a dedicated schema to  make sure. In addition, it will make it clearer what the route returns.

File: src/app/api/routes/installations.py (hunk 2 above)
Comment (florianriche) [reply src/app/api/routes/installations.py:42]: what is the meaning of star8ts and end_ts when it is linked to an installation?

File: src/app/api/routes/installations.py (hunk 3 above)
Comment (florianriche) [reply src/app/api/routes/installations.py:44]: You suggest to create a Dict¨str: Tuple(Any, Optionnal[Any])) where the Opional[Any] would be a logical operator ? It could work but I fear that by trying to be too generic, we make it unclear what we are doing in the code.

File: src/app/api/routes/installations.py (hunk 2 above)
Comment (frgfm) [reply src/app/api/routes/installations.py:42]: @florianriche it is the time range when the installation takes place (imagine retrieving a device or moving it to another site, it will fill the `end_ts` value of the current installation.

File: src/app/api/routes/installations.py (hunk 3 above)
Comment (frgfm) [reply src/app/api/routes/installations.py:44]: I thought about two propositions:
- either we keep the current query_filters that uses the equality operator, and two other arguments `lte_filters` and `gte_filters` for "less than or equal" and "greater than or equal".
- in the query filter dict, if the value is a dict, the key specifies the operator, if not it defaults to equality

File: src/app/api/routes/installations.py (hunk 1 above)
Comment (florianriche) [reply src/app/api/routes/installations.py:37]: Now that I think of it, if we are in the installations routes, it would be weird to have `/installations/{site_id)]/liste-devices` knowing that `/installations/{id}` is to retrieve an installation.
That could be confusing. maybe `/installations/site-devices/{site_id}`

Comment (frgfm) [reply src/app/api/routes/installations.py:37]: I agree with your last suggestion!

File: src/app/api/routes/installations.py (hunk 2 above)
Comment (frgfm) [reply src/app/api/routes/installations.py:42]: @martin1tab regarding the earlier suggestion:
after importing `and_` from sqlalchemy, you can do :
```
    query = installations.select(installations.c.device_id).where(and_(installations.c.site_id == site_id,
                                                                       installations.c.start_ts <= timestamp,
                                                                       installations.c.end_ts >= timestamp))
```

and then simply:
```
return await database.execute(query=query)
```
(don't forget to import database)

File: src/app/api/routes/installations.py (hunk 3 above)
Comment (frgfm) [reply src/app/api/routes/installations.py:44]: We should handle this in a follow-up issue, let's make the query and execution right in the decorated route for now.

Considering that this should be `return await crud.get_custom_entry(query)` rather than `return await crud.get_custom_entry(installations)` and that the tests don't fail, I fear we can't rely on the unittests for now

File: src/app/api/crud.py [hunk 8]
```diff
@@ -70,3 +70,7 @@ async def delete_entry(table: Table, entry_id: int = Path(..., gt=0)) -> Dict[st
     await delete(entry_id, table)
 
     return entry
+
+
+async def get_custom_entry(query):
+    return await database.execute(query=query)
```
frgfm, line 76: Can you remove this for now please? (see the suggestions in installation to implement the route)

src/app/api/routes/installations.py
martin1tab: OK

src/app/api/schemas.py
```diff
@@ -19,6 +19,10 @@ class _Id(BaseModel):
     id: int = Field(..., gt=0)
 
 
+class Timestamp(BaseModel):
+    timestamp: datetime
```
jeanpasquier75, line 23: ```timestamp: datetime = Field(..., example=datetime.utcnow())```

martin1tab:
PR is ready for review now

frgfm (approved):
Thanks a lot! 
Let's keep in mind that we will need to add some unittests for this later!

PR #57, opened 2020-11-23
Fix crud: types & update

jeanpasquier75:
Hello, quick PR to fix some crud: I recently pulled master and got some Python AttributeError: 'Record' object has no attribute 'update'. I dug into [databases postgres implementation](https://github.com/encode/databases/blob/master/databases/backends/postgres.py) and figured out that database.fetch_* operation return an abstract Mapping object and postgres Record is one implementation. 

I guess that a Mapping implementation must implement Python __getitem__ method to perform something like:

``` 
entry = database.fetch_one(...)  # Record
field = entry["field"] 
```

which seams to be OK for API simple routes operations, FastAPI should perform some casting Record -> pydantic.BaseModel

That is why I changed return type hint for low level operations _get_, _fetch_one_ & _fetch_all_ to `Mapping[str, Any]`. I forced dictionary casting for higher level _get_entry_ operation in order for us to deal with Python dictionaries in customized routes and perform something like:
```
entry = crud.get_entry()  # dictionary
entry.update({...})
```

Update device location should work with this patch.

I also improved _update_entry_ to perform only one database operation (update) instead of 2 (get + update), keeping a 404 exception if no update was performed. I changed _mock_put_ in conftest to reflect the modification. All unit tests should pass.


src/app/api/deps.py
```diff
@@ -50,7 +50,7 @@ async def get_current_access(security_scopes: SecurityScopes, token: str = Depen
     except (JWTError, ValidationError, KeyError):
         raise unauthorized_exception("Invalid credentials", authenticate_value)
 
-    entry = await crud.get(entry_id=access_id, table=accesses)
+    entry = await crud.get_entry(accesses, access_id)
```
frgfm, line 53: Over here, `crud.get` was selected so we can specify the exception "Invalid credentials". If we change to `crud.get_entry`, we're gonna perform a double check. I'm not sure it's good idea

src/app/api/crud.py
```diff
@@ -43,7 +43,7 @@ async def put(entry_id: int, payload: BaseModel, table: Table) -> int:
 
 async def delete(entry_id: int, table: Table):
     query = table.delete().where(entry_id == table.c.id)
-    return await database.execute(query=query)
+    await database.execute(query=query)
```
frgfm, line 46: This does the output, could you add the typing `-> None` for this function then?

src/app/api/crud.py
```diff
@@ -55,13 +55,16 @@ async def get_entry(table: Table, entry_id: int = Path(..., gt=0)) -> Dict[str,
...
+    return dict(**entry)
 
 
 async def update_entry(table: Table, payload: BaseModel, entry_id: int = Path(..., gt=0)) -> Dict[str, Any]:
-    await get_entry(table, entry_id)
     entry_id = await put(entry_id, payload, table)
 
+    if not isinstance(entry_id, int):
+        raise HTTPException(status_code=404, detail="Entry not found")
+
```
frgfm, line 67: Nice this will reduce some overhead :+1:

src/app/api/crud.py
```diff
@@ -55,13 +55,16 @@ async def get_entry(table: Table, entry_id: int = Path(..., gt=0)) -> Dict[str,
     entry = await get(entry_id, table)
     if entry is None:
         raise HTTPException(status_code=404, detail="Entry not found")
-    return entry
+
+    return dict(**entry)
```
frgfm, line 59: does the casting also work with `dict(entry)`?

frgfm (commented):
Thanks for the PR! I added a few comments, let me know your opinion on those.

Also, can you produce some instructions to reproduce the error on master please? I think I understand the point you're raising but I have no way to test it.

src/app/api/deps.py
jeanpasquier75: Oh I see, I rollback this change

src/app/api/crud.py
jeanpasquier75: The delete execution was already not producing any return value. I can explicitly note it

jeanpasquier75:
> Also, can you produce some instructions to reproduce the error on master please? I think I understand the point you're raising but I have no way to test it.

Try to update a device location (careful to update a device you own)

src/app/api/crud.py
frgfm: Yes please :pray:  (I remember myself trying to figure out what it was returning, and didn't find anything)

jeanpasquier75: I tried to make the query returning the deleted entry to avoid doing 2 db operations (get + delete) but it seams to be impossible

frgfm: What do you mean? it's alright not returning anything, I was just asking to add typing `-> None` to the return type of the function

frgfm (approved):
Thanks!

Issue #45, opened 2020-11-19
[scopes] Defining scopes for all routes


frgfm:
Hey there :wave: 

Playing around with the PRs, I noticed that some routes are protected by scopes while others aren't. If some of those are correct, I just want to make sure we do consider carefully those choices for each router:

- [x] alerts: _should it be exposed to everyone on Read? admin & device for create, update, delete?_ (#99, #122)
- [x] devices: _"admin" by default for all routes, "me" added to self-ID routes_ (#121)
- [x] events: _should it be exposed to everyone on Read? admin & device for create, update, delete?_ (#59, #99)
- [x] installations: _should it be exposed to everyone on Read? admin only for create, update, delete?_ (#99, #120)
- [x] media: _we need some scope but can't be admin only, since a logged user whose scope includes a given device, should have access to its media_ (#99, #118)
- [x] sites: _should it be exposed to everyone on Read? admin only for create, update, delete?_ (#58, #99) 
- [x] users: _"admin" by default for all routes, "me" added to self-ID routes_ (#15, #44)

Those where I have no clue whether we should add some:
- [x] accesses: _"admin" by default for all routes_ (#104)
- [x] login: no scope

cc @florianriche @jeanpasquier75 @fe51 @martin1tab

jeanpasquier75:
You are right we should protect the API 👍 
* login must not have a required scope -> everyone can request an access
* accesses aims to be removed (I guess) and is used so far for debugging but we should require an admin scope (to prevent exposing list of login/scopes)

fe51:
I think we need to protect all routes expect login to request an access (thus for those you were suggesting, at least, to have any of the scopes ?)

- users: ok with the suggestion
- alerts: R for the owner of the device ? (and admin & device)
- devices: ok as it is (with some route with "device" scope)
- installations: CRUD for the owner of the device ? (and admin)
- media: CRUD for the owner of the device ? (and admin)

frgfm:
So we would get the following :

router | Create | Read | Update | Delete
-- | -- | -- | -- | --
alerts | admin, device | None | admin, device | admin
devices | admin, user | admin, user | admin, user, self | admin, user
events | admin, device | None | admin, device | admin
installations | admin | None | admin | admin
media | admin, device | admin, device | admin, device | admin
sites | admin | None | admin | admin
users | admin | admin, self | admin, self | admin, self
accesses | NA | admin | NA | NA
login | None | NA | NA | NA


[...]

PR #58, opened 2020-11-23
[scopes][sites]Add admin scopes to CUD routes

fe51:
In the issue #45 we decide to define scopes for all routes in order to protect the API

This PR aims at adding scopes for sites route.

However, should we add the right to any auth user to CU sites ? What do you think ?
(I would add "me" in scopes in this case)

Any feedback is welcome !

### DIFF
--- a/src/app/api/routes/sites.py
+++ b/src/app/api/routes/sites.py
@@ -1,15 +1,16 @@
 from typing import List
-from fastapi import APIRouter, Path
+from fastapi import APIRouter, Path, Security
 from app.api import crud
 from app.db import sites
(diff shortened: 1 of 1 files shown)


fe51:
** tests need to be updated**

frgfm (approved):
Looks good to me! Like in the other PR, let's wait for another review. And about your question, modifying/creating sites should be validated by a system administrator, I think it would be dangerous allowing an authenticated user to get those rights.

frgfm:
@jeanpasquier75 @florianriche @martin1tab We'd need your inputs on this PR please :pray:

jeanpasquier75 (approved):

PR #60, opened 2020-11-23
[doc][login] Add description to login route

fe51:
This PR aims at adding description to the login route.

I am hesitating, adding in the description, something like "This API follows the OAuth 2.0 specification.
Depending on the scope of the access related to credentials, the token expires at a given time"

What do you think ? 

image

Any feedback is welcome !

### DIFF
--- a/src/app/api/routes/login.py
+++ b/src/app/api/routes/login.py
@@ -13,7 +13,11 @@ router = APIRouter()
 @router.post("/access-token", response_model=Token)
 async def create_access_token(form_data: OAuth2PasswordRequestForm = Depends()):
     """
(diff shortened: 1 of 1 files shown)


frgfm:
@fe51 yup, I agree that specifying how auth is done is a good idea! And perhaps even specify the default duration of the token before it expires!

fe51:
Screen in description updated. @frgfm :-)
```
    This API follows the OAuth 2.0 specification
    If the credentials are valid, creates a new access token
    By default, the token expires after 1 hour

frgfm (approved):
Thanks for the edits!

PR #62, opened 2020-11-24
Api client v0.1.0

### LINKED ISSUE #61: [client] Extend API client features for platform demo
Following up on https://github.com/pyronear/pyro-platform/issues/19, the following features need to be added to the client:
- [ ] Read all devices within the scope of the user
- [ ] Read the sites
- [ ] Read the list of device IDs of a given site at a given time (which ongoing on the API-side in #49)
- [ ] Read the alerts (while we don't have a webhook)
- [ ] Download the frame (media) linked to an alert

Considering the scopes of several routes are going to change (more restricted) with #45, this is all the more important since, if they are able to use GET on some routes without authentication, this might not be true in the future!

florianriche:
Closes #61 

Add routes to answer the following use cases:

- [x] all devices within the scope of the user
- [x]  Read the sites
- [x]  Read the list of device IDs of a given site at a given time (which ongoing on the API-side in #49)
- [x]  Read the alerts (while we don't have a webhook)
- [x]  Download the frame (media) linked to an alert

Notes:
1: As we currently don't have a storage service, the returned media is always the same.
2: I have added several routes for the alerts to query all the alerts; the unacknowledged ones, the ongoing ones.
3: For the use case to get the list of devices IDS, it might not be the final design as the PR #49  that implements it is not yet merged.


florianriche:
Notes for the reviewers, I would suggest to go commit by commit :)

src/app/config.py
```diff
@@ -29,3 +29,4 @@
         "Missing Credentials. Please set 'SUPERUSER_LOGIN' and 'SUPERUSER_PWD' in your environment variables")
 
 BUCKET_NAME: str = "mypyroneartest"
+DUMMY_BUCKET_FILE = "app/static/dummy_media.jpg"
```
frgfm, line 32: This is typically something we could put in a release attachment, I'm really against making the github history heavier with a content file, if this can be avoided. And since this is for testing, we can even get the URL of a publicly available image, which might be the best option here

frgfm: Here is one for instance: https://ec.europa.eu/jrc/sites/jrcsh/files/styles/normal-responsive/public/growing-risk-future-wildfires_adobestock_199370851.jpeg

frgfm (changes requested):
Thanks for the PR!
I added some comment about two things: the concept of "ongoing alerts", and the content file that was added. Looking forward to your feedback!

src/app/config.py
florianriche: Ok but I think we can't just return a URL, I understood from @akilditu comment in https://github.com/pyronear/pyro-api/issues/61 that the API should return an image directly.
 Downloading that file everytime and then returns it seems heavy to me.

frgfm: Mmmh why not let the API return the URL, and have the client download from the URL? (in case it's something that requires creds, we pass parameters down in the URL?)

That would remove a lot of data transit through the API

src/app/api/routes/alerts.py
```diff
@@ -59,3 +60,13 @@ async def link_media(payload: AlertMediaId,
         )
     existing_alert["media_id"] = payload.media_id
     return await crud.update_entry(alerts, AlertIn(**existing_alert), alert_id)
+
+
+@router.get("/ongoing", response_model=List[AlertOut])
+async def fetch_ongoing_alerts():
+    return await crud.fetch_all(alerts, {"type": AlertType.start})
```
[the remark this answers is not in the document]
florianriche: We may have multiple alerts launched for the same event_id?
From what you are saying the SQL query should be something like:

```SQL
SELECT * from 
    alerts a
where 
     type = AlertType.start
     and event_id not in (select event_id from alerts where type=Alert.end)

```

src/app/config.py
florianriche: I have created 3 methods in the client for the different use case. We will delete what is not required anymore.

src/app/api/routes/alerts.py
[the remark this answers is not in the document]
frgfm: Looks perfect to me!

Akilditu (dismissed):
Thanks for the PR!

Seems fine to me 👍

src/app/api/routes/alerts.py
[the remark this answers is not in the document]
florianriche: I added a new commit to handle that logic. There is probably a way to do it in a clenaer way using Sqlaclhemy models instead of calling the table directly. I think we should do it if we want to handle complex queries in a cleany way. 

However, for now it is working. I will move the created crud function after the refacto.

frgfm (approved):
Thanks a lot! Let's see later how we refactor the crud, but this is already more than enough for the release!

frgfm:
Deleting the branch since it's no longer useful

PR #64, opened 2020-11-24
[scopes][permissions] update permission condition from "and" to "or" 

fe51:
Following the super relevant comment from @florianriche in #45, this PR aims at correcting the authentication behaviour with respect to the scope list.

The list of scopes appears to be an "and" condition, instead of a "or" as we want to design it.

`deps.py` file has been updated to reflect the correct behaviour.

Any feedback is welcome !

### DIFF
--- a/src/app/api/deps.py
+++ b/src/app/api/deps.py
@@ -55,9 +55,8 @@ async def get_current_access(security_scopes: SecurityScopes, token: str = Depen
     if entry is None:
         raise unauthorized_exception("Invalid credentials", authenticate_value)
 
-    for scope in security_scopes.scopes:
-        if scope not in token_data.scopes:
-            raise unauthorized_exception("Permission denied", authenticate_value)
+    if set(token_data.scopes).isdisjoint(security_scopes.scopes):
+        raise unauthorized_exception("Permission denied", authenticate_value)
 
     return AccessRead(**entry)
 


src/app/api/deps.py
```diff
@@ -55,9 +55,8 @@ async def get_current_access(security_scopes: SecurityScopes, token: str = Depen
     if entry is None:
         raise unauthorized_exception("Invalid credentials", authenticate_value)
 
-    for scope in security_scopes.scopes:
-        if scope not in token_data.scopes:
-            raise unauthorized_exception("Permission denied", authenticate_value)
+    if set(token_data.scopes).isdisjoint(security_scopes.scopes):
```
frgfm, line 58: Perhaps a more readable solution: 
```
if not any(scope in token_data.scopes for scope in security_scopes.scopes):
```

frgfm (commented):
Small comment but the rest looks fine!

src/app/api/deps.py
fe51: Question of perspective :-)  However, from a python user point of view, your suggestion might be more readable !
Let's wait another reviewer point of view but honestly I do not have any preference

frgfm: No no, really, if both implementations are equivalent behaviour-wise, let's go with yours!

frgfm (approved):
Thanks for the feedback on my previous comment!

jeanpasquier75:
If I understand we make this change because some routes require several scopes (eg: fetch_my_devices requires scopes=["admin", "me"]) ?

At the beginning I designed scopes to be used to separate "simple" users (me) and "super" users (admin) but a user can be both as well (having an access with scopes="admin me"). This way the routes require only the minimum for itself (eg: fetch_my_devices would have scopes=["me"] but a super user having scopes="admin me" can still query the route). What do you think ?

frgfm:
@jeanpasquier75 yup, but without this PR, some routes were not allowed unless the user had both the "admin" AND the "me" scope. While we would have wanted any of those (or the minimum indeed)

But I realize that you mean the `not in` operator you used was because of the string concatenation, correct?

PR #68, opened 2020-11-27
[Refactor] Use test DB in unit Tests

florianriche:
Gives the ability to use a real sqlite DB instead of patching the crud functions.
The DB is completely cleaned at the end of every test.

It is working on my side; the cumbersome part was to be able to have both async and synchronous unit tests.
For now, I only adapted the `create_user` unit test. If you agree with this design, I will refactor all the unit tests.


.gitignore
```diff
@@ -140,3 +140,6 @@ dmypy.json
 # Client
 client/docs/build
 client/build
+
+# Test DB
+*.db
```
frgfm, line 145: EOF line missing I think

requirements.txt
```diff
@@ -13,3 +13,5 @@ pytest-asyncio>=0.14.0
 requests>=2.22.0
 asyncpg>=0.20.0
 coverage>=4.5.4
+aiosqlite==0.16.0
+httpx==0.16.1
```
frgfm, line 17: lower bound constraints if possible :pray:

src/tests/conf_test_db.py
```diff
@@ -0,0 +1,31 @@
...
+SQLALCHEMY_DATABASE_URL = "sqlite:///./test.db"
+engine = create_engine(
+    SQLALCHEMY_DATABASE_URL, connect_args={"check_same_thread": False}
+)
+metadata.create_all(bind=engine)
+database = Database(SQLALCHEMY_DATABASE_URL)
+
+
+def flush_test_db():
+
```
frgfm, line 15: Perhaps a short doctring for easier understanding?

src/tests/conf_test_db.py
```diff
@@ -0,0 +1,31 @@
...
+    with contextlib.closing(engine.connect()) as con:
+        trans = con.begin()
+        for table in reversed(metadata.sorted_tables):
+            con.execute(table.delete())
+        trans.commit()
+
+
+async def populate_db(test_db, table, data):
```
frgfm, line 23: If possible, could we add some typing?

src/tests/conf_test_db.py
```diff
@@ -0,0 +1,31 @@
...
+    for entry in data:
+        query = table.insert().values(entry)
+        await test_db.execute(query=query)
+
+
+async def get_entry_in_db(test_db, table, entry_id):
```
frgfm, line 29: same here

src/tests/test_users.py
```diff
@@ -102,24 +110,42 @@ def test_fetch_users(test_app, monkeypatch):
...
 
     # Sterilize DB interactions
-    mock_user_table = deepcopy(USER_TABLE)
+    mock_user_table = deepcopy(USER_TABLE_FOR_DB)
     mock_access_table = deepcopy(ACCESS_TABLE)
-    _patch_session(monkeypatch, mock_user_table, mock_access_table)
+    monkeypatch.setattr(crud, "database", test_db)
+    monkeypatch.setattr(security, "hash_password", pytest.mock_hash_password)
+    await populate_db(test_db, db.accesses, ACCESS_TABLE)
+    await populate_db(test_db, db.users, USER_TABLE_FOR_DB)
```
frgfm, line 122: Correct me if I'm wrong, but since this inserts data and doesn't clean beforehand, conducting test sequentially will result in some tests starting with an already filled db?

frgfm (commented):
Thanks for the PR! Would love to switch to an actual mock sqllite table, I added a few comments!

src/tests/test_users.py
florianriche: `flush_test_db`  is called at the end of every test when the fixture `test_db` is no longer used. This function keeps the db schema but delete all rows.

src/tests/conf_test_db.py
florianriche: Good idea indeed, will do.

florianriche:
Replacing all the tests would be huge and hard to review, I would suggest to validate this PR and then do another one to effectively update the tests.

frgfm (approved):
Thanks a lot for the edits!

PR #69, opened 2020-11-27
fix: get_all_at_given_ts_and_site timestamp call

florianriche:
small error (caught by pyflakes on my machine): timestamp was not defined but belonged to the payload object.


frgfm (approved):
Good catch, thanks!

PR #70, opened 2020-11-28
Refactor unit tests to use db

florianriche:
### Massive PR
To review commit by commit; One commit == one file

1. Switch the test DB from sqlite to postgresql because sqlite hasn't exactly the same behavior as postgresql and some features that we use are not available. Rather than alter the code, I changed the used DB. It implied to create a new dedicated Test DB in  the docker-compose. 
2. Found some issues in some routes; fixed those issues. Errors were in upload_media and in linking media to alerts.
3. Replace every test with a version that use the DB.
4. The tests now take a bit longer because I connect/populate with data /empty/disconnect the DB for every test.Could be improved probably.


src/tests/test_alerts.py
```diff
@@ -16,43 +43,43 @@
...
+    await populate_db(test_db, db.accesses, ACCESS_TABLE_FOR_DB)
+    await populate_db(test_db, db.users, USER_TABLE_FOR_DB)
 
-    # Sterilize DB interactions
-    mock_alert_table = deepcopy(ALERT_TABLE)
-    _patch_session(monkeypatch, mock_alert_table)
+    await populate_db(test_db, db.devices, DEVICE_TABLE_FOR_DB)
+    await populate_db(test_db, db.events, EVENT_TABLE_FOR_DB)
+    await populate_db(test_db, db.alerts, ALERT_TABLE_FOR_DB)
+    await populate_db(test_db, db.media, MEDIA_TABLE_FOR_DB)
```
frgfm, line 73: Mmmmh, patching all tables while only two routes require more than the access table might be an overkill (and it helps raising error if it interacts with the table while it shouldn't)

Perhaps we could have a more flexible patching?

src/tests/test_devices.py
```diff
@@ -4,74 +4,73 @@
...
 
-        async def mock_post_access(login, password, scopes):
-            if any(entry['login'] == login for entry in mock_access_table):
-                raise HTTPException(status_code=400, detail=f"An entry with login='{login}' already exists.")
+ACCESS_TABLE_FOR_DB = list(map(update_only_datetime, ACCESS_TABLE))
+USER_TABLE_FOR_DB = list(map(update_only_datetime, USER_TABLE))
+DEVICE_TABLE_FOR_DB = list(map(update_only_datetime, DEVICE_TABLE))
+# EVENT_TABLE_FOR_DB = list(map(update_only_datetime, EVENT_TABLE))
+# ALERT_TABLE_FOR_DB = list(map(update_only_datetime, ALERT_TABLE))
+# MEDIA_TABLE_FOR_DB = list(map(update_only_datetime, MEDIA_TABLE))
```
frgfm: This can be removed

src/tests/test_devices.py
```diff
@@ -4,74 +4,73 @@
...
-        monkeypatch.setattr(devices, "post_access", mock_post_access)
+async def init_test_db(monkeypatch, test_db):
+    monkeypatch.setattr(security, "hash_password", pytest.mock_hash_password)
+    monkeypatch.setattr(crud, "database", test_db)
+    await populate_db(test_db, db.accesses, ACCESS_TABLE_FOR_DB)
+    await populate_db(test_db, db.users, USER_TABLE_FOR_DB)
+    await populate_db(test_db, db.devices, DEVICE_TABLE_FOR_DB)
+    # await populate_db(test_db, db.events, EVENT_TABLE_FOR_DB)
+    # await populate_db(test_db, db.alerts, ALERT_TABLE_FOR_DB)
+    # await populate_db(test_db, db.media, MEDIA_TABLE_FOR_DB)
```
frgfm: Same here

src/tests/test_events.py
```diff
@@ -15,26 +15,34 @@
...
-    monkeypatch.setattr(crud, "get", pytest.mock_get)
-    monkeypatch.setattr(crud, "fetch_all", pytest.mock_fetch_all)
-    monkeypatch.setattr(crud, "post", pytest.mock_post)
-    monkeypatch.setattr(crud, "put", pytest.mock_put)
-    monkeypatch.setattr(crud, "delete", pytest.mock_delete)
+DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%S.%f"
+
+
+def update_only_datetime(entity_as_dict):
+    to_return = entity_as_dict.copy()
```
frgfm, line 20: not a deepcopy and dicts are mutable in python

src/tests/test_sites.py
```diff
@@ -14,27 +15,42 @@
...
+def compare_entries(ref, test):
+    for k, v in ref.items():
+        if isinstance(v, float):
+            # For float issues
+            assert abs(v - test[k]) < 1E-5
+        else:
+            assert v == test[k]
+
+def update_only_datetime(entity_as_dict):
+    to_return = entity_as_dict.copy()
```
frgfm, line 29: same here

src/tests/test_login.py
```diff
@@ -1,42 +1,35 @@
...
-
 @pytest.mark.parametrize(
     "payload, status_code, status_detail",
     [
         [{"username": "foo"}, 422, None],
         [{"password": "foo"}, 422, None],
         [{"username": "unknown", "password": "foo"}, 400, "Invalid credentials"],  # unknown username
         [{"username": "first", "password": "second"}, 400, "Invalid credentials"],  # wrong pwd
-        [{"username": "first_login", "password": "first_pwd"}, 200, None],  # valid
+        [{"username": "first_login", "password": "first_pwd"}, 200, None],  # should throw an error !!valid
```
frgfm, line 20: Why should it throw an error?

src/tests/test_installations.py
```diff
@@ -15,26 +60,48 @@
...
+    monkeypatch.setattr(crud, "database", test_db)
+    await populate_db(test_db, db.accesses, ACCESS_TABLE_FOR_DB)
+    await populate_db(test_db, db.users, USER_TABLE_FOR_DB)
+
+    await populate_db(test_db, db.devices, DEVICE_TABLE_FOR_DB)
+    await populate_db(test_db, db.events, EVENT_TABLE_FOR_DB)
+    await populate_db(test_db, db.alerts, ALERT_TABLE_FOR_DB)
+    await populate_db(test_db, db.media, MEDIA_TABLE_FOR_DB)
+    await populate_db(test_db, db.sites, SITE_TABLE_FOR_DB)
+    await populate_db(test_db, db.installations, INSTALLATION_TABLE_FOR_DB)
```
frgfm: most of those patches are not required for most routes

src/app/api/routes/media.py
```diff
@@ -103,6 +103,7 @@ async def upload_media(media_id: int = Path(..., gt=0),
             status_code=500,
             detail="The upload did not succeed"
         )
+    entry = dict(**entry)
```
frgfm, line 106: entry isn't already a dictionary?

src/app/api/routes/alerts.py
```diff
@@ -58,6 +58,14 @@ async def link_media(payload: AlertMediaId,
...
             detail="Permission denied"
         )
+
+    existing_media = await crud.fetch_one(media, {"id": payload.media_id})
+    if existing_media is None:
+        raise HTTPException(
+            status_code=404,
+            detail="Media does not exist"
+        )
+    existing_alert = dict(**existing_alert)
```
frgfm, line 68: isn't this already a dict?

src/tests/conftest.py
```diff
@@ -114,9 +57,11 @@ async def test_app_asyncio(__app):
 
 
 @pytest.fixture(scope="function")
-def test_db():
+async def test_db():
     try:
+        await test_database.connect()
         yield test_database
     finally:
         print("Clean the DB")
```
frgfm, line 65: Actually, we can remove this, it won't get printed

frgfm (commented):
Thanks for the much needed PR! I left a few comments, let me know what you think!

src/tests/test_installations.py
florianriche: Those are not patches per se, but because of the foreign keys constraints, I can't populate a DB referring another DB if the foreign entities don't exist. And so on, hence why I need to populate everything here.

src/app/api/routes/media.py
florianriche: Nope, it's a record. I got an exception when launching the corresponding test at first.
NB: maybe it's converted implicitly using pydantic when we are in "prod" but the test was failing.

src/app/api/routes/alerts.py
florianriche: same comment:
Nope, it's a record. I got an exception when launching the corresponding test at first.
NB: maybe it's converted implicitly using pydantic when we are in "prod" but the test was failing.

src/tests/conftest.py
florianriche: It's printed in case there is a failing test.
Thought it was informative in case we wanted to be sure it's cleant.

src/tests/test_login.py
florianriche: Comment I forgot to delete. At first I misunderstood what this function was doing.

src/tests/test_devices.py
florianriche: will do

src/tests/test_devices.py
florianriche: will do

docker-compose.yml
```diff
@@ -22,6 +23,15 @@ services:
...
       - POSTGRES_PASSWORD=pyro_api
       - POSTGRES_DB=pyro_api_dev
+  test_db:
+    image: postgres:12.1-alpine
+    volumes:
+      - postgres_data_test:/var/lib/postgresql/data_test/
+    environment:
+      - POSTGRES_USER=pyro_api_test
+      - POSTGRES_PASSWORD=pyro_api_test
+      - POSTGRES_DB=pyro_api_dev_test
```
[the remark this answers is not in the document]
florianriche: Indeed, to do in another review

src/tests/test_alerts.py
florianriche: copypasted: Those are not patches per se, but because of the foreign keys constraints, I can't populate a DB referring another DB if the foreign entities don't exist. And so on, hence why I need to populate everything here.

frgfm (approved):
Thanks for the edits! The extra coverage is much appreciated

PR #73, opened 2020-11-29
chore: Removed mandatory test dependency installation

frgfm:
This PR makes test dependency installations optional. The requirement file is still available inside the docker thanks to the volume binding of docker-cpmpose, and users can install them very easily before running tests.

The goal is to reduce minimal dependencies of the container as those tests should not be run in production, hence the removal from core dependencies.


jeanpasquier75 (dismissed):
I do agree on this PR ! I have one comment: a more common name of the file is `requirements-dev.txt` 😉

frgfm:
> I do agree on this PR ! I have one comment: a more common name of the file is `requirements-dev.txt`

Good point, just fixed it!

jeanpasquier75 (approved):

PR #74, opened 2020-11-29
test: Added unittests and typing verification for API client

frgfm:
This PR introduces the following features:
- typing verification for the API client using mypy
- unittesting the API client package

Any feedback is welcome!
_Please note that I only added tests for the GET routes, since we need to see how we want to test the other method types without modifying the database._


florianriche (commented):
Do you see a way to use the app schemas for additional type checking?

As for how to test the other routes, some ideas in no particular order and without much thinking:
* Test on local DB and we don't care
* Add another test app in the docker-compose and we ping on it
* Create a dedicated fastaAPI (or not) server at unit test launch time with dedicated hardcoded return functions

frgfm:
> Do you see a way to use the app schemas for additional type checking?

Mmmh, I think it would require adding too much dependency on the client side. Right now, it's quite nice that the only dep is requests. Plus since, the client only get the output of routes, each model is casted into dict/json. So I guess the best way is to type it just like in this PR


> As for how to test the other routes, some ideas in no particular order and without much thinking:
> 
> * Test on local DB and we don't care
> * Add another test app in the docker-compose and we ping on it
> * Create a dedicated fastaAPI (or not) server at unit test launch time with dedicated hardcoded return functions

Not sure what you mean with the last option, but I like the 2nd. And the nice thing with parallel jobs is that the environent is completely independent so we can connect the client to an API linked to the test DB.

florianriche (approved):
Is there something to add to the setup.py to run the unit tests?

frgfm:
> Is there something to add to the setup.py to run the unit tests?

I think it is possible, but since the CI is running it, I wouldn't advise to do it on setup.py or package installation may take quite a while for what it is! Anyway, I managed to fix the previous issue of the unittest, so the PR is ready for review!

client/test/test_client.py
```diff
@@ -0,0 +1,33 @@
+import unittest
+from requests import ConnectionError
+
+from pyroclient import client
+from pyroclient.exceptions import HTTPRequestException
+
+
+class ClientTester(unittest.TestCase):
+
+    def _test_route_retun(self, response, return_type):
```
florianriche, line 10: typo?

florianriche (approved):
Looks good to me besides a typo

client/test/test_client.py
frgfm: Fixed, thanks!

florianriche (approved):

PR #76, opened 2020-11-29
feat: Ensures related access is deleted after user/device deletion

frgfm:
This PR ensures that once a user or a device is deleted, its corresponding access is deleted as well.

I had to change the table definitions in unittests since accesses were sometimes not consistent with device/user tables.


src/app/api/routes/devices.py
```diff
@@ -35,7 +35,10 @@ async def update_device(payload: DeviceIn, device_id: int = Path(..., gt=0)):
 
 @router.delete("/{device_id}/", response_model=DeviceOut)
 async def delete_device(device_id: int = Path(..., gt=0), _=Security(get_current_user, scopes=["admin"])):
```
florianriche, line 37: Can we do a "db" transaction like what JB suggested in one o his reviews?

src/app/api/routes/users.py
```diff
@@ -71,4 +71,8 @@ async def update_user_password(
 
 @router.delete("/{user_id}/", response_model=UserRead)
 async def delete_user(user_id: int = Path(..., gt=0), _=Security(get_current_user, scopes=["admin"])):
-    return await crud.delete_entry(users, user_id)
+    # Delete user entry
+    entry = await crud.delete_entry(users, user_id)
+    # Delete access
```
florianriche, line 76: same commet

src/app/api/routes/devices.py
frgfm: Considering the PR is still hanging and I don't know how to use them as of now, I didn't consider this for this PR. But I still have one last thing to check with this PR (some troubles in the unittests)

src/tests/conftest.py
```diff
@@ -15,7 +15,7 @@ async def mock_current_user():
 
 
 async def mock_current_device():
-    return DeviceOut(id=1, owner_id=1, specs="raspberry", login="connected_device", created_at=datetime.now())
+    return DeviceOut(id=3, owner_id=1, specs="raspberry", login="connected_device", created_at=datetime.now())
```
florianriche, line 18: It might break some tests I fear if you merge with master; or maybe you already took care of it?

florianriche (approved):
I just have  a comment that is a question/Warning. If my comment is useless, then let's merge this PR

src/tests/conftest.py
frgfm: Yup already did, though this was cumbersome haha

frgfm:
> I just have a comment that is a question/Warning. If my comment is useless, then let's merge this PR
> 
> **EDIT: Should we not delete user/device when deleting an access as well?**

I wondered as well, but I felt it was quite dangerous. Only admin should be able to delete accesses directly, so I thought that self deletion of user/device should trigger the deletion of their access, but the other way is quite dangerous. But if you have a suggestion, happy to check it!

PR #81, opened 2020-12-01
fix: Fixed installations/site-devices route

frgfm:
This PR has multiple goals:
- making `start_ts` a required field for installations and adapting unittests
- fixing the `installations/site-devices` route and adding a unittest for it (while designing the end-to-end script I figured the route was not working)
- removing the timestamp passed to `installations/site-devices` and using current_ts (since there wouldn't be any need for other ts)

Any feedback is welcome!


src/app/api/routes/installations.py
```diff
@@ -53,13 +54,17 @@ async def delete_installation(installation_id: int = Path(..., gt=0)):
     return await crud.delete_entry(installations, installation_id)
 
 
-@router.post("/site-devices/{site_id}", response_model=List[int],
-             summary="Get all devices related to a specific site")
-async def get_all_at_given_ts_and_site(payload: Timestamp, site_id: int = Path(..., gt=0)):
+@router.get("/site-devices/{site_id}", response_model=List[int],
+            summary="Get all devices related to a specific site")
+async def get_active_devices_on_site(site_id: int = Path(..., gt=0)):
```
florianriche, line 59: We don't want to query for other timestamp? That could be useful for historical exploration though.

src/app/api/routes/installations.py
```diff
@@ -53,13 +54,17 @@ async def delete_installation(installation_id: int = Path(..., gt=0)):
...
-                                                                       installations.c.end_ts >= payload.timestamp))
-    return await database.fetch_all(query=query)
+    current_ts = datetime.utcnow()
+
+    query = installations.select().where(and_(installations.c.site_id == site_id,
+                                              installations.c.start_ts <= current_ts,
+                                              or_(installations.c.end_ts.is_(None),
+                                                  installations.c.end_ts >= current_ts)))
+
+    return [entry['device_id'] for entry in await database.fetch_all(query=query)]
```
florianriche, line 70: assign to a variable, check for length, and propose a separate message  "no entry found" in case there is no device?

florianriche (commented):
Just wondering if one could not query that route with another timestmap for historical exploration considerations?

src/app/api/routes/installations.py
frgfm: I first tried it, and the serialization of the timestamp is tricky (to serialize and both to get through the pydantic validator). So since the route was created for platform needs, I figured we should first get something to work that fits their needs. And them, perhaps extend the feature later if necessary :man_shrugging:

src/app/api/routes/installations.py
frgfm: Thought about it, but what's the difference with returning an empty list?
I mean, it should be an acceptable result that there is currently no active devices on site (without triggering an error), don't you think?

src/app/api/routes/installations.py
florianriche: Ok for me. When the need arise, it should not be too complicated to implement.

src/app/api/routes/installations.py
florianriche: Ok for me then.

florianriche (approved):

fe51 (approved):

PR #82, opened 2020-12-02
feat: Added a route for alert acknowledgement

### LINKED ISSUE #63: [Alerts] Acknowledge and close an alert
We have no route, for now, to specify an alert has been acknowledged and no way for a user to say an alert is complete.

The routes to do that should be quite straightforward.
One should also add functions to access those routes in the client.

**Side note:**
It also raises the question of how to interact with the events? I am not sure I clearly understand what they are supposed to do.

frgfm:
As per #63, this PR adds a route for alert acknowledgement. As discussed in the issue, the question of the route scope will be deferred to #45 and not be handled in this PR.

This will close #63 

Any feedback is welcome!

### DIFF
--- a/src/app/api/routes/alerts.py
+++ b/src/app/api/routes/alerts.py
@@ -2,7 +2,7 @@ from typing import List
 from fastapi import APIRouter, Path, Security, HTTPException
 from app.api import crud
 from app.db import alerts, AlertType, media
-from app.api.schemas import AlertBase, AlertOut, AlertIn, AlertMediaId, DeviceOut
+from app.api.schemas import AlertBase, AlertOut, AlertIn, AlertMediaId, DeviceOut, Ackowledgement, AcknowledgementOut
 from app.api.deps import get_current_device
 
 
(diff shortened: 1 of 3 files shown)


src/app/api/routes/alerts.py
```diff
@@ -70,6 +70,16 @@ async def update_alert(payload: AlertIn, alert_id: int = Path(..., gt=0)):
...
 
 
+@router.put("/{alert_id}/acknowledge", response_model=AlertOut, summary="Acknowledge an existing alert")
+async def acknowledge_alert(alert_id: int = Path(..., gt=0)):
+    """
+    Based on a alert_id, acknowledge the specified alert
+    """
+    payload = await crud.get_entry(alerts, alert_id)
+    payload['is_acknowledged'] = True
+    return await crud.update_entry(alerts, AlertIn(**payload), alert_id)
```
florianriche, line 80: I would suggest to replace by:
```suggestion
from schemas import Acknowledgement

async def acknowledge_alert(alert_id: int = Path(..., gt=0)):   
    return await crud.update_entry(alerts, Acknowledgement(is_acknowledged=True), alert_id)
```
That way if the alert_id doesn't exist you get the usual exception returned by update_entry and you avoid having to do two calls.

frgfm: Done!

florianriche (approved):
Looks good to me!

PR #83, opened 2020-12-02
Check login existence when updating user/access/device

### LINKED ISSUE #30: [accesses] The update_access route does not check that new login might already be taken
Upon access creation, there is a verification to avoid login duplication in the table. However when using the corresponding put method, if the user changes her/his login to another existing one, nothing will check for duplication.

florianriche:
When an access/user/device is updated; if the login is in the payload, we check if the updated login doesn't already belong to another user/access/device.

Close #30

### DIFF
--- a/src/app/api/routes/accesses.py
+++ b/src/app/api/routes/accesses.py
@@ -2,19 +2,28 @@ from typing import List, Dict, Any
 from fastapi import APIRouter, Path, HTTPException
 from app.api import crud, security
 from app.db import accesses
-from app.api.schemas import AccessBase, AccessRead, AccessAuth, AccessCreation, Cred, CredHash
+from app.api.schemas import AccessBase, AccessRead, AccessAuth, AccessCreation, Cred, CredHash, Login
 
 
 router = APIRouter()
 
 
-async def post_access(login: str, password: str, scopes: str) -> AccessRead:
+async def update_access_login(login: str, access_id: int):
+    """ Assume access_id exists and login does not exist elsewhere"""
+    return await crud.update_entry(accesses, Login(login=login), access_id)
+
+
+async def check_for_access_login_existence(login: str):
     # Check that the login does not already exist
     if await crud.fetch_one(accesses, {'login': login}) is not None:
         raise HTTPException(
             status_code=400,
             detail=f"An entry with login='{login}' already exists.",
         )
+
+
+async def post_access(login: str, password: str, scopes: str) -> AccessRead:
+    await check_for_access_login_existence(login)
     # Hash the password
     pwd = await security.hash_password(password)
     access = AccessCreation(login=login, hashed_password=pwd, scopes=scopes)
@@ -66,6 +75,10 @@ async def update_access(payload: AccessBase, access_id: int = Path(..., gt=0)):
     """
     Based on a access_id, updates information about the specified access
     """
+    if payload.login is not None:
+        updated_acccess = await crud.fetch_one(accesses, {"id": access_id})
+        if (updated_acccess is not None) and (updated_acccess["login"] != payload.login):
+            await check_for_access_login_existence(payload.login)
     return await crud.update_entry(accesses, payload, access_id)
 
 
--- a/src/app/api/routes/accesses.py
+++ b/src/app/api/routes/accesses.py
@@ -76,7 +76,7 @@ async def update_access(payload: AccessBase, access_id: int = Path(..., gt=0)):
     Based on a access_id, updates information about the specified access
     """
     if payload.login is not None:
-        updated_acccess = await crud.fetch_one(accesses, {"id": access_id})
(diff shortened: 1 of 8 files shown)


src/app/api/routes/devices.py
```diff
@@ -56,6 +57,13 @@ async def update_device(payload: DeviceIn, device_id: int = Path(..., gt=0)):
     """
     Based on a device_id, updates information about the specified device
     """
+    if payload.login is not None:
+        updated_device = await crud.fetch_one(devices, {"id": device_id})
```
frgfm, line 61: I'd suggest 
```
updated_device = await crud.get(device_id, devices)
```

src/app/api/routes/users.py
```diff
@@ -77,6 +86,14 @@ async def update_user(
     """
     Based on a user_id, updates information about the specified user
     """
+    # Check for access login
+    if payload.login is not None:
+        updated_user = await crud.fetch_one(users, {"id": user_id})
```
frgfm, line 91: I'd suggest:
```
updated_user = await crud.get(user_id, users)
```

src/app/api/routes/users.py
```diff
@@ -26,6 +27,14 @@ async def update_my_info(payload: UserInfo, me: UserRead = Security(get_current_
     """
     Updates information of the current user
     """
+    # Check for access login
+    if payload.login is not None:
+        updated_acccess = await crud.fetch_one(accesses, {"login": me.login})
+        await check_for_access_login_existence(payload.login)
+        if updated_acccess["login"] != payload.login:
+            await check_for_access_login_existence(payload.login)
```
frgfm, line 35: Perhaps I'm not following, but we don't need to run `await check_for_access_login_existence(payload.login)
` twice (the first one can be removed I think)

src/app/api/routes/accesses.py
```diff
@@ -2,19 +2,28 @@
...
+    return await crud.update_entry(accesses, Login(login=login), access_id)
+
+
+async def check_for_access_login_existence(login: str):
     # Check that the login does not already exist
     if await crud.fetch_one(accesses, {'login': login}) is not None:
         raise HTTPException(
             status_code=400,
             detail=f"An entry with login='{login}' already exists.",
         )
```
frgfm, line 22: I would suggest to return the fetched entry, you will be able to use that in devices & users

src/app/api/routes/devices.py
```diff
@@ -56,6 +57,13 @@ async def update_device(payload: DeviceIn, device_id: int = Path(..., gt=0)):
...
+        if updated_device is not None and payload.login != updated_device["login"]:
+            await check_for_access_login_existence(payload.login)
+            updated_acccess = await crud.fetch_one(accesses, {"login": updated_device["login"]})
```
frgfm, line 64: over here, if `check_for_access_login_existence` returns the entry, we can reduce overhead

src/app/api/routes/users.py
```diff
@@ -26,6 +27,14 @@ async def update_my_info(payload: UserInfo, me: UserRead = Security(get_current_
     """
     Updates information of the current user
     """
+    # Check for access login
+    if payload.login is not None:
+        updated_acccess = await crud.fetch_one(accesses, {"login": me.login})
+        await check_for_access_login_existence(payload.login)
```
frgfm, line 33: This can be refactored as well

frgfm (commented):
Thanks for the PR! A few refactoring suggestions in comments

src/app/api/routes/devices.py
florianriche: Unless I have overlooked, we don't want to alter the access object found by `check_for_access_login_existence` so there is no use in retrieving it. As a matter of fact, *IF* there is a found entry, then we should stop the function and raise an error.

src/app/api/routes/accesses.py
florianriche: Well we hope the fetched entry is None; otherwise an exception is raised.

src/app/api/routes/users.py
florianriche: Oups good catch

src/app/api/routes/devices.py
florianriche: Or are you suggesting to put the retrieval of the real updated_access into the ̀check_for_access_login_existencèfunction?

src/app/api/routes/accesses.py
frgfm: my bad, late review :sweat_smile:

src/app/api/routes/devices.py
frgfm: nevermind, I overlooked that we are fetching another ID! Sorry about that!

frgfm (approved):
Thank for the edits, looks good to me!

PR #85, opened 2020-12-03
BUGFIX: Events route accesses as device

florianriche:
The security of some events routes where decorated with get_current_user with expected scopes ["admin", "device"]. 
But get_urrent_user checks for the access authorization and then that a user exists with the required login. As we were trying to query the route with a device; no such user existed and hence an exception was raised.

In order to make the route works for both admin and device, we replace get_current_user by get_current_access

### DIFF
--- a/src/app/api/routes/events.py
+++ b/src/app/api/routes/events.py
@@ -3,14 +3,14 @@ from fastapi import APIRouter, Path, Security
 from app.api import crud
 from app.db import events
 from app.api.schemas import EventOut, EventIn
-from app.api.deps import get_current_user
+from app.api.deps import get_current_user, get_current_access
 
 
 router = APIRouter()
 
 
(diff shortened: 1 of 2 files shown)


fe51:
damn, I think I have introduced this bug !

fe51 (approved):

frgfm:
@florianriche does this mean we could also replace all `get_current_user` and `get_current_device` by `get_current_access`?

florianriche:
Wherever there is something like `_=Security(get_current_user / get_current_device, scopes=["whatever"])`; I would say yes because we are just checking the authorisation and not using the token to identify the requester.

On the other hand, when we use the output of `get_current_user` and `get_current_device`, I would suggest to leave as is because the code in those functions has to be executed.

To double triple check if we ever refactor though as I may have overlooked something.

PR #86, opened 2020-12-03
2 chars Bugfix: fix device route location in client

florianriche:
`Device` instead of `Devices`.

No other error typo observed.


frgfm (approved):
Nice catch!

PR #87, opened 2020-12-03
Qarnot bucket storage

### LINKED ISSUE #84: [media] Add a CSP service for actual media upload
The current media upload only updates the key/path of the media, while it does not upload content for the moment.
Qarnot computing support should be added as a supported bucket service.

Here is the documentation for it: https://computing.qarnot.com/documentation/sdk-python/api/storage/storage.html

martin1tab:
As stated in #84, the current media upload only updates the key/path of the media, while it does not upload content for the moment.

This PR aims to:
- support Qarnot bucket storage service and create all needed methods 

Closing this PR will close #84


src/app/services/bucket/dummyBucketService.py
```diff
@@ -1,15 +0,0 @@
-from app.services.bucket.baseBucketService import BaseBucketService
```
florianriche, line 1: I would keep it for test and debug purposes.

src/app/config.py
```diff
@@ -30,7 +30,8 @@
     raise ValueError(
         "Missing Credentials. Please set 'SUPERUSER_LOGIN' and 'SUPERUSER_PWD' in your environment variables")
 
-BUCKET_NAME: str = "mypyroneartest"
+QARNOT_TOKEN: str = os.getenv("QARNOT_TOKEN")
```
florianriche, line 33: Don't know how QARNOT works but wouldn't it be better to have login/password?
The token has probably an expiration date; plus it may be more cumbersome to retrieve the Token by hand on your machine.

How do you put the token in the environment variable? It must be set in the docker-compose to be able to use it in the docker container.

src/app/services/bucket/qarnotBucketService.py
```diff
@@ -0,0 +1,26 @@
...
+        media_bucket = bucket.Bucket(conn, bucket_name)
+        return await media_bucket.add_file(bucket_name, file_binary, bucket_key)
+
+    async def get_uploaded_file(self, bucket_name: str, bucket_key: str):
+        conn = connection.Connection(client_token=cfg.QARNOT_TOKEN)
+        media_bucket = bucket.Bucket(conn, bucket_name)
+        return await media_bucket.get_file(bucket_name, bucket_key)
+
+    async def fetch_bucket_files(self, bucket_name: str):
+        conn = connection.Connection(client_token=cfg.QARNOT_TOKEN)
```
florianriche, line 21: those two lines are used in every function. Maybe Create  a private function "connet_to_bucket" or "get_bucket" or something like that.
If we refactor the connection mechanism, it will be easier to propagate the changes.

src/app/services/bucket/qarnotBucketService.py
```diff
@@ -0,0 +1,26 @@
...
+        media_bucket = bucket.Bucket(conn, bucket_name)
+        bucket_local_copy_directory = 'bucket_local_copy'
```
florianriche, line 23: I don't understand what it does. There is no way to just get the file names? You have to download them?
Anyway, set the ̀bucket_local_copy_directory` as a config variable

florianriche (changes requested):
I am concerned by the authorization mechanism based on a token.

Otherwise, the service seems to be correctly implemented.

Did you test it?

src/app/services/bucket/qarnotBucketService.py
florianriche: Reading the documentation, I think we are interested in the `list_files` function.
At least that is how I thought about this `fetch_bucket_files` behavior. Maybe ̀fetch_bucket_filenames` was more appropriate. To discuss with the others on what we want to do.

src/app/services/bucket/qarnotBucketService.py
```diff
@@ -0,0 +1,26 @@
...
+
+from app.services.bucket.baseBucketService import BaseBucketService
+from app import config as cfg
+
+
+class QarnotBucketService(BaseBucketService):
+
+    async def upload_file(self, bucket_name: str, bucket_key: str, file_binary: bin):
+        conn = connection.Connection(client_token=cfg.QARNOT_TOKEN)
+        media_bucket = bucket.Bucket(conn, bucket_name)
```
florianriche, line 12: To discuss with the other but I think we will upload the files using the directories.
One could think that one device = one directory

That way if one want to list the files for one device, one could just look at the directory. It would prevent some unauthorized accesses.

src/app/config.py
martin1tab: I am mot sure that we have the possibility to do that with QARNOT; all examples I have seen work with token; I think the token does not expire but I could be regenerated on demand.

Yes, the token must be set in the env variable on Heroku. I didn't put in docker-compose to keep it private and since we don't need it on test/dev side. I am not really testing the buckets in the test; all methods are patched

src/app/services/bucket/qarnotBucketService.py
martin1tab: OK I will fix that

martin1tab:
> I am concerned by the authorization mechanism based on a token.
> 
> Otherwise, the service seems to be correctly implemented.
> 
> Did you test it?

I have tested it on a bucket that I have created on Qarnot and it seems working

src/app/services/bucket/qarnotBucketService.py
```diff
@@ -0,0 +1,21 @@
+from qarnot import connection, bucket
+import os
+
+from app.services.bucket.baseBucketService import BaseBucketService
+from app import config as cfg
+
+
+class QarnotBucketService(BaseBucketService):
+
+    def _connet_to_bucket(self, bucket_name: str):
```
florianriche, line 10: typo ;)

florianriche (dismissed):
It looks good to me but I'd rather have a second opinion from someone else

src/app/config.py
frgfm: This token is not supposed to expire, but on the contrary for security reason, it's much much better to use a token authentication. If the token gets compromised, we can very easily disable it. But if the CSP creds get compromised, we would be in big trouble.

I agree for deployment on Heroku, and for local usage, perhaps we should be exploring making an `.env` file at some point :thinking:

src/app/services/bucket/qarnotBucketService.py
```diff
@@ -0,0 +1,21 @@
...
+
+class QarnotBucketService(BaseBucketService):
+
+    @staticmethod
+    def connect_to_bucket(bucket_name: str):
+        conn = connection.Connection(client_token=cfg.QARNOT_TOKEN)
+        return bucket.Bucket(conn, bucket_name)
+
+    async def upload_file(self, bucket_name: str, bucket_key: str, file_binary: bin):
+        return await self.connect_to_bucket(bucket_name).add_file(bucket_name, file_binary, bucket_key)
```
frgfm, line 15: connection to the bucket should be done in the constructor in my opinion. And then we can use all qarnot python methods afterwards

src/app/services/bucket/qarnotBucketService.py
```diff
@@ -0,0 +1,21 @@
...
+
+    async def get_uploaded_file(self, bucket_name: str, bucket_key: str):
+        return await self.connect_to_bucket(bucket_name).get_file(bucket_name, bucket_key)
```
frgfm, line 18: same here

src/app/services/bucket/qarnotBucketService.py
```diff
@@ -0,0 +1,21 @@
...
+
+    async def fetch_bucket_filenames(self, bucket_name: str):
+        return await self.connect_to_bucket(bucket_name).list_files()
```
frgfm, line 21: same here

src/app/services/bucket/qarnotBucketService.py
frgfm: (that might mean changing the BaseBuckerService, but I believe it's necessary)

frgfm (commented):
Thanks for the PR! I added a few comments about some refactoring option!

src/app/services/bucket/qarnotBucketService.py
florianriche: Could be a good idea but isn't there a risk that the connection closes at some point? 
I don't how it works for Qarnot.

frgfm: Worst case scenario, we change it later with a `try except` but reconnecting all the time could throw an internal security warning for the Qarnot account as well

martin1tab: Yes I wanted to do it at the beginning but the issue is that if we establish the connection in the constructor we would need a valid token (the constructor is called and not patched). Until now, I am not using a token because the methods are all patched.
What do you think ?

martin1tab: Maybe i could store the connection in an attribute (using the static method I've created) and create the connection object only if it does not already exist. I could try this

frgfm:
By the way, @martin1tab would you mind adding a description to this PR?

src/app/services/bucket/qarnotBucketService.py
frgfm: We can also validate this PR, and open another to refactor this if this becomes cumbersome!
But I agree with your 2nd suggestion :ok_hand: That would be great!

martin1tab:
> By the way, @martin1tab would you mind adding a description to this PR?

Yes sure, done

src/app/config.py
```diff
@@ -30,7 +30,8 @@
...
+BUCKET_NAME: str = "pyronear-media-from-devices-1"
```
frgfm, line 34: Just noticed, what is the value supposed to represent? (it's not the bucket name of Qarnot)

frgfm (commented):
Thanks for the edits, just one last comment about the bucket name

src/app/config.py
martin1tab: Yes I am supposing that it is the bucket name of Qarnot; we could change it for sure; did you choose the name of the bucket on Qarnot ? the name has this form (with dashes) because it looks like Qarnot won't accept it otherwise (if I am not wrong)

frgfm: Well yes, it's the one I sent you in the usage instructions :sweat_smile: 
But, I just realized, this cfg param isn't used since the user passes the bucket name as a string to the method/constructor. So can you remove it please?

martin1tab: I'm not sure I've understood well, which constructor do you mean ?

frgfm: Respectively:

- my bad, so far it's not used in the constructor since this is an abstract class

- I'm not sure we understand each other: apart from direct access to the bucket, no one should be able to know the bucket information. They are only used so that the API knows where to upload the content. So either we do keep this cfg param using `os.getenv` (**the bucket name or the token should not be exposed on github**), but then it has to be used in your Qarnot bucket service, or we have to pass the cfg value at the top level (most likely in the route). 

My suggestion here:
```
BUCKET_NAME: str =os.getenv('BUCKET_NAME')
```

and in your qarnot bucket service implementation, remove all user inputs regarding the bucket name, and replace them with `cfg.BUCKET_NAME`.

We can set the value securely on Heroku, and for local interaction, we'll pass it as a user input I guess

martin1tab: OK I see, thanks !! I hope it is better now

frgfm (approved):
Thanks for the edits! Looks good to me, have you tried to upload a file with the actual creds?
If it worked, we can move forward with merge I guess!

martin1tab:
> Thanks for the edits! Looks good to me, have you tried to upload a file with the actual creds?
> If it worked, we can move forward with merge I guess!

Thanks !! yes I have succeeded to upload with the actual creds

src/app/services/bucket/qarnotBucketService.py
frgfm: I agree this might come helpful for devices, but let's explore this in another PR and the directory upload is not especially required for the incoming demo :)

PR #89, opened 2020-12-04
feat: Added an end-to-end script for API route testing

### LINKED ISSUE #65: [test] Add an end-to-end test script
Initially mentioned in #17, an end-to-end testing script would allow us to easily make sure that the API can perform all required tasks for a specific scope. Several "interaction maps" could be designed, but here is a suggestion for a basic one:
- [ ] create a user, a site
- [ ] login with this user
- [ ] update the password
- [ ] create a device of her/his own
- [ ] create an installation with this device and the site
- [ ] the installation creates an event and a related wildfire alert (+ a related media)
- [ ] the installation throws later a second alert that terminates the event

What do you think?

frgfm:
As per #65, and end-to-end script was needed to check that the sequence of requests is yielding the expected behaviour.
Merging this PR will close #65 

Any feedback is welcome!


.github/workflows/main.yml
```diff
@@ -71,3 +71,18 @@ jobs:
           file: ./coverage.xml
           flags: unittests
           fail_ci_if_error: true
+
+  end-to-end:
+    runs-on: ubuntu-latest
+    needs: dokcer-ready
```
florianriche, line 77: Typo I assume?

.github/workflows/main.yml
```diff
@@ -71,3 +71,18 @@ jobs:
...
+      - uses: actions/checkout@v2
+      - name: Build & run docker
+        run: PORT=8002 docker-compose up -d --build
+      - name: Set up Python 3.7
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.7
+          architecture: x64
+      - name: End-to-End test
+        run: python scripts/e2e_interactions.py 8002
```
florianriche, line 88: is it the right script name?

scripts/api_e2e.py
```diff
@@ -0,0 +1,122 @@
...
+    payload = dict(lat=44.1, lon=3.9, event_id=event_id, media_id=media_id, type='start')
+    alert_id1 = api_request('post', f"{api_url}/alerts/from-device", payload, device_auth)['id']
+
+    # Installation throws the end alert
+    payload = dict(lat=44.1, lon=3.9, event_id=event_id, type='end')
+    alert_id2 = api_request('post', f"{api_url}/alerts/from-device", payload, device_auth)['id']
+
+    # Cleaning (order is important because of foreign key protection in existing tables)
+    response = requests.delete(f"{api_url}/alerts/{alert_id1}/", headers=superuser_auth)
+    assert response.status_code // 100 == 2, print(response.json()['detail'])
```
florianriche, line 90: can't you use the api_request method above? you are already doing the assert checks in it.

scripts/api_e2e.py
```diff
@@ -0,0 +1,122 @@
...
+    # Installation creates an event & alert
+    payload = dict(lat=44.1, lon=3.9, type='wildfire')
+    event_id = api_request('post', f"{api_url}/events/", payload, superuser_auth)['id']
+
+    payload = dict(lat=44.1, lon=3.9, event_id=event_id, media_id=media_id, type='start')
+    alert_id1 = api_request('post', f"{api_url}/alerts/from-device", payload, device_auth)['id']
+
+    # Installation throws the end alert
+    payload = dict(lat=44.1, lon=3.9, event_id=event_id, type='end')
+    alert_id2 = api_request('post', f"{api_url}/alerts/from-device", payload, device_auth)['id']
```
florianriche, line 86: Do you see a way to use the upload route as well without sending a media effectively on the bucket?

scripts/api_e2e.py
```diff
@@ -0,0 +1,122 @@
...
+    assert response.status_code // 100 == 2, print(response.json()['detail'])
+    response = requests.delete(f"{api_url}/users/{user_id}/", headers=superuser_auth)
+    assert response.status_code // 100 == 2, print(response.json()['detail'])
+    print("SUCCESS")
+
+    return
+
+
+def parse_args():
+    import argparse
```
florianriche, line 107: import at the top of the doc?

.github/workflows/main.yml
frgfm: Correct, thanks!

scripts/api_e2e.py
frgfm: We don't pass any payload with delete methods, but I can perhaps change `api_request`

scripts/api_e2e.py
frgfm: Maybe if we create a media bucket file deletion route, or make it mandatory when a media gets deleted, that would make more sense. But with the current master version, I don't think we can

florianriche (approved):

PR #90, opened 2020-12-05
Crud refactoring & db transactions

jeanpasquier75:
As drafted in #46 this PR aims to split crud into submodules: 
* crud/base.py - legacy crud.py / still importable via `crud.X` through crud/\_\_init\_\_.py
* crud/accesses.py - legacy routes/accesses.py with some updates (see bellow)
* crud/alerts.py - alerts specific queries (so far _fetch_ongoing_alerts_)

crud.accesses uses database transaction to maintain a consistent state between users/devices & accesses tables. Its functions are used in both routes/users.py and routes/devices.py to avoid code duplication.

Fix devices/create_device: checks that owner_id exists (avoids a db foreign key error)) or raises 404 exception (case added in _test_create_device_invalid_).

I also moved routes testing under a tests/routes/ directory.


src/app/api/crud/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .base import *
+from . import accesses
+from . import alerts
```
frgfm, line 3: Shouldn't it be:
```
from .accesses import *
from .alerts import *
```
keeping the same syntax?

src/app/api/crud/accesses.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+async def delete_accessed_entry(table: Table, accesses: Table, entry_id: int):
+    """Delete an entry (User, Device, ...) and the corresponding access."""
+    # Ensure database consistency between tables with a transaction
+    async with base.database.transaction():
+        entry = await base.delete_entry(table, entry_id)
+        await base.delete_entry(accesses, entry["access_id"])
+
+    return entry
```
frgfm, line 99: Do we have to regroup the table + its access operation in the same function for the db transaction to be used?
It's better if we can tell easily what's going on from the route code when possible

src/app/api/routes/devices.py
```diff
@@ -57,29 +64,20 @@ async def update_device(payload: DeviceIn, device_id: int = Path(..., gt=0)):
...
-    await crud.delete_entry(accesses, entry['access_id'])
-    return entry
+    return await crud.accesses.delete_accessed_entry(devices, accesses, device_id)
 
 
-@router.get("/my-devices", response_model=List[DeviceOut],
-            summary="Get the list of all devices belonging to the current user")
+@router.get(
+    "/my-devices", response_model=List[DeviceOut], summary="Get the list of all devices belonging to the current user"
+)
```
frgfm, line 80: I don't really get which formatter you use :sweat_smile:  Sometimes it looks like black, sometimes it doesn't haha

src/app/db/tables.py
```diff
@@ -27,7 +27,7 @@
     "accesses",
     metadata,
     Column("id", Integer, primary_key=True),
-    Column("login", String(50), unique=True),
+    Column("login", String(50), unique=True, index=True),  # index for fast lookup
```
frgfm, line 30: does this mean that it won't be fast to access using the primary key?

frgfm (commented):

src/app/api/crud/__init__.py
jeanpasquier75: The 2 approaches have pros & cons: your suggestion implies we cannot have the same function name in 2 different subpackages because there would be a conflict here. Also I think it is clearer to use `crud.alerts.{something I know is related to alerts}`, but it is just my opinion ;)

src/app/api/crud/accesses.py
jeanpasquier75: I am not sure to understand your question haha

src/app/api/routes/devices.py
jeanpasquier75: I think that I reformatted the file with black then I rollbacked some of the functions I had not modified to make the review easier. This one look like black as it often turn the code into 
```
def/call function(
   many, many, many, parameters, here
)
``` 

instead of something like this for some IDEs
```
def/call function(many, many, many
                  parameters, here)
```

I do prefer the 1st formatting 😉

src/app/db/tables.py
jeanpasquier75: It won't slow down the access via primary key, it will maintain a secondary index (it means a little bit more storage) but allows "random" access (opposite of sequential). I am investigating about difference between unique contraint and unique index for postgres and sqlalchemy, terms are quite confusing

src/app/api/crud/__init__.py
frgfm: Fair enough, I just thought that since `base` had all the "shared" operations, each submodule would have functions name differently, but I'm fine keeping this as well!

src/app/api/crud/accesses.py
frgfm: Sorry :sweat_smile: 
I meant, if possible, having 2 separate instructions in the route code for the route table operation, and for its access table operation would be nice. Here you regrouped both ops in a function called by the route, which in the end, means that we don't see in the route code which tables get modified

src/app/db/tables.py
frgfm: Thanks!

frgfm:
You might need to merge master as well to solve conflicts btw :/

jeanpasquier75:
> You might need to merge master as well to solve conflicts btw :/

Are you telling me that some PR are merged much faster than mine ? 🤣 
I will, thanks

src/app/api/crud/accesses.py
jeanpasquier75: Ok I see, that's what I planned to do with this db.transaction decorator on each route but I had some issues during the tests and it was not a good practice (a transaction must be closed asap) and then I figured out that there was a lot of code duplication between users & devices: putting the same operations in single functions with some parameters should make it easier to maintain. At least we have the tables in argument (users or devices and accesses) so we know what tables are involved from routes !

frgfm: yes, I was just nitpicking! We can easily think about a refactoring option later!

frgfm:
Thanks for the merge, I'll get my head back into this by the end of the weekend to make a proper review and make sure I didn't miss anything :ok_hand:

frgfm (approved):
Thanks for the last edits!

PR #93, opened 2020-12-07
feat: Added alert acknowledgement option in client

### LINKED ISSUE #92: [client] Add alert acknowledgement to the client
Hello there :wave: 

Following up on https://github.com/pyronear/pyro-platform/issues/27, perhaps we should add right away the possibility of acknowledging an alert in the client. I think we will need to think further later about which scope/access should be required to used it, but for now we can restrict it to admin?

What do you think @florianriche @jeanpasquier75 @fe51 @martin1tab ?

frgfm:
As per #92, the API client does not enable alert acknowledgement. This PR adds support of this feature, a corresponding unittest, and extends end-to-end script.

Merging this PR successfully will close #92 

Any feedback is welcome!


Dockerfile_bis
```diff
@@ -0,0 +1,20 @@
+FROM tiangolo/uvicorn-gunicorn-fastapi:python3.8-alpine3.10
```
florianriche, line 1: Why this dockerfile bis?

client/test/test_client.py
```diff
@@ -24,10 +26,23 @@ def test_client(self):
...
         self._test_route_return(api_client.get_my_devices(), list)
         self._test_route_return(api_client.get_sites(), list)
-        self._test_route_return(api_client.get_all_alerts(), list)
+        all_alerts = self._test_route_return(api_client.get_all_alerts(), list)
         self._test_route_return(api_client.get_ongoing_alerts(), list)
         self._test_route_return(api_client.get_unacknowledged_alerts(), list)
 
+        # Acknowledge an alert
+        alert_id = None
+        # Try to find an unacknowledged alert
```
florianriche, line 35: WOuldn't it be easier to just create one and then acknowledge it ?

florianriche (commented):
Looks good besides my questions in comments

Dockerfile_bis
frgfm: oh my bad, I pushed something I didn't plan on pushing! I'll remove it!

client/test/test_client.py
frgfm: I was lazy to create device + event (foreign key lock), but I just updated it after your comment!

client/pyroclient/client.py
```diff
@@ -27,6 +27,7 @@
     "get-alerts": "/alerts",
     "get-ongoing-alerts": "/alerts/ongoing",
     "get-unacknowledged-alerts": "/alerts/unacknowledged",
+    "acknowledge-alert": "/{alert_id}/acknowledge",
```
florianriche, line 30: you need to put "alerts/{alert_id}/acknowledge"

client/test/test_client.py
```diff
@@ -22,12 +24,21 @@ def test_client(self):
...
         self._test_route_return(api_client.get_ongoing_alerts(), list)
         self._test_route_return(api_client.get_unacknowledged_alerts(), list)
 
+        if len(all_devices) > 0:
+            # Create event
+            event_id = self._test_route_return(api_client.create_event(0., 0.), dict)['id']
+            # Create an alert
+            alert = self._test_route_return(api_client.send_alert(0., 0., event_id, all_devices[0]['id']), dict)
+            # Acknowledge it
+            updated_alert = self._test_route_return(api_client.acknowledge_alert(alert['id']), dict)
```
florianriche, line 39: considering the error I pointed above, I am surprised this test pass

florianriche (changes requested):

client/pyroclient/client.py
frgfm: fixed!

client/test/test_client.py
frgfm: agreed, I'm surprised as well, but it did pass locally for me :man_shrugging:

florianriche (approved):

PR #95, opened 2020-12-07
fix: Fixed Qarnot image upload and image retrieval

florianriche:
Upload routes were not working properly with Qarnot:

1. For the upload: Qarnot doesn't accept an int as a bucket key. Had to convert the hash to a string
2. For the image URL retrieval: Qarnot retrieves the bucket objects by downloading them locally and then simply returning the filepath of the object. It seems here is not way to retrieve a public URL. Hence I didn't touch this part, but we have to be aware it is malfunctioning with Qarnot as it will only return the filename (i.e the bucket_key, which is not very useful)
3. For the image retrieval as Bytes: Used to try to read a URL. Now it reads the Bucket File downloaded by Qarnot, uses it in the Response and launch a Background task to delete that file after the route response..

### DIFF
--- a/src/app/api/routes/media.py
+++ b/src/app/api/routes/media.py
@@ -1,16 +1,13 @@
-from fastapi import APIRouter, Path, Security, File, UploadFile, HTTPException
+from fastapi import APIRouter, Path, Security, File, UploadFile, HTTPException, BackgroundTasks
 from fastapi.responses import StreamingResponse
 from typing import List
 from datetime import datetime
-import requests
-import io
 
 from app.api import crud
 from app.db import media
 from app.api.schemas import MediaOut, MediaIn, MediaCreation, MediaUrl, DeviceOut, BaseMedia
 from app.api.deps import get_current_device, get_current_user
 from app.services import bucket_service
-import app.config as cfg
 
 router = APIRouter()
 
@@ -93,7 +90,7 @@ async def upload_media(media_id: int = Path(..., gt=0),
     Upload a media (image or video) linked to an existing media object in the DB
     """
     entry = await check_for_media_existence(media_id, current_device.id)
-    bucket_key = hash(datetime.utcnow())
+    bucket_key = str(hash(datetime.utcnow()))
 
     upload_success = await bucket_service.upload_file(bucket_key=bucket_key,
                                                       file_binary=file.file)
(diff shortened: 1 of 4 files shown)


src/app/services/bucket/qarnotBucketService.py
```diff
@@ -13,10 +14,23 @@ def connect_to_bucket(self):
...
+            self.connect_to_bucket().add_file(file_binary, bucket_key)
+        except Exception as e:
+            print(e)
+            return False
+        return True
 
     async def get_uploaded_file(self, bucket_key: str):
-        return await self.connect_to_bucket().get_file(bucket_key)
+        try:
+            return self.connect_to_bucket().get_file(bucket_key)
```
martin1tab, line 26: No need for await ?

martin1tab (approved):

src/app/services/bucket/qarnotBucketService.py
florianriche: I got an error stating if I set it. as get_file returns a string a got a `object str can't be used in 'await' expression`

src/app/api/routes/media.py
```diff
@@ -93,7 +90,7 @@ async def upload_media(media_id: int = Path(..., gt=0),
     Upload a media (image or video) linked to an existing media object in the DB
     """
     entry = await check_for_media_existence(media_id, current_device.id)
-    bucket_key = hash(datetime.utcnow())
+    bucket_key = str(hash(datetime.utcnow()))
```
frgfm, line 93: Good catch :+1:

src/app/services/bucket/qarnotBucketService.py
```diff
@@ -13,10 +14,23 @@ def connect_to_bucket(self):
         return self.bucket
 
     async def upload_file(self, bucket_key: str, file_binary: bin):
-        return await self.connect_to_bucket().add_file(file_binary, bucket_key)
+        try:
+            self.connect_to_bucket().add_file(file_binary, bucket_key)
+        except Exception as e:
+            print(e)
```
frgfm, line 20: Perhaps a warning or logging would be better for the exception?

frgfm (approved):
Thanks for the quick fix! Just a few questions, but nothing important, happy to merge if you already have the answers

src/app/services/bucket/qarnotBucketService.py
florianriche: Like 
```python
import logging
logging.basicFongi()
logging.error(e) ? 
```

frgfm: Yup! But regarding the logger choice, I'd recommend the uvicorn logger
```
logger = logging.getLogger("uvicorn.error")
```

frgfm (approved):
Thanks a lot for this save! Can you confirm whether it solved the problem on your end?

florianriche:
yes, on my sied I managed to upload and retriee Qarnot images using  swagger.

florianriche:
The clients tests and e2e are failing though.

frgfm:
> The clients tests and e2e are failing though.

The e2e test seems to be working, but the failing of the client test is not related! We can move forward with merging I think :+1:

PR #96, opened 2020-12-08
fix: handle failing upload case in media retrieval routes

florianriche:
Following some errors in sentry, it appears we have some errors if the image retrieval on qarnot was not successful.

There may be additional fixes to make, let's put them in this PR.


frgfm:
@florianriche the PR looks good, but I'm not sure if you want to tackle all recent issues here? Or do you want to make separate PRs?

florianriche:
At first I thought that the sentry issues were coming from more than one bug.
On second thought it seems this PR fix the recent issues. I think we can merge it.

frgfm (approved):
Thanks for the fix!

PR #98, opened 2020-12-08
feat: Upgraded bucket file naming with SHA256

frgfm:
This PR addresses the naming of bucket file on the CSP by:
- switching from hashing the timestamp to hashing the file content with SHA256
- taking the 32 first characters of the hexadecimal digested string and concatenate it with the file extension
- adding a unittest for the hashing function

Any feedback is welcome!

### DIFF
--- a/src/app/api/routes/media.py
+++ b/src/app/api/routes/media.py
@@ -2,7 +2,6 @@ from fastapi import APIRouter, Path, Security, File, UploadFile, HTTPException, 
 from fastapi.responses import StreamingResponse
 from typing import List
 from datetime import datetime
-import hashlib
 
 from app.api import crud
 from app.db import media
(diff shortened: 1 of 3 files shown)


src/app/api/routes/media.py
```diff
@@ -90,7 +92,9 @@ async def upload_media(media_id: int = Path(..., gt=0),
     Upload a media (image or video) linked to an existing media object in the DB
     """
     entry = await check_for_media_existence(media_id, current_device.id)
-    bucket_key = str(hash(datetime.utcnow()))
+
+    # Concatenate the first 32 chars of SHA256 hash with file extension
```
florianriche, line 96: Maybe mention that 32 is a bit of  a "random" threshold but it should be enough and it will prevent us from meeting issues with some systems interactions.

florianriche (dismissed):

src/app/api/routes/media.py
frgfm: done!

florianriche (approved):
Looks good to me :)

PR #99, opened 2020-12-08
feat: Added admin scopes on tables to protect

frgfm:
Following up on #45, I realized I was able to delete all entries in the alerts & media tables without any authentication. So this PR adds admin scope where it's best to start protecting the API.

Any feedback is welcome!


fe51:
Thanks for the PR !

This is better than nothing ! Is actual tests of Sébastien and Mateo are performed with admin scopes ?

fe51 (approved):

frgfm:
> Thanks for the PR !
> 
> This is better than nothing ! Is actual tests of Sébastien and Mateo are performed with admin scopes ?

They won't be impacted as the routes I modified are not exposed by the client yet!

jeanpasquier75 (approved):

PR #100, opened 2020-12-08
fix: Fixed media content upload to bucket

frgfm:
As witnessed over the past day, we've had some troubles with media upload. Some were fixed in #95 and #96 fortunately, but two issues remained: 
- adding the subfolder in the bucket name was not supported by the CSP client
- empty file could be uploaded, if the byte position of a file were to be offset for instance

This PR fixes both issues by adding a generalized bucket key resolution and resetting the byte position after hashing to avoid uploading empty files. I tried it on my end by deploying the feature branch on Heroku, and it works:
qarnot_bucket

Feel free to try by yourself before the next deployment :ok_hand: 

Any feedback is welcome!

### DIFF
--- a/.github/workflows/client-main.yml
+++ b/.github/workflows/client-main.yml
@@ -55,15 +55,15 @@ jobs:
             ${{ runner.os }}-deps-${{ env.cache-name }}-
             ${{ runner.os }}-deps-
             ${{ runner.os }}-
+      - name: Build & run API server
+        run: |
+          PORT=8002 docker-compose up -d --build
+          docker ps
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
           pip install -e client/
           pip install coverage
-      - name: Build & run API server
-        run: |
-          PORT=8002 docker-compose up -d --build
-          docker ps
       - name: Run client unittests
         run: |
           coverage --version
--- a/.github/workflows/client-main.yml
+++ b/.github/workflows/client-main.yml
@@ -67,6 +67,7 @@ jobs:
       - name: Run client unittests
         run: |
           coverage --version
+          sleep 5
           cd client && coverage run -m unittest discover test/
 
   flake8-py3:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -78,7 +78,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Build & run docker
-        run: PORT=8002 docker-compose up -d --build
+        run: |
+          PORT=8002 docker-compose up -d --build
+          docker ps
       - name: Set up Python 3.7
         uses: actions/setup-python@v1
         with:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -91,4 +91,6 @@ jobs:
           python -m pip install --upgrade pip
           pip install -r scripts/requirements.txt
       - name: End-to-End test
-        run: python scripts/api_e2e.py 8002
+        run: |
+          sleep 5
+          python scripts/api_e2e.py 8002
--- a/src/app/api/routes/media.py
+++ b/src/app/api/routes/media.py
@@ -98,7 +98,7 @@ async def upload_media(media_id: int = Path(..., gt=0),
     bucket_key = f"{hash_content_file(file.file.read())[:32]}.{file.filename.rpartition('.')[-1]}"
     # If files are in a subfolder of the bucket
     if "/" in cfg.BUCKET_NAME:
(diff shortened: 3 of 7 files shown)


src/app/api/routes/media.py
```diff
@@ -94,7 +94,11 @@ async def upload_media(media_id: int = Path(..., gt=0),
...
     # Concatenate the first 32 chars (to avoid system interactions issues) of SHA256 hash with file extension
-    bucket_key = f"{hash_content_file(file.file.read())[:32]}.{file.filename.rpartition('.')[-1]}"
+    file_name = f"{hash_content_file(file.file.read())[:32]}.{file.filename.rpartition('.')[-1]}"
+    # Reset byte position of the file (cf. https://fastapi.tiangolo.com/tutorial/request-files/#uploadfile)
+    await file.seek(0)
+    # If files are in a subfolder of the bucket, prepend the folder path
+    bucket_key = prepend_bucket_folder(file_name)
 
     upload_success = await bucket_service.upload_file(bucket_key=bucket_key,
                                                       file_binary=file.file)
```
florianriche, line 104: Au passage, can you also fix the return message below?
            detail="The upload did not succeed"
==>     detail="The download did not succeed"

florianriche (approved):
Looks good to me. Just a small additionnal fix to include in this PR while we are at it.

src/app/api/routes/media.py
frgfm: Not sure I follow: here we're actually trying to upload. Did you mean in `get_media_image`?

florianriche: Yes,(for some reason I can't comment directly to point this line in the Github interface)

frgfm: Done!

florianriche (dismissed):
Looks good! 
It's a bit unsettling because when I tried it on my side, I didn't see a new media appearing in Qarnot but that's because I was uploading the same image. But it worked!

florianriche (dismissed):

.github/workflows/main.yml
```diff
@@ -78,7 +78,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Build & run docker
-        run: PORT=8002 docker-compose up -d --build
+        run: |
+          PORT=8002 docker-compose up -d --build
+          docker ps
```
florianriche, line 83: docker ps is required?

.github/workflows/main.yml
```diff
@@ -89,4 +91,6 @@ jobs:
           python -m pip install --upgrade pip
           pip install -r scripts/requirements.txt
       - name: End-to-End test
-        run: python scripts/api_e2e.py 8002
+        run: |
+          sleep 5
```
florianriche, line 95: Why the sleep?

.github/workflows/client-main.yml
```diff
@@ -55,18 +55,19 @@ jobs:
...
           pip install -e client/
           pip install coverage
-      - name: Build & run API server
-        run: |
-          PORT=8002 docker-compose up -d --build
-          docker ps
       - name: Run client unittests
         run: |
           coverage --version
+          sleep 5
```
florianriche, line 70: why the sleep?

.github/workflows/client-main.yml
```diff
@@ -55,18 +55,19 @@ jobs:
             ${{ runner.os }}-deps-${{ env.cache-name }}-
             ${{ runner.os }}-deps-
             ${{ runner.os }}-
+      - name: Build & run API server
+        run: |
+          PORT=8002 docker-compose up -d --build
+          docker ps
```
florianriche, line 61: why the docker ps ?

frgfm:
Added a unittest for bucket key resolution (after all, I introduced a feature, this should be tested)
And additionally, I think I found the problem with `ConnectionError`: the tests were run right after the docker started, while the FastAPI server needs some time to run and become reachable :sweat_smile: 
I added a workflow fix (moving the docker start as early as possible, and add a sleep before running the tests) and it seems to be doing the trick :raised_hands:

.github/workflows/main.yml
frgfm: cf. my comment :)

.github/workflows/main.yml
frgfm: cf. my comment :)

.github/workflows/client-main.yml
frgfm: same

.github/workflows/client-main.yml
frgfm: same

florianriche (approved):

Issue #94, opened 2020-12-07
[media] It is possible to upload multiple images for one media and loose the bucket keys


florianriche:
When one upload an image linked to a media (route upload_media) ; we:
1: Generate the CSP bucket key
2: Upload the image
3: If successful, alter the bucket key field in the DB

If one calls this function two times for a unique media, then the first uploaded image will stay in the CSP and we will loose its "key". It won't be linked to anything anymore. If for some reason, we do this too often there will be a load on the bucket storage.

frgfm:
Agreed, I would have a simple suggestion for the key:
- either a concat of device_id and timestamp
- or the better option, the SHA256 hash of the content file

Fully agree that the update should only be made if the upload is successful!

florianriche:
We already do the bucket_key upload only if the upload was successfful.

```python
entry = await check_for_media_existence(media_id, current_device.id)
bucket_key = str(hash(datetime.utcnow()))

upload_success = await bucket_service.upload_file(bucket_key=bucket_key,
                                                      file_binary=file.file)
if upload_success is False:
        raise HTTPException(
            status_code=500,
            detail="The upload did not succeed"
        )
entry = dict(**entry)
entry["bucket_key"] = bucket_key
return await crud.update_entry(media, MediaCreation(**entry), media_id)

```

Agreed for the hashing part.

My point is a bit different, if you read the code above, you will see we don't check there isn't already an uploaded image for a media. If we run the route three times for one unique media, we will have three different images in the CSP. 
Don't know exactly what would be the desired behavior. Overwriting?

frgfm:
Mmmh, if we have only the content and not the timestamp, I guess we can check if the media content already exists on the bucket?

Actually, this might be a more complex problem: the upload can succeed (ie media content is valid and in the bucket) and we try to overwrite it, but the upload can also fail explicitly (connection issue or something similar) or silently (data corruption during transfer).

So what we could do when we upload a media:
- check if there is a bucket key existing in the table
- if there is, we need to decide whether overwrite is allowed or if we do nothing
- if there isn't, we compute the hashed bucket key
- check if the bucket has a file named the same way
- if there is, we again need to make the same decision
- if there isn't, we upload the file
- once uploaded, we could also try to dl the file, and hash the downloaded version to verify data integrity

As a first step, we could "do nothing" when the file is already in the bucket.

frgfm:
Thinking back about that issue, now that data integrity is fully deployed:
[...]

PR #102, opened 2020-12-09
feat: Added integrity check and optimized bucket interactions

### LINKED ISSUE #97: [media] Reduce image download latency on client
Currently, when requesting an image through the client:
- the API accesses the media table to get the bucket path, and adds the CSP creds
- it downloads the image on the API server
- the client then download the byte content from the API server

So we at least get a double latency on content transfer. A more optimal solution would be:
- the API resolves the URL + creds for a CSP file download
- the client uses this information to directly download from CSP

We could expect ~50% drop in latency in doing so!

cc @florianriche @Akilditu

frgfm:
Following up on #94, this PR introduces a few features related to media/bucket interactions:
- removed legacy BucketService classes (if we add support of another CSP, we have no guarantee we won't need additional arguments in the methods, and )
- refactored QarnotBucket: bucket connection in the constructor for early validation of creds, added file deletion method, optimized boto3 object interactions, changed errors to warning for sentry to report more detailed situations
- renamed `check_for_media_existence` into `check_media_registration`
- optimized URL retrieval route by only checking if the file exists in the bucket instead of downloading it
- extended media upload route with suggestion from #94 (cf. below)
- actually made url retrieval route to return a publicly available URL (which will close #97)
- deleted media/content reading route

Regarding the media upload, here is what this PR implements:
1. Each media upload to the bucket gets a data integrity check: we download back the file temporarily and hash it and check it against our source hash.
2. Before attempting an upload, we check if the media entry already has the same bucket key (by hashing the file). If the hashes are the same, we don't upload/overwrite (since this means it's the exact same file), otherwise we proceed with upload.

I deployed the branch on Heroku and tried on my hand, it seems to be working quite well! This is expected to be reducing bucket interaction load and ensure data integrity during transfer.

Any feedback is welcome!


src/app/api/routes/media.py
```diff
@@ -85,64 +86,101 @@ async def delete_media(media_id: int = Path(..., gt=0)):
...
-        )
-    entry = dict(**entry)
-    entry["bucket_key"] = bucket_key
-    return await crud.update_entry(media, MediaCreation(**entry), media_id)
+    # Upload if bucket_key is different (otherwise the content is the exact same)
+    if isinstance(entry['bucket_key'], str) and entry['bucket_key'] == bucket_key:
+        return await crud.get_entry(media, media_id)
+    else:
+        # Failed upload
+        if not await bucket_service.upload_file(bucket_key=bucket_key, file_binary=file.file):
```
florianriche, line 115: If there is already an existing media, we should try to delete it from the bucket don't you think?

src/app/api/routes/media.py
```diff
@@ -85,64 +86,101 @@ async def delete_media(media_id: int = Path(..., gt=0)):
...
+        if uploaded_file is None:
+            raise HTTPException(
+                status_code=500,
+                detail="The data integrity check failed (unable to download media form bucket)"
+            )
+        # Remove temp local file
+        background_tasks.add_task(bucket_service.flush_tmp_file, uploaded_file)
+        # Check the hash
+        with open(uploaded_file, 'rb') as f:
+            upload_hash = hash_content_file(f.read())
```
florianriche, line 132: would suggest
```suggestion
            upload_hash = hash_content_file(open(uploaded_file, 'rb').read())
```

frgfm: Well, here we do want to close the file after hashing, don't we?

src/app/api/routes/media.py
frgfm: I thought about it and figured we should deal with this in another PR (it raises the question of what we need to hash: if 2 different users upload the exact same file, currently both their media will point to the same bucket file)

florianriche:
Shall we merge the PR only after the data4Good demo? I think we said we would freeze the code before the demo

jeanpasquier75 (commented):
I let my first comments about Python stuff, I need to deep dive some more to review functional features

src/app/services/bucket/qarnot.py
```diff
@@ -0,0 +1,84 @@
...
+__all__ = ['QarnotBucket']
+
+
+logger = logging.getLogger("uvicorn.warning")
+
+
+class QarnotBucket:
+
+    _bucket: Optional[Bucket] = None
+    _media_folder: Optional[str] = None
```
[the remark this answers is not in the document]
frgfm: It was more for a clearer code layout since the constructor initializes the value, but it calls a function that needs to check if the values are none.
Their default values are shared across all instances, but each instance can have different values for this attribute since they are set in the constructor. Here is an example:
```
In [1]: class A:
   ...:     a = None
   ...:     def __init__(self, val, ref=2):
   ...:         if val > ref:
   ...:             self.a = val
   ...: 

In [2]: a = A(0)

In [3]: b = A(3)

In [4]: c = A(4)

In [5]: print(a.a, b.a, c.a)
None 3 4
```

src/app/services/bucket/qarnot.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def __init__(self) -> None:
+        self._connect_to_bucket()
+
+    def _connect_to_bucket(self) -> None:
+        """Connect to the CSP bucket"""
+        conn = Connection(client_token=cfg.QARNOT_TOKEN)
+        self._bucket = Bucket(conn, cfg.BUCKET_NAME.rpartition("/")[0])
+        if len(cfg.BUCKET_NAME) > cfg.BUCKET_NAME.find("/") + 1:
+            self._media_folder = cfg.BUCKET_NAME[cfg.BUCKET_NAME.find('/') + 1:]
```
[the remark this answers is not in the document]
frgfm: Agreed, I just fixed this by splitting the env vars!

src/app/services/bucket/qarnot.py
```diff
@@ -0,0 +1,84 @@
...
+        try:
+            # Filter the bucket summary if this is in a subfolder
+            if isinstance(self._media_folder, str):
+                obj_summary = self.bucket.directory(self._media_folder)
+            else:
+                obj_summary = self.bucket.list_files()
+            return len(list(obj_summary.filter(Prefix=bucket_key))) > 0
+        except Exception as e:
+            logger.warning(e)
+            return False
```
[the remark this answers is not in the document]
frgfm: Just renamed it to `check_file_existence` to avoid turning it into a question!

About the HeadObject, I do not have enough experience with that, could you make a more explicit suggestion please?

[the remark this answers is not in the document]
frgfm: Found what you were suggesting and just pushed the implementation :)

frgfm:
Note to reviewers: following up on @jeanpasquier75 suggestion about HeadObject, I managed to get, through Qarnot python sdk, the underlying AWS S3 object. With this I did two things:
- optimized bucket file existence check
- added a method to generate temporary (expiration in 1 hour) public URL of bucket file

Thus I deleted the media content reading route since it's the main bottleneck in terms of latency among routes, and this can be achieved on the client side (which I updated)

src/app/api/routes/media.py
```diff
@@ -85,64 +86,72 @@ async def delete_media(media_id: int = Path(..., gt=0)):
...
-    retrieved_file = await bucket_service.get_uploaded_file(bucket_key=media["bucket_key"])
-    if retrieved_file is False:
-        raise HTTPException(
-            status_code=500,
-            detail="The download did not succeed"
-        )
-    background_tasks.add_task(bucket_service.flush_after_get_uploaded_file, retrieved_file)
-    return StreamingResponse(open(retrieved_file, 'rb'), media_type="image/jpeg")
+async def get_media_url(
+    background_tasks: BackgroundTasks,
```
florianriche, line 148: background task is no longer used.

src/app/services/bucket/baseBucketService.py
```diff
@@ -1,22 +0,0 @@
-from abc import ABC
-from abc import abstractmethod
-
```
florianriche, line 3: I would have kept it and dummyBucketService.
But let's not argue and let's move on.
OK for the deletion.

florianriche (approved):
Looks good to me beside one unused param.
I will approve again after you fix it.

src/app/services/bucket/qarnot.py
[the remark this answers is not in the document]
jeanpasquier75: Yes you're right (the fact of sharing variable might become dangerous when it concern mutable types)

PR #104, opened 2020-12-12
Add security admin scope requirement on accesses routes

jeanpasquier75:
Accesses routes were accessible without any security requirement. Added **admin** scope requirement to all accesses routes


fe51 (approved):

fe51:
Thanks for the quick PR !

src/app/api/routes/accesses.py
```diff
@@ -1,28 +1,31 @@
...
 from app.db import accesses
 from app.api.schemas import AccessBase, AccessRead, AccessAuth, AccessCreation, Cred, CredHash, Login
+from app.api.deps import get_current_access
 
 
 router = APIRouter()
 
 
-async def update_access_login(login: str, access_id: int):
+async def update_access_login(login: str, access_id: int, _=Security(get_current_access, scopes=["admin"])):
```
frgfm, line 12: Isn't this dangerous?
Considering update_access_login can be used by other routers I mean. We could update the scope of the routes that uses this no ? (so that we only have scope constraints on routes and not functions that can be used within a route, which could generate cumbersome errors if we accept the request but reject the scope midway)

frgfm (commented):
Thanks for the PR! Small question/comment about where the scope constraint of the functions should be made, but the scopes on pure routes are fine by me!

src/app/api/routes/accesses.py
jeanpasquier75: You are right the dependency at function level (not route) seems to have no impact: the route PUT device/{device_id}/ calls `update_access_login` without requiring admin scope. The logic of rejection on unauthorised request must be implemented at route level (devices and users) as mentioned in https://github.com/pyronear/pyro-api/issues/45

frgfm: Mind updating the PR accordingly? :pray:

frgfm: Any update? :grin:

frgfm (commented):
Thanks, the codebase looks perfect to me now! Could you add a short unittest to test the behaviour please?

frgfm:
@jeanpasquier75 ? :pray:

frgfm:
Strange, I just checked, the unittests for accesses routes should fail. In the mock access table, there are only accesses with "me" scope. But the tests do pass :thinking:

jeanpasquier75:
> Strange, I just checked, the unittests for accesses routes should fail. In the mock access table, there are only accesses with "me" scope. But the tests do pass 🤔

I do agree it is weird ahah, I am not really familiar with the mocking access part

frgfm (approved):
Since #119 was opened, let's move on with this PR!

PR #112, opened 2021-02-22
Database migration with alembic

### LINKED ISSUE #107: [database] Plan a migration process for the DB
There is no migration process with FastAPI or SQLAlchemy by default, we need to investigate our options here.

jeanpasquier75:
Database migration implementation using Alembic tool: alter database schema (create table, add column, etc.) using SQLAlchemy syntax, more info in `src/alembic/README.md`

- src/alembic.ini is configuration (almost default one, generated by alembic)
- src/alembic/ 
  - versions/ contains migration scripts (so far empty)
  - env.py alembic migration skeletton (default adapted to retrieve database URL from the app config)
  - README.md alembic generated an empty readme, first draft here
  - script.py.mako script generation template (default one)

Open questions:
- alembic added to requirements.txt, afterward maybe it should be in requirements-dev.txt since it concerns CI/CD phase and not running docker image ?
- add a Github workflow to run the migration in production, only once it is merged to master ?
- how to test the scripts before merging to master ? There are libraries such as [pytest-alembic](https://pypi.org/project/pytest-alembic/)
- encode (maintainers of async database package) use another but similar method to set database url `config.set_main_option` in [there doc](https://www.encode.io/databases/tests_and_migrations/ )

resolves #107


requirements.txt
```diff
@@ -2,6 +2,7 @@ fastapi==0.61.1
 uvicorn>=0.11.1
 databases[postgresql]>=0.2.6,<=0.4.0
 SQLAlchemy>=1.3.12
+alembic==1.5.4
```
frgfm, line 5: I agree that this should be moved to dev requirements!

Dockerfile
```diff
@@ -6,6 +6,8 @@ WORKDIR /usr/src/app
 # set environment variables
 ENV PYTHONDONTWRITEBYTECODE 1
 ENV PYTHONUNBUFFERED 1
+ENV PYTHONPATH "${PYTHONPATH}:/app"
+
```
frgfm, line 10: If I'm not mistaking, alembic lets you pass the path to the .ini file. Perhaps it would be better?

src/alembic.ini
```diff
@@ -0,0 +1,84 @@
+# A generic, single database configuration.
+
+# TODO adapt if needed
```
frgfm, line 3: Not sure the comment is required

src/alembic/README.md
```diff
@@ -0,0 +1,22 @@
+[comment]: <> (TODO write migration guide/contributing with docker-compose commands)
```
frgfm, line 1: Can be removed

.github/workflows/main.yml
```diff
@@ -102,3 +102,15 @@ jobs:
...
+
+#  db-migration:
+#    runs-on: ubuntu-latest
+#    needs: end-to-end
+#    steps:
+#      - name: Run database migration
+#        env:
+#          DATABASE_URL: ${{ secrets.DATABASE_URL }}
+#        run: |
+#          alembic upgrade head
```
frgfm, line 116: We just need to test that this would work on an existing DB. Perhaps we could check with our unittests

frgfm (commented):
Thanks for this PR! I added a few comments, but I have no experience with alembic

Dockerfile
jeanpasquier75: The environment variable `PYTHONPATH` allows to make python modules importable. This way `alembic/env.py` can import app/config and app/db from outside the `app/` path. What do you think ?

src/alembic/README.md
jeanpasquier75: This TODO comment reminds to write a more accurate guide for the database migration, i.e using `PORT=8002 docker-compose exec web <command>` once the process is validated & tested

.github/workflows/main.yml
jeanpasquier75: I stil need to check the pytest-alembic or equivalent, anyway we will need to create a specific github action "on push branch master" (and not on pull request): we want to run the migration only once the PR is merged right ?

Dockerfile
frgfm: I get your point. However, since Docker is not the only way to run the API, either we add instructions in the doc, or we try to fix this import :thinking:

.github/workflows/main.yml
frgfm: Correct, so I will add a CI job in another PR I guess

fe51 (dismissed):

src/alembic/README.md
frgfm: Good idea, let's add the proper instruction in this PR then

jeanpasquier75: What do you think ? I added a link from CONTRIBUTING.md to the alembic/README.md

frgfm (approved):
Thanks for the edits! I'm still having doubts for the Dockerfile modifications but it's nitpicking

src/alembic/env.py
```diff
@@ -0,0 +1,82 @@
+from logging.config import fileConfig
```
frgfm, line 1: You need to add the copyright & license notice over here :sweat_smile:

frgfm (commented):
Just the copyright & license notice missing :ok_hand:

fe51 (approved):

frgfm (approved):
All good now!

.github/workflows/main.yml
frgfm: I opened #140 about this

frgfm:
@jeanpasquier75 I think we are ready to merge. Since we don't have a unittest or any CI integration yet, can you confirm that you tried to modify one of the table and used your alembic instructions to perform the migration, which was successful ? :)

fe51 (approved):

jeanpasquier75:
> @jeanpasquier75 I think we are ready to merge. Since we don't have a unittest or any CI integration yet, can you confirm that you tried to modify one of the table and used your alembic instructions to perform the migration, which was successful ? :)

Yes sir I confirm that it works fine in my postgres image 😉

PR #113, opened 2021-02-22
Align devices update 

### LINKED ISSUE #103: Some not specified fields should not be updated to null
Currently when a PUT request is done and one optional field is not specified it is updated to null. 

A solution should be to add an option boolean to crud.put or crud.update_entry, if set to True it will update only the not-None fields

jeanpasquier75:
Relates to discussion in #45 (devices route) and closes #103 for the route `PUT /devices/{device_id}/`

* Add admin scope (let's say that only admin can change the `owner_id`)
* Use the option `only_provided` defined in `crud.update_entry` that updates only the fields provided in the payload (eg. the admin only updates the `owner_id`, the device still gets its position, etc.)
* Change low level `payload` argument of crud.base.put to dictionary instead of `pydantic.BaseModel` & migrate routes that used the function
* Add the only_provided option that remove None values from payload, retrieve from table record afterwards to get all fields ? Waiting for feedbacks on this one

This is first iteration, more concret than issue comments haha

### DIFF
--- a/src/app/api/crud/accesses.py
+++ b/src/app/api/crud/accesses.py
@@ -72,7 +72,7 @@ async def create_accessed_entry(
     return entry
 
 
-async def update_accessed_entry(table: Table, accesses: Table, entry_id: int, payload: Any):
+async def update_accessed_entry(table: Table, accesses: Table, entry_id: int, payload: Any, only_provided=False):
     """Update an entry with a special treatment regarding login: if login is set -> update corresponding access."""
     # Ensure database consistency between tables with a transaction (login must remain the same in table & accesses)
     async with base.database.transaction():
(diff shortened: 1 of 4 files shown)


frgfm:
Thanks for all this work! 
Two things apart from the review:
- the owner of a device should be able to transfer ownership to someone else. Do you see a way for us to integrate this in the PR?
- when I first read "provided_only" I admit I had no clue what you were about to say. I would suggest replacing with "only_specified"

frgfm (commented):
Apart from my previous comments, the rest of the PR looks good to me!

src/app/api/crud/accesses.py
```diff
@@ -72,7 +72,7 @@ async def create_accessed_entry(
     return entry
 
 
-async def update_accessed_entry(table: Table, accesses: Table, entry_id: int, payload: Any):
+async def update_accessed_entry(table: Table, accesses: Table, entry_id: int, payload: Any, only_provided=False):
```
florianriche, line 75: Maybe add a UTest?
IMHO the default boolean should be set to True :)

florianriche (approved):

src/app/api/crud/accesses.py
jeanpasquier75: I changed the argument to `only_specified` and set its default value to True

frgfm (approved):
Thanks a lot!

PR #116, opened 2021-03-14
feat: Added route to fetch past fires

### LINKED ISSUE #109: Fetching past fires via the API
Hello everyone, 

Here is a non-urgent issue related to [this one](https://github.com/pyronear/pyro-platform/issues/35) in the Pyro-Platform repository!

In order to display past fires on the map, we are currently reading a local file stored [within the repository](https://github.com/pyronear/pyro-platform/blob/master/app/data/historic_fires.csv). The goal would be to remove it and instead fetch this information via the Pyro-API. If I understood well, for now, there is no table dedicated to past fires (which could be useful not only for the platform but also for the data science team typically) and this issue is designed to discuss its addition to the database. 

To be more precise about the `historic_fires.csv` file, it was built out of a raw dataset of satellite fire detections which we preprocessed, notably in [this Colab](https://colab.research.google.com/drive/18wDMzsGUwFWhjt4frgxEVngKvhETfME5?usp=sharing). We added several fields like the department and the "commune" where the fire occurred (as well as the population, area and density of the locality). 

We wanted to use the latter to compute the density of population in the area, so as to approximately filter urban fires out of the dataset. However, we have not yet implemented this filtering on the platform and for now, `department` is the only added field which is really required. I mention this in case we are limited for the number of fields.

Last remark: at some point, I guess we might want to add fires detected by our own devices and acknowledged by firemen to the past fires table in the database. 

All that being said, here are my question(s):
- Should we add the `historic_fires.csv` file as is to the database? 
- Or do you prefer that we script the preprocessing steps and use the raw dataset? 

Thanks a lot for your help, I hope that everything is clear!

frgfm:
- added exclusion criteria option in fetch_all
- added route to fetch past events
- added corresponding unittest
- added API client integration
- added client unittest

Successfully merging this PR will close #109

Any feedback is welcome

### DIFF
--- a/src/app/api/crud/base.py
+++ b/src/app/api/crud/base.py
@@ -35,11 +35,19 @@ async def get(entry_id: int, table: Table) -> Mapping[str, Any]:
     return await database.fetch_one(query=query)
 
 
-async def fetch_all(table: Table, query_filters: Optional[Dict[str, Any]] = None) -> List[Mapping[str, Any]]:
+async def fetch_all(
+    table: Table,
+    query_filters: Optional[Dict[str, Any]] = None,
+    exclusions: Optional[Dict[str, Any]] = None,
+) -> List[Mapping[str, Any]]:
     query = table.select()
     if isinstance(query_filters, dict):
-        for query_filter_key, query_filter_value in query_filters.items():
-            query = query.where(getattr(table.c, query_filter_key) == query_filter_value)
+        for key, value in query_filters.items():
+            query = query.where(getattr(table.c, key) == value)
+
+    if isinstance(exclusions, dict):
+        for key, value in exclusions.items():
+            query = query.where(getattr(table.c, key) != value)
     return await database.fetch_all(query=query)
 
 
--- a/src/app/api/routes/events.py
+++ b/src/app/api/routes/events.py
@@ -40,6 +40,14 @@ async def fetch_events():
     return await crud.fetch_all(events)
 
 
+@router.get("/past", response_model=List[EventOut], summary="Get the list of all past events")
+async def fetch_past_events():
+    """
+    Retrieves the list of all events and their information
+    """
+    return await crud.fetch_all(events, exclusions={"end_ts": None})
+
+
 @router.put("/{event_id}/", response_model=EventOut, summary="Update information about a specific event")
 async def update_event(
     payload: EventIn,
--- a/client/pyroclient/client.py
+++ b/client/pyroclient/client.py
@@ -35,6 +35,7 @@ ROUTES: Dict[str, str] = {
     "acknowledge-alert": "/alerts/{alert_id}/acknowledge",
     "get-site-devices": "/installations/site-devices/{site_id}",
     "get-media-url": "/media/{media_id}/url",
+    "get-past-events": "/events/past",
 }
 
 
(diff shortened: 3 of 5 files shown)


src/app/api/crud/base.py
```diff
@@ -35,11 +35,19 @@ async def get(entry_id: int, table: Table) -> Mapping[str, Any]:
...
+async def fetch_all(
+    table: Table,
+    query_filters: Optional[Dict[str, Any]] = None,
+    exclusions: Optional[Dict[str, Any]] = None,
+) -> List[Mapping[str, Any]]:
     query = table.select()
     if isinstance(query_filters, dict):
-        for query_filter_key, query_filter_value in query_filters.items():
-            query = query.where(getattr(table.c, query_filter_key) == query_filter_value)
+        for key, value in query_filters.items():
```
florianriche, line 45: Can you add a UTest for this behavior?

src/app/api/routes/events.py
```diff
@@ -40,6 +40,14 @@ async def fetch_events():
     return await crud.fetch_all(events)
 
 
+@router.get("/past", response_model=List[EventOut], summary="Get the list of all past events")
+async def fetch_past_events():
+    """
+    Retrieves the list of all events and their information
+    """
+    return await crud.fetch_all(events, exclusions={"end_ts": None})
```
florianriche, line 48: Isn't it going to be a lot to return? 

At some point we will return hundreds of fire (and other events in the future), donno if the JSon file may take it. 
Shouldn't we add a clause to restrict the scope?

florianriche (commented):
LGTM besides my two small comments

src/app/api/routes/events.py
frgfm: Do you mean a maximum number of returned entries?

src/app/api/crud/base.py
frgfm: For which lines do you mean exactly?
I only renamed variables in lines 45-46

The real addition is on line 48-50, and it is covered by the unittest I added?

src/app/api/routes/events.py
florianriche: Yes for example, kind of like a pagination.
Or limit the returned results to a department or something. Maybe I am overthinking this.

src/app/api/crud/base.py
florianriche: Yes I meant the lines 48-50.
~~The UTest you added check the route behavior, not the CRUD basic function.~~

But I just realized we haven't such dedicated tests so ¯\_(ツ)_/¯ let's go :)

src/app/api/routes/events.py
frgfm: I agree on limiting this to a region, but we don't have the means to do this currently. Or do we? (I might be missing something)

Perhaps limiting the number of entries would be a good idea. I just checked [here](https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.query.Query.limit) it's possible, but if we limit this, we would need to order the results first (to return the most recent). I think it might be better to keep this for another PR

florianriche: ok :+1:

florianriche (approved):

PR #130, opened 2021-03-18
feat: Added no_alert type to Sites

### LINKED ISSUE #124: [sites] No alert sites
## Feature

- [ ] Add `no_alert` as SiteType in Site table
- [ ] Create new route to allow no admin user  to create a no_alert Site

## Motivation

Some sites can be a source of smoke (like a quarry or a plant) and it is needed to record those specific location in order to not raise useless alerts. That could be done by adding a new site type "no_alert".

Moreover, users without admin scope should be able to create this new type of site. Hence a new route is needed !

Any feedback is welcome !

frgfm:
Following up on #124, this PR introduces the following modifications:
- added a new site type (still default to `tower`)
- added a new route to create sites that doesn't require admin rights but enforces `type='no_alert'` so that users can create them
- added unittest cases to ensure default value, check payload & response

Resolves #124

Any feedback is welcome!

### DIFF
--- a/src/app/api/routes/sites.py
+++ b/src/app/api/routes/sites.py
@@ -6,8 +6,8 @@
 from typing import List
 from fastapi import APIRouter, Path, Security
 from app.api import crud
-from app.db import sites
-from app.api.schemas import SiteOut, SiteIn
+from app.db import sites, SiteType
+from app.api.schemas import SiteOut, SiteIn, SiteBase
 from app.api.deps import get_current_access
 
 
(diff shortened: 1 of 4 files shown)


src/app/api/routes/sites.py
```diff
@@ -24,6 +24,17 @@ async def create_site(payload: SiteIn, _=Security(get_current_access, scopes=["a
     return await crud.create_entry(sites, payload)
 
 
+@router.post("/no-alert/", response_model=SiteOut, status_code=201, summary="Create a new no-alert site")
```
florianriche, line 27: /no-alert/ is confusing I think.

create-no-alert-site or something like that would be better IMHO

florianriche (approved):
Looks good to me as I have done almost the Same PR ahah 

Shall we add a client route though?

src/app/api/routes/sites.py
frgfm: I disagree, having route names that are too long isn't a good idea.
Regarding the naming, I thought that somehow that do a POST method on sites/no-alert/ would easily understand?
(the naming on the route should be independent on the method type, so adding "create" in the name sounds like a bad idea to me :man_shrugging: )

frgfm:
> Looks good to me as I have done almost the Same PR ahah
> 
> Shall we add a client route though?

Oh damn, perhaps we could turn your PR into the API client addition?

src/app/api/routes/sites.py
florianriche: Fair enough.

florianriche:
> > Looks good to me as I have done almost the Same PR ahah
> > Shall we add a client route though?
> 
> Oh damn, perhaps we could turn your PR into the API client addition? Or do the opposite, I don't mind

Let's merge this one, I prefer your Utest

florianriche (approved):

PR #132, opened 2021-03-18
[Site] Create no alert sitesType and add matching route

### LINKED ISSUE #124: [sites] No alert sites

florianriche:
Closes #124

### DIFF
--- a/client/pyroclient/client.py
+++ b/client/pyroclient/client.py
@@ -22,6 +22,7 @@ ROUTES: Dict[str, str] = {
     "heartbeat": "/devices/heartbeat",
     "update-my-location": "/devices/update-my-location",
     "create-event": "/events",
+    "create-no-alert-site": "/sites/create-no-alert-site",
     "send-alert": "/alerts",
     "send-alert-from-device": "/alerts/from-device",
     "create-media": "/media",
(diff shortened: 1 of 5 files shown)


src/app/api/routes/sites.py
```diff
@@ -24,6 +24,17 @@ async def create_site(payload: SiteIn, _=Security(get_current_access, scopes=["a
     return await crud.create_entry(sites, payload)
 
 
+@router.post("/create-no-alert-site", response_model=SiteOut, status_code=201, summary="Create a new site")
```
frgfm, line 27: Would you mind renaming it to something shorter please?

client/pyroclient/client.py
```diff
@@ -22,6 +22,7 @@
     "heartbeat": "/devices/heartbeat",
     "update-my-location": "/devices/update-my-location",
     "create-event": "/events",
+    "create-no-alert-site": "/sites/create-no-alert-site",
```
frgfm, line 25: will have to be updated if we move forward with my other comment :sweat_smile:

client/pyroclient/client.py
```diff
@@ -131,6 +132,31 @@ def create_event(self, lat: float, lon: float) -> Response:
...
+
+        Returns:
+            HTTP response containing the created site
+        """
+        payload = {"lat": lat,
+                   "lon": lon,
+                   "name": name,
+                   "country": country,
+                   "geocode": geocode}
+        return requests.post(self.routes["create-no-alert-site"], headers=self.headers, json=payload)
```
frgfm, line 158: same here

frgfm (changes requested):
Apart from the subroute naming, we should move forward with your PR rather than mine since you added the client side!

PR #134, opened 2021-03-18
feat: Added angle_of_view to the device table

### LINKED ISSUE #131: [devices] Add angle_of_view field to device table
This is important information to have a comprehensive understanding of the visual situation of a device raising an alert.

Using this information + visual data + device orientation (yaw), the device can compute the azimuth of the alert/wildfire.

frgfm:
Following up on #131, this PR introduces the following modifications:
- added a float field `angle_of_view` to the device table (mandatory in payload)
- added unittest cases to ensure  value constraints (between 0 and 360), check payload & response
- updated other unittests to make it compatible

Resolves #131

Any feedback is welcome!

### DIFF
--- a/src/app/api/schemas.py
+++ b/src/app/api/schemas.py
@@ -142,6 +142,7 @@ class EventOut(EventIn, _CreatedAt, _Id):
 class MyDeviceIn(Login, DefaultPosition):
     specs: str = Field(..., min_length=3, max_length=100, example="systemV0.1")
     last_ping: datetime = Field(default=None, example=datetime.utcnow())
+    angle_of_view: float = Field(..., gt=0, lt=360)
 
 
 class DeviceIn(MyDeviceIn):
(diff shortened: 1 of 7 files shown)


florianriche:
Can you update the client accordingly?
Is it linked to an event afterward? You may do all in one commit don't you think?

frgfm:
@florianriche I'm not following: this is the angle of view of the camera (doesn't change for a device apart from physical modifications), there is nothing to add to the client :)

florianriche:
Ineed, spoke without thinking,
But I have a hard time seeing how the different PR are supposed to be articulated.

I think I just understood that angle_of_view is the camera angle ?

frgfm:
@florianriche well as per #131, this is the angular field of view of the camera.
Let's say from the picture of the detection, we are able to tell the wildfire is on the right corner. To be able to point towards that wildfire, we need the position on the picture, the orientation of the camera and the angle of view. This PR is implementing exactly that

florianriche (approved):

PR #137, opened 2021-03-20
feat: add group object 

### LINKED ISSUE #125: [users] Implementing user_group
Following up on #45, I think it is sound to have a discussion about the multiple access levels in the API. With the current design of scopes, we cannot efficiently separate accesses of large user groups.

## Requirements
- We need to restrict "superadmin" from common users
- Users need to be segmented into groups, where each group only has access to a restricted scope (i.e. a local firefighter, while having a "user" scope, should not have access to the same data as someone from another region, while that person is also a "user")
- Groups need to be easily editable
- security on routes will have to combine those to provide proper and secure access to each route

## Design suggestion
Simply put, I believe two fields are required:
- access_type: admin, user, device, me
- access_group: ID that would reference a group (we could add a `groups` table with `group_id`, and `group_name`)

florianriche:
Create a new object called "group".
That object is linked to a user entity.
Right now, only the Unit tests and the code related to the group management has been added.

Before implementing the code linked to the authorization management, I would like opinions on this schema.

Closes #125

### DIFF
--- a/src/app/api/schemas.py
+++ b/src/app/api/schemas.py
@@ -7,7 +7,7 @@ from typing import List, Optional
 from datetime import datetime
 from pydantic import BaseModel, Field, validator
 
-from app.db.tables import SiteType, EventType, MediaType, AlertType, AccessType
+from app.db.tables import SiteType, EventType, MediaType, AccessType
 
 
 # Template classes
@@ -218,7 +218,6 @@ class AlertMediaId(BaseModel):
 
 class AlertBase(_FlatLocation, AlertMediaId):
     event_id: int = Field(..., gt=0)
-    type: AlertType = AlertType.start
     is_acknowledged: bool = Field(False)
     azimuth: float = Field(default=None, gt=0, lt=360)
 
--- a/src/app/api/schemas.py
+++ b/src/app/api/schemas.py
@@ -56,8 +56,7 @@ class AccessRead(AccessBase, _Id):
 # Users
 class UserInfo(Login):
     # Abstract information about a user
-    group_id: int = Field(..., gt=0)
-
+    pass
 
 class UserRead(UserInfo, _CreatedAt, _Id):
     # Visible info
--- a/src/app/api/schemas.py
+++ b/src/app/api/schemas.py
@@ -56,7 +56,7 @@ class AccessRead(AccessBase, _Id):
 # Users
 class UserInfo(Login):
     # Abstract information about a user
-    pass
+    group_id: int = Field(..., gt=0)
 
 
 class UserRead(UserInfo, _CreatedAt, _Id):
@@ -72,6 +72,19 @@ class UserAuth(UserInfo, Cred):
 class UserCreation(UserInfo):
     # Creation payload
     access_id: int = Field(..., gt=0)
+
+
+# Groups
+class GroupBase(BaseModel):
+    name: str = Field(..., min_length=3, max_length=50, example="Fireman85")
+
+
+class GroupIn(GroupBase):
+    pass
+
+
+class GroupOut(GroupBase, _Id):
+    pass
 
 
 # Token
--- a/src/app/api/schemas.py
+++ b/src/app/api/schemas.py
@@ -7,7 +7,7 @@ from typing import List, Optional
 from datetime import datetime
 from pydantic import BaseModel, Field, validator
 
-from app.db import SiteType, EventType, MediaType, AlertType
+from app.db.tables import SiteType, EventType, MediaType, AlertType, AccessType
 
 
 # Template classes
@@ -38,7 +38,7 @@ class CredHash(BaseModel):
 
 
 class AccessBase(Login):
-    scopes: str = Field(..., example="me")
+    scopes: AccessType = AccessType.user
 
 
 class AccessAuth(AccessBase, Cred):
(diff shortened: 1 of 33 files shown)


frgfm:
Thanks for the PR @florianriche !
Since this is a rather big deal as a feature, I think we will most likely have to split this into several PRs depending on the design :sweat_smile:
Let's discuss this in the issue #125 quickly :ok_hand:

florianriche:
> Thanks for the PR @florianriche !
> Since this is a rather big deal as a feature, I think we will most likely have to split this into several PRs depending on the design 
> Let's discuss this in the issue #125 quickly

I have updated the PR to remove the reference to group_id in user table

src/tests/routes/test_alerts.py
```diff
@@ -31,6 +31,10 @@
     {"id": 2, "login": "connected_user", "access_id": 2, "created_at": "2020-11-13T08:18:45.447773"},
 ]
 
+GROUP_TABLE = [
+    {"id": 1, "name": "test_group"},
+]
+
```
frgfm, line 37: Let's remove everything in the unittests that doesn't involve groups yet (so everything apart from test_groups.py)

src/tests/routes/test_devices.py
```diff
@@ -17,6 +17,10 @@
     {"id": 2, "login": "connected_user", "access_id": 2, "created_at": "2020-11-13T08:18:45.447773"},
 ]
 
+GROUP_TABLE = [
+    {"id": 1, "name": "test_group"},
+]
+
```
frgfm, line 23: same

src/app/api/schemas.py
```diff
@@ -74,6 +73,19 @@ class UserCreation(UserInfo):
     access_id: int = Field(..., gt=0)
 
 
+# Groups
+class GroupBase(BaseModel):
```
frgfm, line 77: You can rename this into GroupIn, no need for GroupBase

src/app/api/schemas.py
```diff
@@ -74,6 +73,19 @@ class UserCreation(UserInfo):
...
+    name: str = Field(..., min_length=3, max_length=50, example="Fireman85")
+
+
+class GroupIn(GroupBase):
+    pass
```
frgfm, line 82: we can remove this

src/app/api/schemas.py
```diff
@@ -74,6 +73,19 @@ class UserCreation(UserInfo):
...
+
+
+class GroupOut(GroupBase, _Id):
```
frgfm, line 85:
```
class GroupOut(GroupIn, _Id):
```

frgfm (commented):
Hey there!
Thanks for the PR, I added a few comments. I suggest waiting for #138 to be merged before writing the actual group unittests

frgfm (changes requested):
Just need to fix flake8 checks on the new unittests ;)

frgfm:
Shoot, this also needs some merging @florianriche :sweat_smile:

src/tests/routes/test_groups.py
```diff
@@ -0,0 +1,169 @@
...
+]
+
+
+def compare_entries(ref, test):
+    for k, v in ref.items():
+        if isinstance(v, float):
+            # For float issues
+            assert abs(v - test[k]) < 1E-5
+        else:
+            assert v == test[k]
```
frgfm, line 34: Considering that there is no float field in the group table, perhaps we don't need this?

src/tests/routes/test_groups.py
```diff
@@ -0,0 +1,169 @@
...
+async def test_get_group(test_app_asyncio, init_test_db, group_id, status_code, status_details):
+
+    response = await test_app_asyncio.get(f"/groups/{group_id}")
+    response_json = response.json()
+    assert response.status_code == status_code
+
+    if isinstance(status_details, str):
+        assert response.json()['detail'] == status_details
+    if response.status_code == 200:
+        compare_entries(response_json, GROUP_TABLE[group_id - 1])
```
frgfm, line 65: I think we can check for perfect match (no float field here)

src/tests/routes/test_groups.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+@pytest.mark.asyncio
+async def test_fetch_groups(test_app_asyncio, init_test_db):
+
+    response = await test_app_asyncio.get("/groups/")
+    assert response.status_code == 200
+    response_json = response.json()
+    for (i, entry) in enumerate(response_json):
+        compare_entries(entry, GROUP_TABLE[i])
```
frgfm, line 75: same here

src/tests/routes/test_groups.py
```diff
@@ -0,0 +1,169 @@
...
+    auth = await pytest.get_token(ACCESS_TABLE[access_idx]['id'], ACCESS_TABLE[access_idx]['scopes'].split())
+
+    response = await test_app_asyncio.delete(f"/groups/{group_id}/", headers=auth)
+    assert response.status_code == status_code
+
+    if isinstance(status_details, str):
+        assert response.json()['detail'] == status_details
+
+    if response.status_code // 100 == 2:
+        compare_entries(response.json(), GROUP_TABLE[group_id - 1])
```
frgfm, line 167: same here

frgfm (approved):
Thanks for the PR! I only added a last comment about `compare_entries`: since there is no float field in the table, we can check for exact match without it. But there the rest looks good to me!

PR #138, opened 2021-03-23
test: Fixed unittests scoping issues

### LINKED ISSUE #119: [test] Access check does not work correctly in unittests
While I was changing scopes on some routes, I noticed something:
- changes on the scope requirements do impact the API behaviour expectedly
- however in the unittest, even if the `get_current_access` yields an access with insufficient scope, the request is still being processed

We would need to investigate and fix this to be able to check scopes in the unittests.

frgfm:
Following up on #119, I conducted numerous tests to understand why our scope constraint wasn't taken into account by our unittests. The reason behind that is that we overrode the checking function improperly. This PR introduces the following modifications:
- renamed `reusable_oauth2` to `oauth2_scheme` in `deps.py`
- added extra credential check in `get_current_access`
- renamed `create_device` to `register_device`
- fixed status code in `alerts.link_media` and `device.register_device`
- fixed scope constraint on `create_noalert_site`
- renamed `conf_test_db` to `db_utils` and refactored it
- refactored `conftest` by removing unnecessary fixtures and adding appropriate ones
- added unittest for `get_current_access` and `create_access_token`
- refactored every single unittest file as follows:
1. turned `init_test_db` into a fixture
2. whenever a scope is required, I created a token just like in production to avoid mocking access-related functions (this allows switching between credentials)
3. added numerous test cases to ensure that the route matches its scope constraints

The PR is unfortunately quite heavy since it's cumbersome to do this in separate PRs without triggering hundreds of errors. Sorry about that! The core logic of the unittest is unchanged, but they are now covering much more than before
Resolves #119

Any feedback is welcome!


src/tests/routes/test_accesses.py
```diff
@@ -7,49 +7,72 @@
...
     await populate_db(test_db, db.accesses, ACCESS_TABLE)
 
-    response = await test_app_asyncio.get("/accesses/1")
-    assert response.status_code == 200
-    assert response.json() == {k: v for k, v in ACCESS_TABLE[0].items() if k != "hashed_password"}
-
 
 @pytest.mark.parametrize(
-    "access_id, status_code, status_details",
+    "access_idx, access_id, status_code, status_details",
```
florianriche, line 27: why idx and not ids?

florianriche (approved):
I haven't do a thorough review to be honest but nothing strucked me.

src/tests/routes/test_accesses.py
frgfm: simply because it was more cumbersome to retrieve the correct ID from ACCESS_TABLE :sweat_smile:

PR #141, opened 2021-03-26
refactor: Removed the "me" scope and replace it with "user"

### LINKED ISSUE #139: [scope] The "me" scope is supposedly useless
Having refactored most of the unittests, I came to the conclusion that the "me" scope is useless. 

Simply put, here is a route using it now:
```
@router.get("/me", response_model=UserRead, summary="Get information about the current user")
async def get_my_user(me: UserRead = Security(get_current_user, scopes=["admin", "me"])):
    """
    Retrieves information about the current user
    """
    return me
```

Now if we use the `user` scope instead:
```
@router.get("/me", response_model=UserRead, summary="Get information about the current user")
async def get_my_user(me: UserRead = Security(get_current_user, scopes=["admin", "user"])):
    """
    Retrieves information about the current user
    """
    return me
```
assuming all entries in the user table with "me" will be replaced by "user", we get the exact same behaviour. 

There is no difference in "access level" between a "user" and "me" (or "device" and "me" depending on the situation). So I suggest that we remove it to avoid unnecessary complication. We will have to discuss access levels later on to distinguish a bare user from a "supervisor" or a "manager".

I suggest waiting for #138 approval before starting a PR to avoid conflicts.

frgfm:
- fixed oauth2 scheme (was missing "user")
- replaced all security scope constraints with "me" by "user"
- switched access scope to Enum type in DB
- updated unittests accordindly

Resolves #139

Any feedback is welcome!


fe51:
Okay, now with the PR and remembering that sometimes scopes were referring to "user "which was not existing in oauth2 scheme, I understand the aim of the issue which even more like a bug correction than an enhancement ! :)

Thanks for fixing it !

It looks okay to me !

fe51 (approved):

frgfm:
> Okay, now with the PR and remembering that sometimes scopes were referring to "user "which was not existing in oauth2 scheme, I understand the aim of the issue which even more like a bug correction than an enhancement ! :)
> 
> Thanks for fixing it !
> 
> It looks okay to me !

Not exactly: not being available in the oauth2_scheme is different from being non-existent in the database. That being said, so far since we haven't used the "me" or the "user" scope on Heroku, we would not have noticed.

The point of the PR is:
- indeed fix the oauth2_scheme
- since the "me" doesn't seem to have any useful purpose compared to "user", removing the "me" (which we were sometimes using for both device or user)

PR #142, opened 2021-03-26
refactored: Removed alert type field

frgfm:
- removed alert type since only "start" types are created
- updated ongoing alert fetching
- updated unittests accordingly

Any feedback is welcome!


fe51:
Thanks for the PR. 

Looks ok to me !

however, I realise, we are not able to get the time to acknowledge an alert (was the case before this PR).
It could be interesting to have this information, to see the reactivity of our users ? What do you think ? I could open the related issue if need (could be solved by adding and "ended_at" field filled when the alert is acknowledged.

fe51 (dismissed):

frgfm:
@fe51 regarding the timestamp an alert gets acknowledged, I'm not sure we need it. On the long-term it might be interesting to check the duration between detection and the moment firefighters dispatch teams, but I'm pretty sure we would have forgotten a few other things to track for this.

fe51:
> @fe51 regarding the timestamp an alert gets acknowledged, I'm not sure we need it. On the long-term it might be interesting to check the duration between detection and the moment firefighters dispatch teams, but I'm pretty sure we would have forgotten a few other things to track for this.

Okay then ! Not a priority anyway !

fe51 (approved):

Issue #128, opened 2021-03-18
[alerts] Reference to installation_id rather than device_id


frgfm:
When we first designed all the table fields, we added the `device_id` as the emitter of the alert. I'm starting to think it would make way more sense to replace this by `installation_id`. There are pros and cons

## Pitch
Here is how the change would play out:

### Triggering the alert
This would mean that a device per se cannot send an alert without being an installation itself i.e. linked to a site (the biggest con). The API client will handle the alert sending by resolving the installation from the device. Additionally:
- We could either revise our notion of installation: imagine an individual positioning the device in her/his garden, there is no site to be linked to necessarily. But it is an "installation" since the device will stay positioned over there.
- or accept the inconvenience

### People receiving the alert
Before you were receiving the device_id:
- GET method to reading of device position & specs
- more thorough resolution of installation if you want to get the site it was sent from

Now you would receive the installation_id:
- direct GET on site information
- using site information, you get the same device information

One last option would be to have both but I feel like this would be a dangerous path later on. What do you think @pyronear/back-end ?

fe51:
Thank you for opening the discussion ! 

What do you mean by _The API client will handle the alert sending by resolving the installation from the device_ ? 
(sending an alert would consist in first getting the installation id currently related to the device and then send the alert ?)

In the case of an individual and its nice garden, both options seems okay to me.

By the way, elevation, lat, lon, yaw pitch are a bit redondant between the two table (device and installation) right ?

frgfm:
On the redundancy part, I fully agree, I suggest we remove it from the installation table (which should only be a pairing of a device with a site for a given time interval)

frgfm:
For the reason above (that it would prevent a device from sending an alert if it's not on a site), I'm closing this issue

PR #143, opened 2021-03-29
refactor: Removed location fields from the installation table

frgfm:
Following up on https://github.com/pyronear/pyro-api/issues/128#issuecomment-808556151, this PR introduces the following modifications:
- removed location-related fields from the installation table (since they are present in the device table, which gets updated)
- updated unittests & e2e script accordingly

Any feedback is welcome!


fe51:
What a responsiveness :)

Following the comment, I fully agree ! 

The PR is okay to me !
Thanks FG !

fe51 (approved):

PR #148, opened 2021-03-31
Add function to check for group

florianriche:
**Current content of the review:**

* Modified all the scopes string into AccesstType Enum values
* Added an `authorizations.py` files in `/crud` folder to test groups match and that an access is an admin or not.
  * My current vision would be to simply use the functions defined in this file everywhere it is needed (in routes mostly). I think it would handle most of our use cases we have defined in #137  (except the fetch use case)
  * Added a  ̀groups.py` file in  `/crud` folder to find the group of any entity.
  * ADded corresponding uTest

Admittedly this design is not very clean but to be honest I don't know how else to do it.
(EDITED)


src/app/api/crud/authorizations.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+async def are_same_group(group_id_a: int, group_id_b: int) -> bool:
+    if group_id_a != group_id_b:
+        raise HTTPException(status_code=400, detail="You don't belong to the same group")
+    return True
+
+
+async def is_admin_access(access_id: int) -> bool:
+    access = await crud.base.get_entry(accesses, access_id)
```
frgfm, line 32: Note for later: we only need 1 field, perhaps we should either add the flexibility to specify the fields or write a custom query.

src/app/api/crud/authorizations.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+async def is_access_in_same_group(access_id: int, group_id: int) -> bool:
+    access = await crud.base.get_entry(accesses, access_id)
+    return await are_same_group(access.group_id, group_id)
+
+
+async def are_same_group(group_id_a: int, group_id_b: int) -> bool:
+    if group_id_a != group_id_b:
+        raise HTTPException(status_code=400, detail="You don't belong to the same group")
```
frgfm, line 27: 401 I believe ;)
```
from fastapi import status
raise HTTPException(status_code=status.HTTP_401_UNAUTHORIZED, detail="Permission denied")
```

src/app/api/crud/authorizations.py
```diff
@@ -0,0 +1,33 @@
...
+    user_or_device = await crud.base.get_entry(table, entry_id)
+    return await is_access_in_same_group(user_or_device.access_id, group_id)
+
+
+async def is_access_in_same_group(access_id: int, group_id: int) -> bool:
+    access = await crud.base.get_entry(accesses, access_id)
+    return await are_same_group(access.group_id, group_id)
+
+
+async def are_same_group(group_id_a: int, group_id_b: int) -> bool:
```
frgfm, line 25: this doesn't need to be async

src/app/api/crud/authorizations.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+async def is_in_same_group(table: Table, entry_id: int, group_id: int) -> bool:
+    user_or_device = await crud.base.get_entry(table, entry_id)
+    return await is_access_in_same_group(user_or_device.access_id, group_id)
+
+
+async def is_access_in_same_group(access_id: int, group_id: int) -> bool:
+    access = await crud.base.get_entry(accesses, access_id)
+    return await are_same_group(access.group_id, group_id)
```
frgfm, line 22: if this is only a value comparison, perhaps the extra function isn't needed?

src/app/api/crud/authorizations.py
```diff
@@ -0,0 +1,33 @@
...
+from app.api.schemas import AccessType
+from app.db import accesses
+
+
+async def is_in_same_group(table: Table, entry_id: int, group_id: int) -> bool:
+    user_or_device = await crud.base.get_entry(table, entry_id)
+    return await is_access_in_same_group(user_or_device.access_id, group_id)
+
+
+async def is_access_in_same_group(access_id: int, group_id: int) -> bool:
```
frgfm, line 20: I suggest `is_access_in_group` considering the signature

frgfm (commented):

src/app/api/crud/authorizations.py
florianriche: Its purpose is mainly to raise the HTTPException and to be called directly by objects that contain a group_id directly (like `sites`)
We can do without but I would personally keep it

florianriche:
Thanks for the feedbacks, do you approve this design?

If I end up with additional obstacles I will mention them here.

frgfm:
@florianriche yes thanks for taking care of this!

src/app/api/crud/authorizations.py
florianriche: Ok, I copy-pasted from antoher part, there are still 400 errors returned with that message :)

florianriche:
Shall we add many scopes in this pull request?

Personnal check-list that I will update.
- [x]  User:  routes all have Admin or can't alter group so no check needed
- [x] Device routes: routes all have Admin or can't alter group so no check needed
- [X] Alerts: routes all have aDmin or can't alter group so no check needed
- [x] Installations: Only Admin scopes
- [x] Sites: Added when requested (and added scopes for GET ressources)
- [x] media: updated  GET resources, Fetch sce
- [X] events: route access
- [x] accesses: only Admin

frgfm:
I think we should keep this PR only about being able to trace the group for each entity. And in another PR implement the next group constraints :ok_hand: (those are hard to review carefully)

frgfm:
For reference, this PR is linked to #125

frgfm:
You got some import conflicts during test collection :sweat_smile:

florianriche:
> You got some import conflicts during test collection

Yes fixed it :)

src/app/api/crud/groups.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+async def _get_media_group_id(entry_id: int) -> int:
+    entry = await crud.base.get_entry(media, entry_id)
+    return await _get_device_group_id(entry["device_id"])
+
+
+async def _get_installation_group_id(entry_id: int) -> int:
+    entry = await crud.base.get_entry(installations, entry_id)
+    return await _get_device_group_id(entry["device_id"])
```
frgfm, line 73: Slight improvement suggestion: it's cheaper using _get_site_group_id in terms of DB operations ;)

src/app/api/crud/groups.py
```diff
@@ -0,0 +1,78 @@
...
+
+from app.api import crud
+from app.api.schemas import AccessType
+from app.db import (
+    accesses, users, groups,
+    sites, events, devices, media,
+    installations, alerts)
+
+
+async def get_entity_group_id(table: Table, entry_id: int) -> int:
```
frgfm, line 19: Improvement suggestion: for cleaner signature, I suggest passing directly the entry. Then the ifs can be checking whether it is `AccessRead` or other reading schemas

src/tests/crud/test_authorizations.py
```diff
@@ -0,0 +1,183 @@
...
+        [2, 1, False],  # Because Admin
+        [2, 2, False],
+        [2, None, False]
+
+    ],
+)
+@pytest.mark.asyncio
+async def test_check_group_update(test_app_asyncio, init_test_db, access_id, group_id, should_raise):
+    if should_raise:
+        with pytest.raises(Exception):
```
frgfm, line 180: HTTPException?

src/tests/crud/test_authorizations.py
```diff
@@ -0,0 +1,183 @@
...
+        [1, 1, False],
+        [1, 2, True],
+        [2, 1, False],  # Because Admin
+        [2, 2, False]
+    ],
+)
+@pytest.mark.asyncio
+async def test_check_group_read(test_app_asyncio, init_test_db, access_id, group_id, should_raise):
+    if should_raise:
+        with pytest.raises(Exception):
```
frgfm, line 159: HTTPException?

frgfm (commented):
Thank you very much! Everything looks good, I only added a few adjustment suggestions :ok_hand:

src/app/api/crud/groups.py
florianriche: Not sure it will be more efficient. One will need to retrieve the entity before doing the group_id check

florianriche: But it's true that it is already what we do.
Ok I'll try. But I fear that the Schema checks might be a more brittle design.

frgfm (approved):
Thanks a lot!

Issue #146, opened 2021-03-29
[alerts][events] How to define the event life cycle with alerts


frgfm:
Let's discuss here how we should create events, and link alerts to it.

## Context
- An alert is: a signal sent by a device that localizes a wildfire. Natively, it holds: a timestamp, the localization information, the `device_id`
- an event holds: a type, a localization, a starting timestamp and an end timestamp

## Problem statement
- How do we determine the event an alert refers to?
- How do we update the starting & end timestamps of the event?

frgfm:
Here is a suggestion

### Linking an alert to an event
There are two subquestions:
#### Matching the alert to existing events:
1. we first fetch all events that have no `end_ts` to get our candidates
2. if an alert was raised by the same device in the last 5 minutes, we take the referred event
3. otherwise taking the GPS location + azimuth of the current alert, and considering a limit of 10km in that direction, we check whether existing active events were reported in that area.
#### whether we need to create a new event
if the above matching fails, then we create an event first, and the alert will refer to it.

### Updating the event
- start_ts: the timestamp of the first alert referring to this event
- end_ts: after a fixed duration (set in the config file most likely) since the last alert referring to this event, we set `end_ts` to the timestamp of this last alert.
- location: if alerts have been raised about the same event by different devices, using GPS + azimuth we can pinpoint an approximate location.

fe51:
Thanks for opening the issue ! This is a nice suggestion !

So the input`create_alert_from_device`might change in a way the `AlertBase` would not include the  `event_id` as long as considering the proposition, an alert is not yet related to an event when sent. 

The five minutes and the fixed duration are the same elt isn't it ?

frgfm:
We actually have the choice:
- either the `event_id` will be filled after the alert creation
- or we first resolve the event, then create the alert with the field value

Most likely yes !

frgfm:
Alright, I'll go ahead with 3 PRs:
- [ ] implement the event existence check with the 5 minutes duration + start_ts of the event
- [ ] implement the conversion device GPS + alert azimuth --> event location
- [ ] implement the end_ts resolution of an event

frgfm:
Lowering importance since we could first move forward with the sole relaxation duration criteria (event location resolution isn't critical for MVP)

frgfm:
Closed by #405 & #412
The GPS resolution is out of scope for now

PR #149, opened 2021-04-02
feat: Added event existence check upon alert creation

frgfm:
Following up on #146, this PR introduces the following modifications:
- fixed fetch_ongoing_alerts
- made `event_id` optional in alert creation payloads
- added a check of existing event referred by similar alerts -> if none exists, it creates an event, otherwise takes the latest
- updated client accordingly
- added corresponding unittests

Any feedback is welcome!


client/pyroclient/client.py
```diff
@@ -157,7 +163,14 @@ def create_no_alert_site(self, lat: float, lon: float, name: str, country: str,
...
         return requests.post(self.routes["no-alert-site"], headers=self.headers, json=payload)
 
-    def send_alert(self, lat: float, lon: float, event_id: int, device_id: int, media_id: int = None) -> Response:
+    def send_alert(
+        self,
+        lat: float,
+        lon: float,
+        device_id: int,
+        event_id: Optional[int] = None,
+        media_id: Optional[int] = None
```
fe51, line 172: By setting media_id optional, an alert can be sent without related media ? (I am just wandering if its related to this PR but why not ! )

fe51 (approved):
This PR looks good to me as long as it responds to the problem of linking an alert flow to an event !

Thanks for opening it !

client/pyroclient/client.py
frgfm: Oh no, this is something that we decided a long time ago :sweat_smile: (cf. the SQL table field ;) )
It's for the future case when someone notifies an alert but has no mean to send the picture

PR #151, opened 2021-04-02
feat: Added software_hash for device update check

### LINKED ISSUE #150: [devices] Implement an update mechanism for devices
In order to avoid requiring ssh access to our devices, it would be handy for them to check whether they need to update and automatically do it if necessary.

## Context
- the device needs to be able to check which software/docker version it needs
- depending on the result of the comparison with its current version, it might need to go download the software (outside of the scope of this repo)

## Implementation suggestion
I gave it some thoughts and here is what I suggest:
- similarly to the `specs` field (which we should rename to `hardware_specs`), we introduce a `software_hash` field.
- the corresponding value will be a deterministic result for a given software config (let's say the docker image hash, or the commit hash)
- just like devices that will ping regularly, the device will come check the `software_hash`. If it's different from the one it uses, then it updates.

_I strongly suggest going for the commit hash, at least to check the mechanism_

What do you think @pyronear/back-end ?

frgfm:
Following up on #150, this PR introduces the following modifications:
- adds a device route for self GET
- adds a `software_hash` string field to the `devices` table
- adds a route to update the `software_hash` of the current device
- adds corresponding unittests
- adds corresponding client methods

Resolves #150

Any feedback is welcome!

### DIFF
--- a/src/app/api/routes/devices.py
+++ b/src/app/api/routes/devices.py
@@ -28,7 +28,7 @@ router = APIRouter()
 
 
 @router.post("/", response_model=DeviceOut, status_code=201, summary="Create a new device")
-async def register_device(payload: DeviceAuth, _=Security(get_current_user, scopes=[AccessType.admin])):
+async def register_device(payload: DeviceAuth, _=Security(get_current_access, scopes=[AccessType.admin])):
     """Creates a new device based on the given information
 
     Below, click on "Schema" for more detailed information about arguments
@@ -54,7 +54,7 @@ async def register_my_device(
 
 
 @router.get("/{device_id}/", response_model=DeviceOut, summary="Get information about a specific device")
-async def get_device(device_id: int = Path(..., gt=0), _=Security(get_current_user, scopes=[AccessType.admin])):
+async def get_device(device_id: int = Path(..., gt=0), _=Security(get_current_access, scopes=[AccessType.admin])):
     """
     Based on a device_id, retrieves information about the specified device
     """
@@ -70,7 +70,7 @@ async def get_my_device(me: DeviceOut = Security(get_current_device, scopes=["de
 
 
 @router.get("/", response_model=List[DeviceOut], summary="Get the list of all devices")
-async def fetch_devices(_=Security(get_current_user, scopes=[AccessType.admin])):
+async def fetch_devices(_=Security(get_current_access, scopes=[AccessType.admin])):
     """
     Retrieves the list of all devices and their information
     """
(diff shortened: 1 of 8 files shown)


src/app/api/schemas.py
```diff
@@ -153,10 +153,15 @@ class EventOut(EventIn, _CreatedAt, _Id):
 
 
 # Device
+class SoftwareHash(BaseModel):
```
chloeskt, line 156: Quick question: how is the software hash computed ? (hash of the commit yes,  but of which repo ?) We have pyro-engine but also pyro-vision no ? How will the device be able to know which repo it needs to pull ?

src/app/api/routes/devices.py
```diff
@@ -140,6 +149,20 @@ async def update_my_location(
     return device
 
 
+@router.put("/hash", response_model=DeviceOut, summary="Update the software hash of the current device")
+async def update_my_hash(
+    payload: SoftwareHash, device: DeviceOut = Security(get_current_device, scopes=["device"])
+):
+    """
+    Updates the location of the current device
```
chloeskt, line 157: Just need to modify the docstring I think.

This gives to the device the latest hash it should have right ? Then the device will have to check if the hash it had before calling this method is the same as the new one ?

chloeskt (commented):
Hi @frgfm 
I just had a look at what you've done and have some quick questions !

src/app/api/schemas.py
frgfm: The hash is computed on device, so that the convention about how the deterministic function is computed is deferred to your choice :)
The comparison will be performed on the device anyway

src/app/api/routes/devices.py
frgfm: My bad, I got confused: the device itself should not be able to update its expected hash. It should only fetch the expected hash and check whether it's up-to-date. I'll fix this

src/app/api/schemas.py
chloeskt: Mhmm okay. Not sure yet how the hash will be computed, especially if we have several repo to update on the rpi. Don't know yet what will be installed on the rpi and what might need update. 

If there are several repo, how will the rpi how which repo to update ? If we have a unique hash (hash of hashes let's say), the rpi will update everything (pull all the repos using git or we can directly DL the source code as tar.gz for each repo).
Another way could be to add a parameter to the route to get the rpi to check the hash for every repo which might need updating.

We will need to discuss this @pyronear/hardware. Depending on our use case, I think that DL the source code might be easier (for instance we won't need to install and use git on the rpi) but then the .env file would need to be in another directory (not to get erase when DL the source code during the update).

However it is maybe too soon to talk about it and decide if we still don't know what form will take the tools that will be installed on the rpi (do we know to have the source code, will we use a Docker image etc etc ?) 

Anyways this is not the right place to discuss this. I will open an issue on the appropriate repo asap.

frgfm: Well that's the point: the computation should not be decided by the API :ok_hand: 
If there are several repos, it's rather easy : computing the hash of the concatenation of all git commit hashes.

But either way, in my opinion, as you said, this will not be handled on the API side

frgfm:
Any opinion @pyronear/back-end ? :pray:

frgfm:
@pyronear/back-end Let's get some reviews and merge this before we need to resolve conflicts again :sweat_smile: 
The hardware team needs this PR to be merged :pray:

florianriche (approved):

PR #152, opened 2021-04-03
feat: Added webhook system and alert notifications to third-parties

### LINKED ISSUE #145: [alerts] Add a callback on alert creation to provide developers with a webhook
In order to send notifications on multiple systems based on alert raising on the API-side, we need to provide developers with a webhook option. Another front-end service of pyronear is already in need for this feature cf. https://github.com/pyronear/pyro-platform/issues/38

## Implementation suggestion
Fortunately, FastAPI [callbacks](https://fastapi.tiangolo.com/advanced/openapi-callbacks/) seem to be quite appropriate for this use case. However, I'm not exactly sure how to make a webhook available using that.

frgfm:
Following up on #145, this PR introduces the following modifications:
- adds a `webhooks` table with the fields `id`, `callback` and `url`
- using the above, added an alert_notification background task upon alert creation
- added corresponding unittests

Resolves #145

Any feedback is welcome!

### DIFF
--- a/src/app/api/crud/webhooks.py
+++ b/src/app/api/crud/webhooks.py
@@ -1,0 +1,19 @@
+# Copyright (C) 2021, Pyronear contributors.
+
+# This program is licensed under the Apache License version 2.
+# See LICENSE or go to <https://www.apache.org/licenses/LICENSE-2.0.txt> for full license details.
+
+from typing import List
+
+from app.api import crud
+from app.db import webhooks
+
+
+async def fetch_webhook_urls(route: str) -> List[str]:
+
+    query = (
+        select([webhooks.c.url])
+        .where(webhooks.c.callback == route)
+    )
+
+    return await crud.base.database.fetch_all(query=query)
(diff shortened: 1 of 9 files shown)


src/app/api/crud/webhooks.py
```diff
@@ -0,0 +1,19 @@
...
+# This program is licensed under the Apache License version 2.
+# See LICENSE or go to <https://www.apache.org/licenses/LICENSE-2.0.txt> for full license details.
+
+from typing import List
+
+from app.api import crud
+from app.db import webhooks
+
+
+async def fetch_webhook_urls(route: str) -> List[str]:
```
florianriche, line 12: Why not using `fetch_all`?
Because we want to retrieve only the webhook? How about altering fetch_all to get only one field instead?

florianriche (dismissed):

src/app/api/crud/webhooks.py
frgfm: I hesitated but I didn't want to import the webhook table in the alert route code so I had to move it here anyway. Thought I'd go for the most efficient query possible in the process :sweat_smile:

florianriche (approved):

florianriche (approved):

PR #153, opened 2021-04-08
refactor: Make group_id mandatory 

florianriche:
This PR make it mandatory for an access to have a group_id.
The rest of the code has been updated accordingly:

NB: The Delete GROUP will delete all linked entity


frgfm:
Flake8 is failing, but also: is it a good idea to enforce group belonging for admins?

florianriche:
> Flake8 is failing, but also: is it a good idea to enforce group belonging for admins?

Why would it be an issue? IMHO, for consistency considerations, it would make sense to make it compulsory for all accesses objects. If you think admins should not be linked to a group, one could create a dedicated group for every admin.

frgfm:
@florianriche I disagree with transferring admin status check to group check. But perhaps it's not what you're suggesting? The access type is one filter for route restrictions, group belonging is another. They can be combined together, but I wouldn't mix their usage. Since admin are quite different from other users whatever their group.

Would you mind fixing flake8 please?

florianriche:
> @florianriche I disagree with transferring admin status check to group check. But perhaps it's not what you're suggesting? The access type is one filter for route restrictions, group belonging is another. They can be combined together, but I wouldn't mix their usage. Since admin are quite different from other users whatever their group.
> 
I just want to prevent the fields "group_id" from being null. That help us to make sure all user/device have a group and that our code is consistent. The goal of this PR was not replace access scopes by the group :) 


> Would you mind fixing flake8 please?
Done :)

src/app/api/crud/groups.py
```diff
@@ -76,3 +76,23 @@ async def _get_installation_group_id(entry_id: int) -> int:
...
+        # 1  Delete the devices
+        for access in group_accesses:
+            device = await crud.base.fetch_one(devices, {"access_id": access["id"]})
+            if device is not None:
+                await crud.accesses.delete_accessed_entry(devices, accesses, device["id"])
+        # 2  Delete the users
+        for access in group_accesses:
+            user = await crud.base.fetch_one(users, {"access_id": access["id"]})
+            if user is not None:
+                await crud.accesses.delete_accessed_entry(users, accesses, user["id"])
```
frgfm, line 94: I think this is way too long & heavy on the DB, I would suggest two alternatives:
- setting an "on_delete" link between access, users and devices (deleting the access will trigger the deletion of the other)
- fetching all device_ids with access_id "in" the list of group_accesses, and then add a multi-entry deletion crud function.

src/app/api/routes/devices.py
```diff
@@ -13,29 +13,32 @@
...
+async def register_device(payload: AdminDeviceAuth, _=Security(get_current_user, scopes=[AccessType.admin])):
     """Creates a new device based on the given information
 
     Below, click on "Schema" for more detailed information about arguments
     or "Example Value" to get a concrete idea of arguments
     """
     if await crud.get(payload.owner_id, users) is None:
         raise HTTPException(status_code=404, detail=f"Unknown user for owner_id={payload.owner_id}")
-    return await crud.accesses.create_accessed_entry(devices, accesses, payload, DeviceCreation)
+    device_payload = DeviceAuth(**payload.dict(), group_id=await get_entity_group_id(users, payload.owner_id))
```
frgfm, line 40: I suggest setting the owner's group id only if the admin didn't specify it in the payload

src/app/api/routes/users.py
```diff
@@ -99,4 +99,5 @@ async def delete_user(user_id: int = Path(..., gt=0), _=Security(get_current_use
     """
     Based on a user_id, deletes the specified user
     """
+    # TODO: Doesn't work when there is a owned device
```
frgfm, line 102: I get that this is temporary but I don't understand the comment?

frgfm (commented):

src/app/api/routes/users.py
florianriche: Will rephrase but it will yield an exception if the deleted user is the owner_id field of one device.

src/app/api/crud/groups.py
florianriche: Went for the first option. Much cleaner indeed, I forgot we could do that.

src/app/api/routes/devices.py
florianriche: Done but does it make sens to have a group_id different from the group of the owner?

frgfm: For the admin, I think it does: if a user isn't able to register her/his device, the admin will be able to do so while registering the correct group

frgfm (changes requested):
Thanks for the edits! I think you only need to update the end-to-end script and we're good to go!

frgfm (approved):
Thanks!

PR #156, opened 2021-04-15
Refactor: Use SQLAlchemy models instead of Table Declaration

### LINKED ISSUE #154: Refactor: Use SQLAlchemy DataModels
See usage of SQLAlchemy and Pydanic here: https://fastapi.tiangolo.com/tutorial/sql-databases/
That would greatly simplify complex queries that we are starting to have.

Refactor might be quite heavy

florianriche:
### Context

It is hard to do complex queries and to reason with our current ORM abstraction. In particular, if one wants to handle Group accesses, we end up creating a lot of boilerplate, cumbersome code. The more complex queries (>2 tables involves) and the joins are hard to handle in our current implementation.

### Content of the PR:
Inspired by https://fastapi.tiangolo.com/tutorial/sql-databases/

It ended up being far easier and short than expected.
* Migrate the fields definition from `tables.py` to a newly created file `models.py`
* For legacy consistency, set references to the Models tables in `tables.py`
* Adapt Metadata calls in test and app.py
* Add function to get a DB session in a route

We can now add more complex functionalities in future PR.

Closes #154


src/app/api/routes/users.py
```diff
@@ -65,7 +65,7 @@ async def get_user(user_id: int = Path(..., gt=0), _=Security(get_current_user,
 
 
 @router.get("/", response_model=List[UserRead], summary="Get the list of all users")
-async def fetch_users(_=Security(get_current_user, scopes=[AccessType.admin])):
+async def fetch_users(_=Security(get_current_user, scopes=[AccessType.admin]), session=Depends(get_session)):
     """
```
florianriche, line 69: Adding this dependency one can have access to a session

src/app/db/tables.py
```diff
@@ -3,162 +3,24 @@
...
+groups = Groups.__table__
+sites = Sites.__table__
+events = Events.__table__
+devices = Devices.__table__
+media = Media.__table__
+installations = Installations.__table__
+alerts = Alerts.__table__
+webhooks = Webhooks.__table__
+
+metadata = Base.metadata
```
frgfm, line 26: Is this remaining only for the transition or will we have to keep it?

frgfm (commented):
Thanks for the PR! Only a little question since I'm not an expert in ORM!

src/app/db/tables.py
florianriche: It remains to keep the legacy structure of the code and CRUDs. Maybe we will be able to migrate in a latter PR but that will be a heavy one. 
For now, I would prefer to just leverage the ORM capacities to ease some implementation  (e.g the fetch_all with the same group_id).

frgfm: Makes sense, thanks!

src/app/api/routes/users.py
```diff
@@ -5,10 +5,10 @@
 
 from typing import List
 
-from fastapi import APIRouter, Path, Security, HTTPException, status
+from fastapi import APIRouter, Path, Security, HTTPException, status, Depends
 
 from app.api import crud
-from app.db import users, accesses
+from app.db import users, accesses, get_session, models
```
frgfm, line 11: no need to import models here

src/app/db/models.py
```diff
@@ -0,0 +1,146 @@
+# Copyright (C) 2021, Pyronear contributors.
+
+# This program is licensed under the Apache License version 2.
+# See LICENSE or go to <https://www.apache.org/licenses/LICENSE-2.0.txt> for full license details.
+
+import enum
+from .session import Base
+from sqlalchemy.sql import func
+from sqlalchemy import Column, DateTime, Integer, Float, String, Table, Enum, Boolean, ForeignKey, MetaData
```
frgfm, line 9: no need to import Table

src/app/db/tables.py
```diff
@@ -3,162 +3,24 @@
 # This program is licensed under the Apache License version 2.
 # See LICENSE or go to <https://www.apache.org/licenses/LICENSE-2.0.txt> for full license details.
 
-import enum
-from sqlalchemy import Column, DateTime, Integer, Float, String, Table, Enum, Boolean, ForeignKey, MetaData
-from sqlalchemy.sql import func
 
+from .session import Base
+from .models import (Users, Accesses, Alerts, Devices, Events,
+                     Groups, Installations, Media, Sites, Webhooks, EventType, AccessType, SiteType, MediaType)
```
frgfm, line 9: no need to import the Types

frgfm (changes requested):
Just a few unused imports left and we're good to go!

frgfm (approved):
Thanks!

PR #157, opened 2021-04-17
Client: Add azimuth in alerts functions

florianriche:
Small PR to add the azimuth information in the client messages.

Also add default example values to the Alert Schemas for the software_hash and the azimuth.

### DIFF
--- a/client/pyroclient/client.py
+++ b/client/pyroclient/client.py
@@ -173,6 +173,7 @@ class Client:
         lat: float,
         lon: float,
         device_id: int,
+        azimuth: Optional[float] = None,
         event_id: Optional[int] = None,
         media_id: Optional[int] = None
     ) -> Response:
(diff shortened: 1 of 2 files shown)


src/app/api/schemas.py
```diff
@@ -163,8 +163,8 @@ class SoftwareHash(BaseModel):
 class MyDeviceIn(Login, DefaultPosition):
     specs: str = Field(..., min_length=3, max_length=100, example="systemV0.1")
     last_ping: datetime = Field(default=None, example=datetime.utcnow())
-    angle_of_view: float = Field(..., gt=0, lt=360)
-    software_hash: str = Field(None, min_length=8, max_length=16)
+    angle_of_view: float = Field(..., gt=0, lt=360, example=10)
+    software_hash: str = Field(None, min_length=8, max_length=16, example="0123456789ABCDEF")
```
Akilditu, line 167: Just for me to understand, is this linked to the azimuth ?

Akilditu (dismissed):
Thanks a lot for this PR and adding this to the client !

src/app/api/schemas.py
florianriche: No but I took rthe opportunity.  It os not worth a Pr

client/pyroclient/client.py
```diff
@@ -222,6 +227,7 @@ def send_alert_from_device(
         Args:
             lat: the latitude of the alert
             lon: the longitude of the alert
+            -azimuth: the azimuth of the alert
```
frgfm, line 230: typo here, the hyphen isn't needed

frgfm (changes requested):
Thanks for the PR! Only a small typo in the docstring of the client but the rest looks good!

frgfm (approved):
Thanks for the edits!

PR #159, opened 2021-04-19
chore: Upgraded Python to 3.7

florianriche:
Since the introduction of #156 , the API can't be deploy because one of the introduced feature is not compatible with python 3.6 
See Heroku log stack trace:

image

This PR upgrade the minimum python version to 3.7


frgfm (approved):
Thanks for the fix!

PR #160, opened 2021-04-20
fix: Revert "Refactor: Use SQLAlchemy models instead of Table Declaration"

florianriche:
Reverts pyronear/pyro-api#156 because it relies on python 3.7 and it crashes the API


frgfm (approved):
Thanks!

PR #162, opened 2021-04-20
Refactor: Use SQLAlchemy models instead of Table Declaration (after fix)

florianriche:
Reverts pyronear/pyro-api#160


frgfm (approved):
I just checked if this branch is working on Heroku and it's all good now!

PR #163, opened 2021-04-20
chore: Reduced python version of the client to 3.6

florianriche:
We migrate to python 3.7 for the API. At the same time we migrated the client python version while it is not required and created a bug.

This PR downgrade the python version of the client package back to python 3.6


frgfm (approved):
Thanks!

fe51 (approved):

PR #164, opened 2021-04-21
feat: add default group_id

florianriche:
When we don't specify a group_id in inpout when creating a site, it yields an exception.
This PR adds a check that:
1/ The requester is allowed to specify a group (is admin or belongs to the group_id set in the payload)
2/ Add the group_id of the requester in the payload if the group_id is not specified.

This is done for both the standard site creation and the noAlert sites.


src/tests/routes/test_sites.py
```diff
@@ -97,9 +97,9 @@ async def test_fetch_sites(test_app_asyncio, init_test_db):
          201, None],
         [1, {"name": "my_site", "group_id": 1, "lat": 0., "lon": 0.,
              "country": "FR", "geocode": "01"}, True, 201, None],
-        [0, {"name": "my_site", "group_id": 1, "lat": 0., "lon": 0.,
+        [0, {"name": "my_site", "group_id": 2, "lat": 0., "lon": 0.,
              "country": "FR", "geocode": "01"}, False,
-         401, "Permission denied"],
+         401, "You can't specify another group"],
```
frgfm, line 102: Sorry, I'm not sure I understand the logic: did the test case change?

If possible, I'd rather keep the group_id 1 and change the expected outcome if it did, and then adding a test case for group_id 2

frgfm (changes requested):

src/tests/routes/test_sites.py
florianriche: Done

src/app/api/routes/sites.py
```diff
@@ -15,23 +16,37 @@
 
 
 @router.post("/", response_model=SiteOut, status_code=201, summary="Create a new site")
-async def create_site(payload: SiteIn, _=Security(get_current_access, scopes=[AccessType.admin])):
+async def create_site(payload: SiteIn,
+                      requester=Security(get_current_access, scopes=[AccessType.admin, AccessType.user])):
```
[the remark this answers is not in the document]
florianriche: Done

fe51 (approved):
Thanks for the PR !

PR #165, opened 2021-04-22
[Sites] Fix noAlertSite route creation with default group_id

florianriche:
Fix the route and alter the UTest


frgfm (approved):
Thanks for the fix!

PR #166, opened 2021-04-27
[groups] Add group logic to fetching routes

florianriche:
# Context
Now that we have added SQLAlchemy models, we can now use them to ease the implementation of queries that rely on multiple joins.
Among those queries are the fetching routes that must return results only limited in a specified group_id.

# Content of the PR
* Add relationships in SQLALchemy models
* Add group_id in standard fetching routes (not yet implemented for all fetching routes)
* Adapted the unit Tests
* Fixed the relationship between events/alerts and the deletion cascading function (previous behavior was buggy.

## Reviews
Can you please check that:
* You agree with this new fetching behavior
* You are fine with the implementation and code organization

## Next steps
* PR to include the group_id logic in the remaining fetching functions
* PR to include the group_id logic in the update functions

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -23,7 +23,7 @@ Client for the [Pyronear API](https://github.com/pyronear/pyro-api)
 
 ### Prerequisites
 
-- Python 3.6 (or more recent)
+- Python 3.6   (or more recent)
 - [pip](https://pip.pypa.io/en/stable/)
 
 ### Installation
--- a/client/README.md
+++ b/client/README.md
@@ -23,7 +23,7 @@ Client for the [Pyronear API](https://github.com/pyronear/pyro-api)
 
 ### Prerequisites
 
-- Python 3.6   (or more recent)
+- Python 3.6 (or more recent)
 - [pip](https://pip.pypa.io/en/stable/)
 
 ### Installation
--- a/src/app/api/routes/alerts.py
+++ b/src/app/api/routes/alerts.py
@@ -107,7 +107,6 @@ async def fetch_alerts(requester=Security(get_current_access, scopes=[AccessType
                             .join(models.Accesses)
                             .filter(models.Accesses.group_id == requester.group_id).all())
         retrieved_alerts = [x.__dict__ for x in retrieved_alerts]
-        print("retrieved_alert:", retrieved_alerts)
         return retrieved_alerts
 
 
--- a/src/app/api/routes/alerts.py
+++ b/src/app/api/routes/alerts.py
@@ -94,7 +94,7 @@ async def get_alert(alert_id: int = Path(..., gt=0), _=Security(get_current_acce
 
 
 @router.get("/", response_model=List[AlertOut], summary="Get the list of all alerts")
-async def fetch_alerts(requester=Security(get_current_access, scopes=[AccessType.admin, AccessType.user]), 
+async def fetch_alerts(requester=Security(get_current_access, scopes=[AccessType.admin, AccessType.user]),
                        session=Depends(get_session)):
     """
     Retrieves the list of all alerts and their information
@@ -102,12 +102,12 @@ async def fetch_alerts(requester=Security(get_current_access, scopes=[AccessType
     if await is_admin_access(requester.id):
         return await crud.fetch_all(alerts)
     else:
-        print(session.query(models.Alerts).first())
         retrieved_alerts = (session.query(models.Alerts)
                             .join(models.Devices)
                             .join(models.Accesses)
                             .filter(models.Accesses.group_id == requester.group_id).all())
         retrieved_alerts = [x.__dict__ for x in retrieved_alerts]
+        print("retrieved_alert:", retrieved_alerts)
         return retrieved_alerts
 
 
--- a/src/app/api/routes/alerts.py
+++ b/src/app/api/routes/alerts.py
@@ -4,13 +4,13 @@
 # See LICENSE or go to <https://www.apache.org/licenses/LICENSE-2.0.txt> for full license details.
 
 from typing import List
-from fastapi import APIRouter, Path, Security, HTTPException, status, BackgroundTasks
+from fastapi import APIRouter, Path, Security, HTTPException, status, BackgroundTasks, Depends
 from sqlalchemy import select
 from functools import partial
 from datetime import datetime, timedelta
 
 from app.api import crud
-from app.db import alerts, events, media
+from app.db import alerts, events, media, get_session, models
 from app.api.schemas import (
     AlertBase,
     AlertOut,
@@ -22,6 +22,7 @@ from app.api.schemas import (
     AccessType)
 from app.api.deps import get_current_device, get_current_access
 from app.api.external import post_request
+from app.api.crud.authorizations import is_admin_access, is_in_same_group
 
 
 router = APIRouter()
@@ -93,11 +94,21 @@ async def get_alert(alert_id: int = Path(..., gt=0), _=Security(get_current_acce
 
 
 @router.get("/", response_model=List[AlertOut], summary="Get the list of all alerts")
-async def fetch_alerts(_=Security(get_current_access, scopes=[AccessType.admin])):
+async def fetch_alerts(requester=Security(get_current_access, scopes=[AccessType.admin, AccessType.user]), 
+                       session=Depends(get_session)):
     """
     Retrieves the list of all alerts and their information
     """
-    return await crud.fetch_all(alerts)
+    if await is_admin_access(requester.id):
+        return await crud.fetch_all(alerts)
+    else:
+        print(session.query(models.Alerts).first())
+        retrieved_alerts = (session.query(models.Alerts)
+                            .join(models.Devices)
+                            .join(models.Accesses)
+                            .filter(models.Accesses.group_id == requester.group_id).all())
+        retrieved_alerts = [x.__dict__ for x in retrieved_alerts]
+        return retrieved_alerts
 
 
 @router.put("/{alert_id}/", response_model=AlertOut, summary="Update information about a specific alert")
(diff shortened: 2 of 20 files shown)


client/README.md
```diff
@@ -23,7 +23,7 @@ Client for the [Pyronear API](https://github.com/pyronear/pyro-api)
 
 ### Prerequisites
 
-- Python 3.6 (or more recent)
+- Python 3.6   (or more recent)
```
frgfm, line 26: ? why having two spaces?

frgfm:
It's gonna be tough to review, but I'll try my best :sweat_smile:

src/app/api/routes/alerts.py
```diff
@@ -22,6 +22,7 @@
     AccessType)
 from app.api.deps import get_current_device, get_current_access
 from app.api.external import post_request
+from app.api.crud.authorizations import is_admin_access, is_in_same_group
```
frgfm, line 25: The 2nd import isn't used

src/app/api/routes/alerts.py
```diff
@@ -93,11 +94,20 @@ async def get_alert(alert_id: int = Path(..., gt=0), _=Security(get_current_acce
 
 
 @router.get("/", response_model=List[AlertOut], summary="Get the list of all alerts")
-async def fetch_alerts(_=Security(get_current_access, scopes=[AccessType.admin])):
+async def fetch_alerts(requester=Security(get_current_access, scopes=[AccessType.admin, AccessType.user]),
+                       session=Depends(get_session)):
```
frgfm, line 98: If it doesn't fit on one line, let's switch to black formatting instead for consistency :ok_hand:

src/app/api/routes/alerts.py
```diff
@@ -93,11 +94,20 @@ async def get_alert(alert_id: int = Path(..., gt=0), _=Security(get_current_acce
...
     Retrieves the list of all alerts and their information
     """
-    return await crud.fetch_all(alerts)
+    if await is_admin_access(requester.id):
+        return await crud.fetch_all(alerts)
+    else:
+        retrieved_alerts = (session.query(models.Alerts)
+                            .join(models.Devices)
+                            .join(models.Accesses)
+                            .filter(models.Accesses.group_id == requester.group_id).all())
```
frgfm, line 108: SQL wise, it would be faster & ore efficient to :
- fetch access_ids corresponding to an access in the same group
- fetch device_ids whose access_id is in the previous list
- finally fetching all alerts whose device_id is in the previous list

src/app/api/routes/devices.py
```diff
@@ -26,6 +26,9 @@
 from app.api.crud.groups import get_entity_group_id
 
 
+from app.api.crud.authorizations import is_admin_access, is_in_same_group
```
frgfm, line 29: same here

src/app/api/routes/devices.py
```diff
@@ -26,6 +26,9 @@
 from app.api.crud.groups import get_entity_group_id
```
frgfm, line 28: no need for these blank lines

src/app/api/routes/devices.py
```diff
@@ -76,11 +79,21 @@ async def get_my_device(me: DeviceOut = Security(get_current_device, scopes=["de
...
+async def fetch_devices(requester=Security(get_current_access, scopes=[AccessType.admin, AccessType.user]),
+                        session=Depends(get_session)):
     """
     Retrieves the list of all devices and their information
     """
-    return await crud.fetch_all(devices)
+    if await is_admin_access(requester.id):
+        return await crud.fetch_all(devices)
+    else:
+        print(session.query(models.Devices).first().access)
```
frgfm, line 90: I think there's a debug print left here

src/app/api/routes/devices.py
```diff
@@ -76,11 +79,21 @@ async def get_my_device(me: DeviceOut = Security(get_current_device, scopes=["de
...
+        retrieved_devices = (session.query(models.Devices)
+                             .join(models.Accesses)
+                             .filter(models.Accesses.group_id == requester.group_id).all())
```
frgfm, line 93: same comment here

src/app/api/routes/alerts.py
frgfm: Alternatively, when joining, only take the columns that you need. Otherwise we keep information and use memory that doesn't need to

src/app/api/routes/media.py
```diff
@@ -67,12 +68,21 @@ async def get_media(media_id: int = Path(..., gt=0), requester=Security(get_curr
...
-    return await crud.fetch_all(media)
+    if await is_admin_access(requester.id):
+        return await crud.fetch_all(media)
+    else:
+        retrieved_media = (session.query(models.Media)
+                                  .join(models.Devices)
+                                  .join(models.Accesses)
+                                  .filter(models.Accesses.group_id == requester.group_id).all())
+        retrieved_media = [x.__dict__ for x in retrieved_media]
+        print("retrieved_media:", retrieved_media)
```
frgfm, line 84: to be removed

src/tests/routes/test_alerts.py
```diff
@@ -36,7 +36,7 @@
 ACCESS_TABLE = [
     {"id": 1, "group_id": 1, "login": "first_login", "hashed_password": "hashed_pwd", "scope": "user"},
     {"id": 2, "group_id": 1, "login": "second_login", "hashed_password": "hashed_pwd", "scope": "admin"},
-    {"id": 3, "group_id": 2, "login": "third_login", "hashed_password": "hashed_pwd", "scope": "device"},
+    {"id": 3, "group_id": 1, "login": "third_login", "hashed_password": "hashed_pwd", "scope": "device"},
```
frgfm, line 39: mind adding a comment when modifying a test case? (hard to review without reading again all the unittests :sweat_smile: )

src/tests/routes/test_devices.py
```diff
@@ -9,12 +9,13 @@
...
 from app import db
 from app.api import crud, security
-from tests.db_utils import get_entry, fill_table
+from tests.db_utils import get_entry, fill_table, TestSessionLocal
 from tests.utils import update_only_datetime, parse_time
 
 USER_TABLE = [
     {"id": 1, "login": "first_login", "access_id": 1, "created_at": "2020-10-13T08:18:45.447773"},
     {"id": 2, "login": "second_login", "access_id": 2, "created_at": "2020-11-13T08:18:45.447773"},
+    {"id": 3, "login": "sixth_login", "access_id": 5, "created_at": "2020-11-13T08:18:45.447773"},
```
frgfm, line 18: This entry and the one you added in the device table can't share the same access_id

src/tests/routes/test_devices.py
```diff
@@ -24,6 +25,9 @@
     {"id": 2, "login": "fourth_login", "owner_id": 2, "access_id": 4, "specs": "v0.1", "elevation": None, "lat": None,
      "lon": None, "yaw": None, "pitch": None, "last_ping": None, "angle_of_view": 68., "software_hash": None,
      "created_at": "2020-10-13T08:18:45.447773"},
+    {"id": 3, "login": "fifth_login", "owner_id": 3, "access_id": 5, "specs": "v0.1", "elevation": None, "lat": None,
```
frgfm, line 28: access_id already in use by user 3

frgfm (commented):
Thanks for the PR! I added comments with suggestions, let me know what you think!

src/app/api/routes/alerts.py
florianriche: You mean applying black?
So get rid of flake8?

src/tests/routes/test_alerts.py
florianriche: Ok, here it was to have one device in the group_id 1 (otherwise we have not alert linked to group 1)

src/app/api/routes/alerts.py
frgfm: I mean, if you split the signature into multiple lines, then let's use one line per argument :)

src/tests/routes/test_alerts.py
frgfm: alright, and did the previous test case output change? I'm trying to check why we can't keep both / if it's useful

src/app/api/routes/alerts.py
florianriche: The joins are done automatically on the ForeignKey columns and only the columns of the queried objects (here alerts) are retrieved.  The Database does the underlying optimization

Here is the statement for the event query (accessible simply by doing (str(<sqlalchemyQuery>) ) for example:
```
SELECT events.id AS events_id, events.lat AS events_lat, events.lon AS events_lon, events.type AS events_type, events.start_ts AS events_start_ts, events.end_ts AS events_end_ts, events.created_at AS events_created_at 
FROM events JOIN alerts ON events.id = alerts.event_id JOIN devices ON devices.id = alerts.device_id JOIN accesses ON accesses.id = devices.access_id 
WHERE accesses.group_id = %(group_id_1)s
```

src/app/api/routes/devices.py
florianriche: same reply as in the other comment, only the devices columns are retrieved; and doing a single SQL query can't be beaten.

src/tests/routes/test_alerts.py
florianriche: Previously the group_id was not used so I just added group_id:1 to all accesses object just so that the field is not empty. No Utest was using it (in test_alerts at least)

src/app/api/routes/alerts.py
florianriche: Ok done

src/app/api/routes/alerts.py
frgfm: Oh damn, that's neat!

frgfm (approved):
Thanks a lot!

PR #167, opened 2021-04-30
[Groups] Add group logic to remaining routes

### LINKED ISSUE #125: [users] Implementing user_group

florianriche:
## Context 
This PR is the last part of the group implementation. After having implemented the group logic for the fetching route, we have added the group logic to the remaining routes.

## Content:
* Group logic in Sites
* Group logic in Installations
* Group logic in Alerts
* Group logic in Events
* Group logic in Media

Routes for which I have a strong doubt have a # TODO  comment in them.

The commits are separated to ease the review, there is one commit per object.

This PR should close #125


src/app/api/routes/events.py
```diff
@@ -5,11 +5,13 @@
 
 from typing import List
 from fastapi import APIRouter, Path, Security, status, HTTPException, Depends
+from sqlalchemy import and_, or_
```
frgfm, line 8: `or_` doesn't need to be imported

src/app/api/crud/groups.py
```diff
@@ -40,39 +40,55 @@ async def get_entity_group_id(table: Table, entry_id: int) -> int:
 
 async def _get_access_group_id(entry_id: int) -> int:
     entry = await crud.base.get_entry(accesses, entry_id)
+    if entry is None:
+        return None
     return entry["group_id"]
```
frgfm, line 45: Three things to consider : 
- if we keep this logic, the implementation is fine by me, we can also switch to a one-liner
```
return entry['group_id'] if isinstance(entry, dict) else None
```
- only a question but : shouldn't we raise an error if the entry-id doesn't exist?
- if we want to start optimizing SQL queries, here we actually only need the group_id field

src/app/api/routes/alerts.py
```diff
@@ -81,15 +81,18 @@ async def create_alert_from_device(
...
     """
-
     return await create_alert(AlertIn(**payload.dict(), device_id=device.id), background_tasks)
 
 
 @router.get("/{alert_id}/", response_model=AlertOut, summary="Get information about a specific alert")
-async def get_alert(alert_id: int = Path(..., gt=0), _=Security(get_current_access, scopes=[AccessType.admin])):
+async def get_alert(alert_id: int = Path(..., gt=0),
+                    requester=Security(get_current_access,
+                    scopes=[AccessType.admin, AccessType.user])):
```
frgfm, line 90: Let's keep black formatting when we have several lines:
```
async def get_alert(
    alert_id: int = Path(..., gt=0),
    requester=Security(get_current_access, _=[AccessType.admin, AccessType.user])
 ):
```

src/app/api/routes/alerts.py
```diff
@@ -81,15 +81,18 @@ async def create_alert_from_device(
...
     """
     Based on a alert_id, retrieves information about the specified alert
     """
+    requested_group_id = await get_entity_group_id(alerts, alert_id)
+    await check_group_read(requester.id, requested_group_id)
     return await crud.get_entry(alerts, alert_id)
```
frgfm, line 96: over these 3 lines, we are querying the alerts table twice, we should try to improve this I think

src/app/api/routes/alerts.py
```diff
@@ -112,22 +115,26 @@ async def fetch_alerts(requester=Security(get_current_access,
...
 
 @router.put("/{alert_id}/", response_model=AlertOut, summary="Update information about a specific alert")
-async def update_alert(
-    payload: AlertIn,
-    alert_id: int = Path(..., gt=0),
-    _=Security(get_current_access, scopes=[AccessType.admin])
-):
+async def update_alert(payload: AlertIn,
+                       alert_id: int = Path(..., gt=0),
+                       requester=Security(get_current_access, scopes=[AccessType.admin, AccessType.user])):
```
frgfm, line 120: Let's keep black formatting for consistency

src/app/api/routes/alerts.py
```diff
@@ -112,22 +115,26 @@ async def fetch_alerts(requester=Security(get_current_access,
...
     """
     Based on a alert_id, updates information about the specified alert
     """
+    requested_group_id = await get_entity_group_id(alerts, alert_id)
+    await check_group_update(requester.id, requested_group_id)
     return await crud.update_entry(alerts, payload, alert_id)
```
frgfm, line 126: same here

src/app/api/routes/sites.py
```diff
@@ -71,11 +72,14 @@ async def fetch_sites(requester=Security(get_current_access,
 async def update_site(
     payload: SiteIn,
     site_id: int = Path(..., gt=0),
-    _=Security(get_current_access, scopes=[AccessType.admin])
+    requester=Security(get_current_access, scopes=[AccessType.admin, AccessType.user])
 ):
     """
     Based on a site_id, updates information about the specified site
     """
+    # TODO: validate this one
```
frgfm, line 80: validate the general design of group checking you mean?

src/tests/routes/test_alerts.py
```diff
@@ -89,11 +91,12 @@ async def init_test_db(monkeypatch, test_db):
 @pytest.mark.parametrize(
     "access_idx, alert_id, status_code, status_details",
     [
-        [0, 1, 401, "Permission denied"],
+        [0, 1, 200, None],
         [1, 1, 200, None],
         [2, 1, 401, "Permission denied"],
         [1, 999, 404, "Entry not found"],
         [1, 0, 422, None],
+        [4, 1, 401, "You can't access this ressource"],
```
frgfm, line 99: A "Permission denied" would be better to unify error messages

frgfm (commented):
Thanks for the PR!
Overall, I think we only need to change route signature formatting (stick to black when it's not a one-liner), and most importantly, optimize the group checking part of this PR. I added a few comments for this

src/app/api/routes/alerts.py
florianriche: Not sure I get it, didn't you ask in another review to have several lines in those cases?

src/app/api/routes/alerts.py
florianriche: Only way I see to slightly speed up the process would be to do 
```python

    retrieved_alert = await crud.get_entry(alerts, alert_id)
    requested_group_id = await get_entity_group_id(devices, retrieved_alert["device_id"])
    await check_group_read(requester.id, requested_group_id)
    return retrieved_alert
```

src/app/api/routes/alerts.py
florianriche: same suggestion as above roughly, not sure we can really speed up the process

src/app/api/routes/sites.py
florianriche: Validate that the route need to haveccess restrcted to group

src/tests/routes/test_alerts.py
florianriche: HMm, I personnally think that it would be helpful for the user to let him know that he is allowed to use that route but that he should check its input

src/app/api/routes/alerts.py
frgfm: Sure, but 1 line per function argument
Besides, I think we can annotation typing to `requester`

src/tests/routes/test_alerts.py
frgfm: Perhaps let's make it a more detailed message then, because I understand it the same way as "Permission denied" (which can be applied to route or sub entry getting)

What I mean is: the route will fail anyway, the user won't know where and in one case that person will get "Permission denied" and another "You can't access this resource". I would understand that they meant there is a difference but I wouldn't have any clue what it is :sweat_smile:

frgfm (approved):
Thanks for the PR! If you feel like it, I think we need to add a good explanation of how access is handled in the API in our README. While our documentation is very detailed for each route, the general process might sound obscure to most

PR #172, opened 2021-06-05
feat: Added console log to notify that Sentry is enabled

frgfm:
This PR simply adds a console log message when Sentry is enabled and correctly set up


fe51 (approved):
Looks good to me ! Thanks

PR #173, opened 2021-06-05
docs: Fixed client docstring examples

frgfm:
This PR simply fixes the docstring examples from the API client


fe51 (approved):
Thanks Fg !

PR #174, opened 2021-06-07
feat: Added nginx configuration in docker orchestration

frgfm:
- moved core API dockerfile in `src/` and refactored it using [tiangolo Docker images](https://github.com/tiangolo/uvicorn-gunicorn-fastapi-docker)
- added nginx configuration to secure the API
- unpinned FastAPI version to get recent updates
- updates corresponding CI jobs
- updates usage instructions

Any feedback is welcome!


.github/workflows/main.yml
```diff
@@ -87,7 +87,7 @@ jobs:
           BUCKET_NAME: ${{ secrets.BUCKET_NAME }}
           BUCKET_MEDIA_FOLDER: ${{ secrets.BUCKET_MEDIA_FOLDER }}
         run: |
-          PORT=8002 docker-compose up -d --build
```
florianriche, line 90: Why change the port?
Also, setting it would "document" that the number "8080"  in e2E test is a port

docker-compose.yml
```diff
@@ -1,39 +1,48 @@
...
+      - pyroapi
+  db_test:
     image: postgres:12.1-alpine
     volumes:
       - postgres_data_test:/var/lib/postgresql/data_test/
     environment:
-      - POSTGRES_USER=pyro_api_test
-      - POSTGRES_PASSWORD=pyro_api_test
-      - POSTGRES_DB=pyro_api_dev_test
+      - POSTGRES_USER=dummy_pg_tuser
```
florianriche, line 43: tuser, tpwd, tdb as "test_user"... ?

florianriche (approved):
Looks good to me

docker-compose.yml
frgfm: Correct!

.github/workflows/main.yml
frgfm: It's more that I remove the port flexibility, since it can be changed in the docker-compose. So I picked a 80xx port out of convenience. Not sure about what you mean by "document" ?

PR #177, opened 2021-06-13
chore: Extended timeout on uvicorn

frgfm:
This PR increases the timeout of uvicorn from 5 sec to 30sec


florianriche (approved):
I haven't followed the subject, why do we want to increase to 30 seconds?

PR #179, opened 2021-07-06
feat: Upgraded admin token to unlimited duration

frgfm:
This PR removes the expiration of admin tokens so that master services (such as the platform) don't get issues with API access.

Any feedback is welcome!


frgfm:
This is actually quite dangerous, closing this in favour of #180

PR #180, opened 2021-07-06
feat: Added token refresh method to client

frgfm:
- added token refresh method to API client
- added unittest

Any feedback is welcome!

### DIFF
--- a/client/pyroclient/client.py
+++ b/client/pyroclient/client.py
@@ -84,7 +84,7 @@ class Client:
         self.headers = {"Authorization": f"Bearer {self.token}"}
 
     def refresh_token(self, login: str, password: str) -> None:
-        self.token = self._retrieve_token(credentials_login, credentials_password)
+        self.token = self._retrieve_token(login, password)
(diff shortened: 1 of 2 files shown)


fe51 (commented):
Hi There, thanks for the PR !

Looks okay to me. 
However, just a question (sorry I might have not been following recent update), but what is the aim of this update ? Does it solve an issue from any other team ?

frgfm:
> Hi There, thanks for the PR !
> 
> Looks okay to me.
> However, just a question (sorry I might have not been following recent update), but what is the aim of this update ? Does it solve an issue from any other team ?

Hi @fe51,

Yes, it allows instantiated client to refresh their token that expire. For the platform this happens since user do not have unlimited tokens :)

fe51:
Okay perfect then :)

Thanks a lot

fe51 (approved):

Akilditu (approved):
Thanks for this PR that will ensure the platform longtime run !

PR #182, opened 2021-07-11
feat: Specified HTTP status code and error messages

frgfm:
Following up on #161, this PR introduces the following modifications:
- fixes the instruction to run the unittest in CONTRIBUTING
- Specified error codes (cf. https://kinsta.com/blog/http-status-codes/): 404 when an entry isn't found, 401 when token expires or creds aren't valid, 409 when there is a conflict, 403 when the authorization level doesn't allow an operation
- Added detailed error messages for each `HTTPException`
- updated unittest accordingly

Closes #161 

Any feedback is welcome!

### DIFF
--- a/src/app/api/crud/groups.py
+++ b/src/app/api/crud/groups.py
@@ -5,12 +5,13 @@
 
 
 from sqlalchemy import Table
+from typing import Optional
 
 from app.api import crud
 from app.db import accesses, users, sites, events, devices, media, installations, alerts
 
 
-async def get_entity_group_id(table: Table, entry_id: int) -> int:
+async def get_entity_group_id(table: Table, entry_id: int) -> Optional[int]:
 
     if table == accesses:
         return await _get_access_group_id(entry_id)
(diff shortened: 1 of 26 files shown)


src/app/api/crud/groups.py
```diff
@@ -29,31 +30,25 @@ async def get_entity_group_id(table: Table, entry_id: int) -> int:
...
 async def _get_user_group_id(entry_id: int) -> int:
     entry = await crud.base.get_entry(users, entry_id)
-    if entry is None:
-        return None
     return await _get_access_group_id(entry["access_id"])
 
 
 async def _get_site_group_id(entry_id: int) -> int:
     entry = await crud.base.get_entry(sites, entry_id)
-    if entry is None:
```
florianriche, line 51: won't it generate errors if entry is none andn we try to call entry["ghroup_id"]?

src/app/api/deps.py
```diff
@@ -48,37 +40,53 @@ async def get_current_access(security_scopes: SecurityScopes, token: str = Depen
...
+    try:
+        access_id = int(payload["sub"])
+        token_scopes = payload.get("scopes", [])
+        token_data = TokenPayload(access_id=access_id, scopes=token_scopes)
+    except (KeyError, ValidationError):
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+            detail="Invalid token payload.",
+            headers={"WWW-Authenticate": authenticate_value},
+        )
```
florianriche, line 59: Don't you want to reuse unauthorized-exception with an additionnal argument for the status code ?

florianriche (approved):

src/app/api/crud/groups.py
frgfm: the `crud.base.get_entry` cannot return a `None`. If it founds a `None` entry, the function itself throws a 404 error, so I remove the case that won't ever happen :)

src/app/api/deps.py
frgfm: The point of the PR is to get more informative / specific error messages and status code. Apart from the header, there isn't much in common with other exceptions that are thrown here :thinking: 

If it fails here, the token hasn't expired but for some reasons, its payload is invalid (missing entries in the dictionary most likely), so I figured it's best to throw a 422 error which is exactly made for this

PR #185, opened 2021-09-06
refactor: Moved azimuth angle from (-180, +180) to (0, 360)

frgfm:
The azimuth was wrongly constrained between -180 and +180, while it should have been 0 and 360.
This PR fixes that, pinned the `httpx` version which caused problems during testing, and I fixed the client unittests which were failing on a very rare edge case

### DIFF
--- a/client/test/test_client.py
+++ b/client/test/test_client.py
@@ -4,6 +4,7 @@
 # See LICENSE or go to <https://www.apache.org/licenses/LICENSE-2.0.txt> for full license details.
 
 import unittest
+import time
 from copy import deepcopy
 from requests import ConnectionError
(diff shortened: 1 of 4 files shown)


fe51 (dismissed):
Thank FG ! Okay to me

Akilditu (approved):
Thanks for the PR ! 

Seems that some tests failed but the improvement looks good to me !

mateoIdemia (approved):

MateoLostanlen:
oups wrong account :)

MateoLostanlen (approved):
Thanks for this little change that will simplify our lives :)

PR #191, opened 2022-05-20
ci: Optimized CI cache

frgfm:
This PR simply fixes the cache key for env restoration and optimized secondary keys. The result should be a speed up on PRs with multiple CI runs.

Any feedback is welcome!


fe51 (commented):
For me it's okay. Thanks for the PR

PR #215, opened 2022-09-22
fix: Updated POST request structure to retrieve token

### LINKED ISSUE #214: [client][Retrieve Token] Issues passing creds with Content Type "application/x-www-form-urlencoded"
Hi there, 

Using the api client, I have an issue when retrieving token ( in the frame of the installation of the newest test tower in Ardeche)

With correct creds (tested at https://api.pyronear.org/docs ), it basically works with some creds, and not with others !

At first, I thought it might be due to the encoding of the characters, as "application/x-www-form-urlencoded" as Content type [is not recommended to send text with non ASCII characters](https://www.microfocus.com/documentation/idol/IDOL_12_0/MediaServer/Guides/html/English/Content/Shared_Admin/_ADM_POST_requests.htm), in the request headers. 

Hence, switching from request with application/x-www-form-urlencoded to a "classic" post request with a json data, solves the issue.
However I realise that this solution is fortuitous because creds tested do not contain non-ASCII characters 🤣 

So, to be pragmatic, why using content type application/x-www-form-urlencoded in the first place to [retrieve token](https://github.com/pyronear/pyro-api/blob/master/client/pyroclient/client.py#L95) ? 
Is there any specific reason ? Otherwise, this PR #215  could be useful to avoid the problem mentioned above

happy to discuss it ! 

If anyone has an explanation as to why this doesn't work, I'm interested (text encoding then decoding causing issues?)

fe51:
Proposed solution to the issue #214

This PR introduces the following modifications in _retrieve_token method :
- Updates data to pass creds to json
- Deletes specific headers 

Closes #214


frgfm (changes requested):
Thanks Felix!

But that's the type of bug correction we cannot leave without testing. Do you think we could integrate that in the unittests? Either by changing the dummy creds to something reproducing the bug or simply checking the encoding by decoding the payload somewhere?

fe51:
You are right, it has to be integrated to unittests ! 
As this is a bug for which we don't have a theoretical explanation, I made an update of the dummy creds in the light of what the current fix allows (hence including special characters in the frame of the creds at the origin of the bug)

frgfm:
Can you confirm that those unittests fail without your modification? (if that's the case, we can move forward)

frgfm:
ping @fe51 :pray:

fe51:
@frgfm tested updated, we can merge !

fe51:
Okay, I have been able to run tests locally and solve issues !



As a reminder, the problems encountered were : 
- login must fit regular expression imposed in base.py

`  login: str = Field(..., min_length=3, max_length=50, regex="^[a-zA-Z0-9_-]+$", example="JohnDoe")
`
- Did not managed to pass a double $$ in password var to docker-compose env variable salthough doc was suggesting it would work with singles quotes ' ' ([here](https://github.com/docker/compose/issues/8607) and [there](https://deploy-preview-13474--docsdocker.netlify.app/compose/env-file/#syntax-rules))

frgfm (approved):
Thanks Felix :pray:

PR #227, opened 2022-10-16
docs: Improves README and documentation

frgfm:
- adds details in CONTRIBUTING
- revamped README (adds table description, full workflow description, usage description and configuration steps)
- passes the Sentry creds in docker compose
- added column description for all schemas
- renames `SERVER_NAME` into `SENTRY_SERVER_NAME`

Any feedback is welcome!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,7 +11,7 @@ Whatever the way you wish to contribute to the project, please respect the [code
 - [src/app](src/app) - The actual API codebase
 - [src/tests](src/tests) - The APi unit tests
 - [nginx](nginx) - NGINX configuration
-- [client](client) - The API Python client
+- [client](client) - The API Python client (please refer to its specific [contribution guidelines](client/CONTRIBUTING.md))
 
 
 ## Continuous Integration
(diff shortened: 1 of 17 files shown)


CONTRIBUTING.md
```diff
@@ -79,6 +77,12 @@ In order to run the same unit tests as the CI workflows, you can run unittests l
 make test
 ```
 
+This will run the full suite of core API unittests. However, if you're trying to run some specific unittests, you can do as follows:
```
fe51, line 80: Maybe we should specify that these tests only concern the api and not the client. The current wording suggests that the command will test everything, whereas there is a specific command for the client. Maybe specify the client test command (while specifying that the current test coverage is not complete yet) ?

README.md
```diff
@@ -32,53 +32,114 @@
...
+The API has been designed to provide, for each wildfire detection, the alert metadata:
+- timestamp
+- the picture that was used for detection
+- the location is was taken at, and the direction it was taken from
 
-If you wish to deploy this project on a server hosted remotely, you might want to be using [Docker](https://www.docker.com/) containers. Beforehand, you will need to set a few environment variables either manually or by writing an `.env` file in the root directory of this project, like in the example below:
+With the previously described tables, here are all the steps to send a wildfire alert:
+- Prerequisites (ask the instance administrator): register user
+- Register a device: declare your device on the API, using your new user credentials.
+- Create a media object & upload content: using the device credentials, save the picture metadata and upload the image content.
```
fe51, line 91: Do you think it would be needed to explain that there is service to store images (s3 alike) and other tables are in an postgres table ?

README.md
```diff
@@ -32,53 +32,114 @@
...
+![Swagger](https://github.com/pyronear/pyro-api/releases/download/v0.1.2/swagger_interface.png)
 
+In order to stop the service, run:
 ```shell
-git clone https://github.com/pyronear/pyro-api.git
+make stop
 ```
 
-## Usage
+### How is the database organized
```
fe51, line 54: Do you think an ER diagram would be needed ?

fe51 (commented):
Hi FG ! 

Thanks a lot for the PR, it makes this repo usage clearer ! few questions but not blocking :)

CONTRIBUTING.md
frgfm: I thought "core API" was specific enough :sweat_smile:

README.md
frgfm: This is explained in CONTRIBUTING (I think we should scare people only coming for usage / light development rather than "hardcore" developers)

Issue #231, opened 2022-10-20
[docs] Add a tutorial/walkthrough to deploy everything on your own


frgfm:
It would be greatly beneficial for contributors to know how to orchestrate everything on their own. I suggest making a tutorial for this

blenzi:
The README suggests that cloning the package and calling `make run` should start the API but `curl http://localhost:8080/docs` returns `Empty reply from server`. If something else is needed, it would be good to mention it in the README

frgfm:
Hi @blenzi :wave:

I just tried again myself and the curl yields:
```
    <!DOCTYPE html>
    <html>
    <head>
    <link type="text/css" rel="stylesheet" href="https://cdn.jsdelivr.net/npm/swagger-ui-dist@4/swagger-ui.css">
    <link rel="shortcut icon" href="https://fastapi.tiangolo.com/img/favicon.png">
    <title>Pyronear API - Swagger UI</title>
    </head>
    <body>
    <div id="swagger-ui">
    </div>
    <script src="https://cdn.jsdelivr.net/npm/swagger-ui-dist@4/swagger-ui-bundle.js"></script>
    <!-- `SwaggerUIBundle` is now available on the page -->
    <script>
    const ui = SwaggerUIBundle({
        url: '/openapi.json',
    "dom_id": "#swagger-ui",
"layout": "BaseLayout",
"deepLinking": true,
"showExtensions": true,
"showCommonExtensions": true,
oauth2RedirectUrl: window.location.origin + '/docs/oauth2-redirect',
    presets: [
        SwaggerUIBundle.presets.apis,
        SwaggerUIBundle.SwaggerUIStandalonePreset
        ],
    })
    </script>
    </body>
    </html>
    % 
```
You can also see that the GH CI is performing the same action and get an answer from curl, so I don't really know what caused this :shrug: 

Please open a GH issue with the docker logs to debug this :pray:

RonanMorgan:
we can closed this one, no ?

frgfm:
Correct, we're way more on a `docker compose up` instruction now than previously :+1:

Issue #217, opened 2022-10-08
[storage] Add an option to have the storage as a local container


frgfm:
To get the full process to work, there is no way currently to get around the storage solution.
We should have a local docker alternative to have everything offline

I suggest the following steps:
- [x] Standardize storage to S3-interfaced buckets (#220)
- [x] Adds a local container as S3 storage (#244)
- [x] Incorporate that so that the CI is running all bucket interactions during unittests (#244)

frgfm:
LocalStack might be a good option:
- repo : https://github.com/localstack/localstack
- aws-cli wrapper: https://github.com/localstack/awscli-local
- tuto : https://nikhilvijayan.com/aws-s3-create-bucket-copy-files-localstack

fe51:
Hi there, thanks for open the issue ! 

I have briefly experienced localstack in my previous experience (in an AWS stack context) and that could fit our need in replicating storage process ! 

How do you see the implementation ? 
- having a specific docker compose (with local stack) for local dev ? 
- whatever the solution adopted, a nice documentation to guide our dev is needed :)

frgfm:
Sorry I thought I had replied!

> having a specific docker compose (with local stack) for local dev ?

yes, that exactly (having all services being able to be run locally or outsourced to the cloud so that everything can be tested reliably)

> a nice documentation to guide our dev is needed :)

Of course, the contributing file will be significantly extended haha

PR #235, opened 2022-11-06
feat: alert notification via websocket

blenzi:
This PR adds support for notifying clients such as pyro-platform instances of new alerts via websocket. Previously the platform had to periodically ping the API to ask if new alerts were sent.


blenzi:
Websockets have many differences and subtleties with respect to HTTP routes. Without being able to test the setup locally, due to #217, #231, the development is quite hard.

The feature is implemented in a simple setup consisting of an [API](https://github.com/blenzi/test-api-websocket/) passing messages received via HTTP to websocket clients and a [dash application](https://dash-websocket.herokuapp.com/) ([code](https://github.com/blenzi/dash-websocket)) that receives those messages via websocket. Demo below.

The alternatives I see to move forward:
1. Accept the PR basically as it is with the websocket route unprotected, meaning any one can listen to the alerts (but not posting, like before). This allows implementing the feature in pyro-platform as soon as it is deployed
2. Wait for #217, #231 to be able to test the setup locally
3. Port the security implementation to the test-API above, which can be long and maybe never exactly like this API so back to 1. or 2.

Any suggestions? @frgfm ?

websocket

frgfm:
Before diving into a review, and iterating on this, I want to address something here:
- if in our stack, there is something that is critical to the whole workflow, it's the API. So we have to be quite careful about that
- now here, I understand that the websocket could help the platform reduce its ping on the API. If that's the case, we need to investigate. But then, I suggest making sure that could work, by running the API & the platform locally to make sure that works
- if that doesn't, let's avoid adding a new feature :)

frgfm:
Closing this since we've moved to the new datamodel, but we might want to reexplore this once we have better webhook reception capabilities on the frontend side. Thanks!

PR #247, opened 2023-04-11
refactor: Removes sample recording in Sentry

frgfm:
- removes the sample recording in Sentry (on top of recording errors, it was recording volume of transactions, which is hitting our quota)
- renames the API more accurately


fe51 (approved):
Hi There, thanks for the update !

PR #252, opened 2023-05-12
test: add requirements workflow

blenzi:
Add a new workflow requirements to check consistency of `poetry.lock` and `requirements.txt`


frgfm:
I found an option actually quite good for my other projects & keep those synchronized, let me open a PR :)

PR #253, opened 2023-05-28
fix: Allows admin scopes to get media URL

### LINKED ISSUE #242: [admin scope issue][media]
Hi there, 

As a user with **admin scope**, I have encountered the following issue : 

I have been denied access on a route where I am supposed to have access rights

Indeed, I get a error **403** using the get route : `/media/{media_id}/url`
and the following msg : 
```
{
  "detail": "This access can't read resources from group_id=3"
}
```

I am available to provide further information if required, 
Thanks a lot :)

frgfm:
This PR allows admin scopes to bypass the group verification to retrieve the URL of a media.

Closes #242


frgfm:
`check_group_read` is actually already doing that, the problem came from #264

PR #264, opened 2023-06-20
fix: Fixes access ID resolution on media URL resolution

frgfm:
- fixes the access check on media URL resolution. Interesting fact about why this wasn't spotted earlier: we were using the user ID as access ID, which is not a problem when you create all the users before any device (because then access ID = user ID). Hence the problem @MateoLostanlen (I got the idea because I didn't encounter the problem when using the superadmin which is the first access that gets created, but had the problem with other admins, where the user ID != access ID)
- avoids DB direct exposition for external HTTP requests

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -30,8 +30,8 @@ services:
     image: postgres:12.1-alpine
     volumes:
       - postgres_data:/var/lib/postgresql/data/
-    ports:
-      - 5432:5432
+    expose:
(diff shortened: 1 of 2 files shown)


frgfm:
Just checked by deploying on the dev server, it fixes the problem :) 
So I'm merging & deploying the fix!

MateoLostanlen:
thanks a lot @frgfm

PR #244, opened 2023-03-15
feat: Improve dev experience by using local S3 

Acruve15:
## Description

Linked to [Issue 217](https://github.com/pyronear/pyro-api/issues/217). Developing locally is not possible if the developer does not have an AWS S3 access. 

## What was done

We used [localstack](https://github.com/localstack) to solve this issue.

PR #267, opened 2023-07-13
feat: Adds localization column in the alerts table

frgfm:
- adds a `localization` field in the alerts table. Meant to be used as a JSON dump of a list of bounding boxes
- updates tests
- updates client
- adds alembic migration

Here is the new database UML:
uml

cc @MateoLostanlen


MateoLostanlen:
Thanks a lot for this PR @frgfm !

frgfm:
To be honest, it's a mystery to me why the CI fails: locally I can build, run the test and the E2E script
And I tried on the dev server, everything runs smoothly :shrug:

fe51:
Hi there, is it the same CI issue as in #244  ? (to find out if it would be worth passing this PR too :) )

fe51:
Also, thanks a lot for the PR :)

blenzi:
Very nice addition! But unfortunately the problem here is linked to the db service. It works locally (after deleting the volume) but not in the CI :(

On 31 Jul 2023, at 15:02, fe51 ***@***.******@***.***>> wrote:



Hi there, is it the same CI issue as in #244<https://github.com/pyronear/pyro-api/pull/244> ? (to find out if it would be worth passing this PR too :) )

blenzi (changes requested):
Thanks for this addition @frgfm ! We initially thought this was urgent but in the end I propose we take the time to fix the issues with the CI and we deploy a data model consistent with all the other tables and classes:
- introduce a class Bbox, ROI / RegionOfInterest or equivalent with `x0, y0, x1, y1, confidence`, and a list of those to `AlertIn`, instead of strings
- add a table with the info above and `alert_id`, to allow multiple bboxes for each alert

I also managed to run the API and the tests locally. My attempts with the CI did not work either so we can try to apply the various changes one by one in another PR:
- python 3.8 -> 3.9
- pydantic 1.10 -> 2
- code-quality fixes 
- postgres 12 -> 15
- changes in docker-compose, including health-checks

blenzi:
@frgfm : the minimal changes for this to work are already in #289

frgfm:
> @frgfm : the minimal changes for this to work are already in #289

Yup I saw, but it was missing the alembic migration & I took care of the rest sequentially to handle all PR in order :sweat_smile:

PR #268, opened 2023-07-27
feat: add route to retrieve alerts for a given event

blenzi:
Currently there is a problem for the platform: the list of ongoing alerts is limited to 50 and can fail to retrieve the alerts of an unacknowledged event. This PR adds route `/events/{event_id}/alerts` and the corresponding method `Client.get_alerts_for_event` to retrieve the list of alerts for a given event.


src/app/api/endpoints/alerts.py
```diff
@@ -146,3 +147,17 @@ async def fetch_ongoing_alerts(
         )
         retrieved_alerts = [x.__dict__ for x in retrieved_alerts.all()]
         return retrieved_alerts
+
+
+@router.get("/event-alerts/{event_id}/", response_model=List[AlertOut], summary="Get the list of alerts for event")
```
frgfm, line 152: Just a suggestion:
- with this, the route will be `alerts/event-alerts/event_id`
- common public APIs would go more for the `events/event_id/alerts`

The rest is all good :+1:

frgfm (dismissed):
Thanks for the PR! I added a suggestion about a potential better route to place that feature in. We're just missing the client method to handle this but the rest looks alright :+1:

client/pyroclient/client.py
```diff
@@ -230,6 +231,20 @@ def get_ongoing_alerts(self) -> Response:
         """
         return requests.get(self.routes["get-ongoing-alerts"], headers=self.headers, timeout=self.timeout)
 
+    def get_alerts_for_event(self, event_id: int) -> Response:
+        """Get all the alerts in the DB for the given event
+
+        >>> from pyroclient import client
+        >>> api_client = client.Client("http://pyronear-api.herokuapp.com", "MY_LOGIN", "MY_PWD")
```
Akilditu, line 238: client api is still on Heroku ?

Akilditu (approved):
Hi Bruno Thanks for this PR, everything seems fine to me !

PR #269, opened 2023-07-28
refactor: change alert relaxation from 5 to 30 minutes

blenzi:
When a device issues multiple alerts, it is associated to an existing event if it is within "relaxation time". This PR increases the relaxation time from 5 to 30 minutes to avoid creating multiple events for the same fire detection. The risk to miss another fire within the relaxation time is considered small at this point.


MateoLostanlen (approved):
Thank you for this PR, as discussed it seems much more suitable.

Akilditu (approved):
Thanks for the PR, indeed more relevant and will slow new event creation

PR #271, opened 2023-08-02
feat: backup db content locally and copy sql file to s3 bucket

blenzi:
This PR adds a script and adapts the code to be able to backup db content and upload it to s3 bucket

- add scripts/dump_db.sh
- s3.py S3Bucket.upload_file: support for ascii files
- add python-dotenv dependency and load .env from root directory to be able to use env variables outside docker compose
- docker-compose.yml: add volume db_backup to frontend service

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,11 +112,14 @@ This file will have to hold the following information:
 - `S3_REGION`: your S3 bucket is geographically identified by its location's region
 - `S3_ENDPOINT_URL`: the URL providing a S3 endpoint by your cloud provider
 - `BUCKET_NAME`: the name of the storage bucket
+- `DATABASE_URL`: database URL, including username and password 
 
 Optionally, the following information can be added:
 - `SENTRY_DSN`: the URL of the [Sentry](https://sentry.io/) project, which monitors back-end errors and report them back.
 - `SENTRY_SERVER_NAME`: the server tag to apply to events.
 - `BUCKET_MEDIA_FOLDER`: the optional subfolder to put the media files in
+- `BUCKET_DB_BACKUP_FOLDER`: subfolder where to store db backups
+- `POSTGRES_USER`: db username ; required for backups
 
 So your `.env` file should look like something similar to:
 ```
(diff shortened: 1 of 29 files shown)


README.md
```diff
@@ -125,6 +128,8 @@ S3_SECRET_KEY=YOUR_SECRET_KEY
 S3_REGION=bucket-region
 S3_ENDPOINT_URL='https://s3.mydomain.com/'
 BUCKET_NAME=my_storage_bucket_name
+POSTGRES_USER=my_postgres_user
+DATABASE_URL=postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@db/${POSTGRES_DB}
```
fe51, line 132: POSTGRES_PASSWORD and POSTGRES_DB are not in .env ?

fe51 (dismissed):
Hi Bruno, thanks a lot for this PR that bring an interesting feature. 

I have made a small comment

frgfm:
Great idea, we definitely need some backup!
One suggestion though: to avoid mixing DB backup & images in the same bucket, it might be worth setting a simple cron job instead?

What I have in mind would be the CRON job doing in order:
- dumping the sql DB from outside the postgres docker into a data/binded volume
- using the backend docker to upload the dump from the volume (similar to the design here)
That way we don't need to add any dependencies on root (otherwise, if we're OK with that, we can do everything without interacting with the backend container).

What do you think?

frgfm:
Closing this as this was built on the v1, but giving it additional thoughts:
- we can't have a full-featured and easily maintanable/orchestrable DB backup system. So here for instance, we'd be storing backups that we usually restore manually which raise the question whether we'd use it actually
- if we are in need of backup, I'd suggest just having the DB remote on Supabase for instance (they handle backup, etc.)
- if we want something minimal, I think we should only focus on restoring critical tables: organizations, users, cameras which handle access. The rest is basically historical data

Ideally, we'd want a safeguard to do:
```
make backup-access
# pull/upgrade API version, etc
make restore-access
```

Issue #263, opened 2023-06-05
How webhooks work


psd24:
What is the operation of Webhooks?

In what phase of the application are they sent?

I would like to be able to create a bot for Telegram and my idea is to collect information from a Webhook when there is an alert from the application.

If there is more information about this I would like to know it.

Thank you,

fe51:
@frgfm @blenzi :)

frgfm:
Hey there :wave: 

Depends what you mean by that precisely, but the "webhook" table is meant to add post-route actions freely. Only admins can create some and manage the subscriptions.

For your Telegram bot:
- if a HTTP post request can send the message, that can easily be added to the API without adding dependencies
- if you need specific dependencies, I would suggest creating another service routing alert notifications to the right channel

frgfm:
Closing this as this seems to have answered your question :)
Feel free to reopen, but it would be more suited as a discussion (not a bug or feature request)

PR #272, opened 2023-08-03
feat: send notifications when an alert is received 

blenzi:
With this PR, a notification message is sent when the first alert in each event is created. The notification is sent via telegram (or another type, to be implemented if needed) to the recipients subscribed and associated to a group, and logged in the db.

- add routes and tables for recipients and notifications


src/app/api/crud/alerts.py
```diff
@@ -30,14 +30,19 @@ async def resolve_previous_alert(device_id: int) -> Optional[AlertOut]:
...
 
 
-async def create_event_if_inexistant(payload: AlertIn) -> int:
+async def create_event_if_inexistant(payload: AlertIn) -> Tuple[Optional[int], bool]:
+    """Return the id of the event to be associated with the alert and a boolean flag to tell if a new event was created
+
+    Args:
+        payload: alert object
+
+    Returns: tuple (bool, int) -> (new alert ?, event_id)
```
fe51, line 39: it returns tuple(int, bool) instead of tuple (bool, int) right ?

src/app/services/utils.py
```diff
@@ -16,3 +22,28 @@ def resolve_bucket_key(file_name: str, bucket_folder: Optional[str] = None) -> s
...
     return f"{bucket_folder}/{file_name}" if isinstance(bucket_folder, str) else file_name
+
+
+def send_telegram_msg(chat_id: str, message: str, autodelete: bool = False) -> Optional[Response]:
+    """Send telegram message to the chat with the given id
+
+    Args:
+        chat_id (str): chat id
+        message (str): message to send
+        autodelete (str): delete msg after sending. Used for tests
```
fe51, line 33: autodelete will be delete once the feature has been fully tested?

fe51 (commented):
Hello Bruno, Thanks a lot for this such useful feature ! 

Just a few questions and remarks before approving !

src/app/api/crud/alerts.py
blenzi: Indeed! Fixed

src/app/services/utils.py
blenzi: It is actually a bool, not a string. This is used in unittests, that send a real msg (without notification) and delete it right after, not to pile-up. I tried to improve the docstring:

`autodelete (bool, default=False): delete msg after sending. Used by unittests`

blenzi:
Once it is deployed, we can close #263

fe51 (approved):

frgfm:
Sorry I wasn't available to review here!
FYI, this is typically something we designed webhooks for :sweat_smile: 

Meaning the webhooks would be a content (Telegram msg + recipient) and a destination (Telegram API). And in that regard, I would argue we don't need the notifications table (we could get the timestamp of an alert or add a "notification_sent_at" column) and I think the recipient table could be put in the webhook if we do this correctly.

Not a problem, but perhaps it would be helpful to consider that for the tables. The more we have, the bigger the DB, the slower the API & the tests, and the more we have to maintained :upside_down_face: 

Happy to discuss this a bit later!

PR #277, opened 2023-08-07
feat: add limit and offset parameters to routes that fetch lists from db

### LINKED ISSUE #243: [Ongoing Alerts] Specify the limit size of the response in API documentation
Hi there, 

Using the route alerts, I realize the max response size for /alerts/ongoing is limited to 50

If it can be guessed from the [code](https://github.com/pyronear/pyro-api/blob/eb4fcc1d60ae2d9ac8f3aac372fb974112521337/src/app/api/endpoints/alerts.py#L139
), it would be useful for the user to have this information directly in the documentation. 

I don't know if this information is specific to this route, if not, it would be good documentation practice in the general case. 

Thanks a lot, 
Happy to discuss it !

### LINKED ISSUE #250: Allow to remove max request response size
#190 introduce a response size limitation of 50, I get the point of doing this but can we add an optional prams to drop this limitation if we need it ?

### LINKED ISSUE #276: Fetch routes do not limit number of items for non-admin users
The routes that fetch objects from the db behave differently for admin and non-admin users. For admin, they are limited to 50 items while for non-admin users all the items are returned. For example: https://github.com/pyronear/pyro-api/blob/main/src/app/api/endpoints/alerts.py#L135

blenzi:
This PR adds limit and offset parameters with descriptions, to all routes that fetch lists from db (fix #243, fix #250). Also:

- use crud.fetch_all in all the routes above, to limit the number of items returned for non-admin users (fix #276)
- forbid /installation/site-devices to access site outside of group, instead of filtering devices
- improve test_installations.py (test_get_active_devices_on_site)
- Dockerfile-dev: update FROM image name

### DIFF
--- a/src/app/api/crud/base.py
+++ b/src/app/api/crud/base.py
@@ -8,6 +8,8 @@ from typing import Any, Dict, List, Mapping, Optional
 from fastapi import HTTPException, Path, status
 from pydantic import BaseModel
 from sqlalchemy import Table
+from sqlalchemy.orm import Query
+from sqlalchemy.sql import Select
 
 from app.db import database
 
@@ -41,8 +43,11 @@ async def fetch_all(
     query_filters: Optional[Dict[str, Any]] = None,
     exclusions: Optional[Dict[str, Any]] = None,
     limit: int = 50,
+    offset: Optional[int] = None,
+    query: Optional[Select] = None,
 ) -> List[Mapping[str, Any]]:
-    query = table.select().order_by(table.c.id.desc())
+    if query is None:
+        query = table.select()
     if isinstance(query_filters, dict):
         for key, value in query_filters.items():
             query = query.where(getattr(table.c, key) == value)
@@ -50,7 +55,10 @@ async def fetch_all(
     if isinstance(exclusions, dict):
         for key, value in exclusions.items():
             query = query.where(getattr(table.c, key) != value)
-    return (await database.fetch_all(query=query.limit(limit)))[::-1]
+    query = query.order_by(table.c.id.desc()).limit(limit).offset(offset)
+    if isinstance(query, Query):
+        return [item.__dict__ for item in query[::-1]]
+    return (await database.fetch_all(query=query))[::-1]
 
 
 async def fetch_one(table: Table, query_filters: Dict[str, Any]) -> Optional[Mapping[str, Any]]:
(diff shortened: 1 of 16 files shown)


src/app/api/crud/base.py
```diff
@@ -41,16 +43,22 @@ async def fetch_all(
...
             query = query.where(getattr(table.c, key) == value)
 
     if isinstance(exclusions, dict):
         for key, value in exclusions.items():
             query = query.where(getattr(table.c, key) != value)
-    return (await database.fetch_all(query=query.limit(limit)))[::-1]
+    query = query.order_by(table.c.id.desc()).limit(limit).offset(offset)
+    if isinstance(query, Query):
+        return [item.__dict__ for item in query[::-1]]
+    return (await database.fetch_all(query=query))[::-1]
```
frgfm, line 61: Good idea. Two questions though:
- I think I understand this is an equivalent of the offset syntax in SQL. But here we seem to be making an arbitrary choice on order direction. Should we make that a boolean param?
- regarding the query arg, we expect it to already have a `.select()`?

If we only want to paginate the queries, there are a few plugin options for this I think

src/app/api/endpoints/accesses.py
```diff
@@ -25,8 +26,12 @@ async def get_access(access_id: int = Path(..., gt=0), _=Security(get_current_ac
 
 
 @router.get("/", response_model=List[AccessRead], summary="Get the list of all accesses")
-async def fetch_accesses(_=Security(get_current_access, scopes=[AccessType.admin])):
+async def fetch_accesses(
+    limit: Annotated[int, Query(description="maximum number of items", ge=1, le=1000)] = 50,
+    offset: Annotated[Optional[int], Query(description="number of items to skip", ge=0)] = None,
+    _=Security(get_current_access, scopes=[AccessType.admin]),
+):
```
frgfm, line 33: Just a reminder for me: how do we change those values when sending a request?

src/app/api/crud/base.py
frgfm: (paginate will prevent the case where we have a huge table and someone send `limit=100000` which might blow up the RAM)

frgfm (commented):
Thanks for the PR! I added a few comments, but the overall one is: do we want to paginate query results or allow users to specify limits? (in the last case, if a large value is selected, that can blow up the RAM. even though we might not have a lot of malicious users around)

I saw something a few months ago to paginate easily in fastapi: https://uriyyo-fastapi-pagination.netlify.app/

What do you think?

frgfm:

Issue #274, opened 2023-08-05
Clarify when (and how) users take events into account


fe51:
Hi there ✋ 

## Feature 

Know when (by whom, and how) an alert/event was seen

## Motivation

As of today, knowledge of how events are taken into account by users is limited to `is_acknowledged` ,(a boolean) [information](https://github.com/pyronear/pyro-api/blob/c10ff1cf8eb37456196991e5b7da457d908ab9bb/src/app/models/event.py#L30C11-L30C11). 

The problem is that it doesn't contain any information about when (I think, tell if I am wrong) the information was seen **and** moreover it means: "I have seen this alert/event" AND "I no longer wish to see it on the platform". 

However, having information about when the information was considered is important if we want to be able to estimate the time saved for firefighter (compared, for example, with an alert received via an emergency call).

### Update in pyro-api
A solution could be to add new fields to event table;

| Field | Type |
| ----------- | ----------- |
| viewed_at | datetime |
| viewed_by | user_id |
| viewed_with | UI_type (plateform, telegram, mail, etc) |

However : The suggestion here maybe poses a problem if several users see the information (only insert if field are empty to store the earlier info) (Maybe, as admin users, no info should be sent when seeing an alert ? so update those field only if the user_id is associated to the group_id of the owner_id of the device that have sent the alert ? 

- Pro: 
   - Feature easy to implement
   - Long-term storage of this information, and ease of implementation of associated dashboarding (dashboard building stats relating to our performance, requesting only the api).

It will be useful for the feedback we'll be doing on the summer experiments ! 

- Cons : too speficic to be stored in the api  ?

### Web Analytics App
Those information could be obtained setting up a web analytics app (like [matomo](https://fr.matomo.org/
), similar to google analytics but respecting privacy)

This will require a benchmark of the web analytics solution we want to use. Implementing it, and I tend to think that although it's necessary to implement one in general, for this specific use case, having the information in the API makes it easier to get the stats / performance of our solution. Moreover (I do not know it we would be able to monitor with ease those info if it cames from telegram or mail)

What do you think ? 
Happy to discuss it ! 😄  
This is a fairly important feature if we want to be able to get feedback from our experiments this summer ! 

Of course, there will be further developments to come on the platform side. We'll open dedicated issues once the back-end is clear :)

fe51:
@blenzi @frgfm 

@MateoLostanlen @GHCamille

GHCamille:
Hello, 

[...]

PR #278, opened 2023-08-07
feat: add fields to store when an event was acknowledged and by whom

blenzi:
To address #274 , this PR adds acknowledged_ts and acknowledged_by to events table

### DIFF
--- a/src/app/models/event.py
+++ b/src/app/models/event.py
@@ -5,7 +5,7 @@
 
 import enum
 
-from sqlalchemy import Boolean, Column, DateTime, Enum, Float, Integer
+from sqlalchemy import Boolean, Column, DateTime, Enum, Float, ForeignKey, Integer
 from sqlalchemy.orm import RelationshipProperty, relationship
 from sqlalchemy.sql import func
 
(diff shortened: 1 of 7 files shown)


fe51 (dismissed):
Hi Bruno, thanks a lot for the changes addressing #274 ! Looks good to me !

blenzi:
Thanks @fe51 ! There was a problem that caused the client tests to fail: I was recording the access_id instead of the user_id in events/acknowledge_by. This is fixed now.

fe51:
Thanks a lot @blenzi

fe51 (approved):

src/app/models/event.py
```diff
@@ -28,9 +28,12 @@ class Event(Base):
     start_ts = Column(DateTime, default=func.now())
     end_ts = Column(DateTime, default=None, nullable=True)
     is_acknowledged = Column(Boolean, default=False)
+    acknowledged_by = Column(Integer, ForeignKey("users.id"))
+    acknowledged_ts = Column(DateTime, default=None, nullable=True)
```
frgfm, line 32: to follow the other named fields, perhaps a "acknowledged_at"?

src/app/models/user.py
```diff
@@ -23,6 +23,7 @@ class User(Base):
 
     access: RelationshipProperty = relationship("Access", uselist=False, back_populates="user")
     device: RelationshipProperty = relationship("Device", uselist=False, back_populates="owner")
+    acknowledged_events: RelationshipProperty = relationship("Event", back_populates="acknowledger")
```
frgfm, line 26: Mmmh, that's a tricky column to maintain I think?
We could get that information by querying the event table with the user.id. Especially for French regions were everyone uses the same login, that field value could become huge otherwise

frgfm (commented):
Thanks! I only have a doubt about adding the acknowledged_events list in the user table. If that's information needed by an external service, we can query the event table by filtering on the acknowledged_by column, don't you think?

frgfm:

PR #279, opened 2023-08-11
feat: send alert image with telegram message, if requested 

blenzi:
This PR adds the possibility to send the alert image together with a telegram message, if requested

- add boolean field send_image to recipient table
- add int field media_id to notification table
- add and use python-telegram-bot dependency (negligible increase in docker image size)


src/Dockerfile-dev
```diff
@@ -1,4 +1,4 @@
-FROM pyroapi:python3.8-alpine3.10
+FROM pyronear/pyro-api:python3.8-alpine3.10
```
MateoLostanlen, line 1: @blenzi your are changing this in severals PR it should be done once in a PR I guess

blenzi: Thanks @MateoLostanlen. Now that #281 was merged, this will be part of main

frgfm:

Issue #280, opened 2023-08-25
Grouping alerts 


MateoLostanlen:
Hi there,

I'm trying to understand why the grouping of alerts doesn't work. Maybe I'm misunderstanding but it seems to me that we should put a symbol <= [here](https://github.com/pyronear/pyro-api/blob/fa2624d48b7c4de638ea87a589b6f118960ec7f7/src/app/api/crud/alerts.py#L23) to have the alerts created within the last 30 minutes, right?

@frgfm @blenzi what do you think ?

blenzi:
Hi @MateoLostanlen , I think `>=` is correct since `max_ts = now - 30 minutes` and this is querying previous alerts. A query with `<=` would return all previous alerts (not within the last 30 minutes).

blenzi:
In the end the mechanism is working but the deployed version had the old value of 5 minutes. Closing

PR #281, opened 2023-08-25
test: explicitly test alert grouping (#280)

blenzi:
- test alert grouping within ALERT_RELAXATION_SECONDS (+ tolerance)
- fix image path in Dockerfile-dev


MateoLostanlen (approved):
Thanks for increasing test coverage :)

PR #287, opened 2023-09-21
Add support for a s3 bucket proxy

Nastaliss:
Add a proxy configuration for the s3 bucket that redirects requests from the proxy's url to the bucket's url. 
This is used to hide the bucket's real url, that is sometimes flagged as "harmful" by some companies' network policies.

The proxy can be toggled on and off with `S3_USE_PROXY`. By default the urls served from the [get_media_url route](https://github.com/pyronear/pyro-api/blob/main/src/app/api/endpoints/media.py#L175) are the ones generated by Boto3 directly. 
If the proxy is enabled it will replace the bucket's url by the `S3_PROXY_URL`. Keeping the media details as well as auth query params. It basically only changes the hostname.


.github/workflows/builds.yml
```diff
@@ -30,6 +30,8 @@ jobs:
           S3_ACCESS_KEY: ${{ secrets.S3_ACCESS_KEY }}
           S3_SECRET_KEY: ${{ secrets.S3_SECRET_KEY }}
           S3_REGION: ${{ secrets.S3_REGION }}
+          S3_USE_PROXY: ${{ secrets.S3_USE_PROXY }}
+          S3_PROXY_URL: ${{ secrets.S3_PROXY_URL }}
```
frgfm, line 34: Can't the S3_ENDPOINT_URL be used instead? I'd be easier to minimize the number of config parmeters

frgfm: Actually, even removing the S3_USE_PROXY, just passing the endpoint url. independently on whether it's the proxy or not

src/app/services/bucket/s3.py
```diff
@@ -24,12 +24,16 @@ class S3Bucket:
...
+        proxy_url: the proxy url
     """
 
-    def __init__(self, region: str, endpoint_url: str, access_key: str, secret_key: str, bucket_name: str) -> None:
+    def __init__(self, region: str, endpoint_url: str, access_key: str, secret_key: str, bucket_name: str, use_proxy: bool, proxy_url: str) -> None:
         _session = boto3.Session(access_key, secret_key, region_name=region)
         self._s3 = _session.client("s3", endpoint_url=endpoint_url)
         self.bucket_name = bucket_name
+        self.use_proxy = use_proxy
+        self.proxy_url = proxy_url
```
frgfm, line 36: Can't we pass the value of proxy_url into "endpoint_url" to boto3 client?

frgfm (commented):
Thanks! Only have two questions:
- cf. my comments but perhaps we can reduce the number of config params?
- the proxy should only be used in the test env shouldn't it? If so, changes should only appear in the dev docker compose and we'd need the local S3 service as an accessible container?

We're almost there, thanks for all your work :)

Nastaliss:
@frgfm I used a separate url for the proxy to only use it when serving medias to the users, as the proxy is only useful in that specific case. 
Actually as the change was initiated for engie I don´t think it would only be used in dev. 
Is there a separate s3 endpoint that is used for dev ?

src/app/config.py
```diff
@@ -47,6 +47,12 @@
 S3_SECRET_KEY: str = os.environ["S3_SECRET_KEY"]
 S3_REGION: str = os.environ["S3_REGION"]
 S3_ENDPOINT_URL: str = os.environ["S3_ENDPOINT_URL"]
+S3_USE_PROXY: bool = os.environ.get("S3_USE_PROXY", "").lower() != "false"
+S3_PROXY_URL: str
+if S3_USE_PROXY:
+    S3_PROXY_URL = os.environ["S3_PROXY_URL"]
+else:
+    S3_PROXY_URL = ""
```
frgfm, line 55: Let's use a one line condition:
```suggestion
S3_PROXY_URL: Union[str, None] = os.environ["S3_PROXY_URL"] if S3_USE_PROXY else None
```

src/app/services/bucket/s3.py
frgfm, line 36: Those two should be optional if I get the purpose of this (or default to false for use_proxy, and optional for proxy_url)

frgfm (approved):
Thanks for the PR!

src/app/config.py
```diff
@@ -47,6 +47,8 @@
 S3_SECRET_KEY: str = os.environ["S3_SECRET_KEY"]
 S3_REGION: str = os.environ["S3_REGION"]
 S3_ENDPOINT_URL: str = os.environ["S3_ENDPOINT_URL"]
+S3_USE_PROXY: bool = os.environ.get("S3_USE_PROXY", "").lower() != "false"
```
frgfm, line 50:
```suggestion
S3_USE_PROXY: bool = os.environ.get("S3_USE_PROXY", "false").lower() != "false"
```

frgfm (changes requested):
I think I found the problem!

frgfm (approved):
Looks all good! Probably a permission issue from the PR since it comes from a fork (didn't know it was an issue though)

Nastaliss:
[issue](https://github.com/pyronear/pyro-api/issues/300)

PR #289, opened 2023-10-03
feat: Adds localization column in the alerts table

blenzi:
Following (#267), this PR adds a localization field in the alerts table, including minimal necessary changes.
The localization field is meant to be used as a JSON dump of a list of bounding boxes.


MateoLostanlen (commented):
Hi, thanks for the PR, it might be nice to have a test with a different localization than None.

Makefile
```diff
@@ -17,7 +17,7 @@ lock:
 # Build the docker
 build:
 	poetry export -f requirements.txt --without-hashes --output src/app/requirements.txt
-	docker build src/. -t pyronear/pyro-api:python3.8-alpine3.10
+	docker build src/. -t pyronear/pyro-api:python3.8-alpine3.10-prod
```
MateoLostanlen, line 20: I might not be the right PR to change this

blenzi: Thanks for spotting. I copied it from the previous one. Removed now

blenzi:
Thanks @MateoLostanlen . There is a value that is not None but it was not according to the desired format (`conf` missing). I added some validation checks and fixed it

frgfm:
Closing this since it has been merged in #267

PR #291, opened 2023-10-09
feat: fields and methods to set event type (wildfire or something else)

blenzi:
This PR adds the possibility to set an event type, different than wildfire. This event assessment is currently done via `pyro-platform` but this information is only stored locally (in the machine running it).

- add options to EventType enum: domestic fire, chimney, cloud, other, undefined (default)
- add type_set_by and type_set_ts to events table
- add PUT /events/{event_id}/type endpoint and `set_event_type` method to client

### DIFF
--- a/client/pyroclient/client.py
+++ b/client/pyroclient/client.py
@@ -37,6 +37,7 @@ ROUTES: Dict[str, str] = {
     "get-unacknowledged-events": "/events/unacknowledged",
     "get-past-events": "/events/past",
     "acknowledge-event": "/events/{event_id}/acknowledge",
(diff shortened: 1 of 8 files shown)


frgfm:
Thanks for the PR!
I'm aligned with everything apart from the "set_by" and "set_ts". I think we're mixing the role of APM (being able to access logs / interactions) vs. what needs to be recorded in the DB. My suggestion:
- simply removing the set columns
- I can come up with a PR to add a cool APM I've been using (posthog) where we can explore events like those

Here is a screenshot once it's set:
image

frgfm:

PR #294, opened 2023-10-17
fix: Adds missing reference to the endpoint_url for the S3 proxy

Nastaliss:
The endpoint_url reference in the method to generate file download URL is missing. This PR fixes that

### DIFF
--- a/src/app/services/bucket/s3.py
+++ b/src/app/services/bucket/s3.py
@@ -31,7 +31,6 @@ class S3Bucket:
         self, region: str, endpoint_url: str, access_key: str, secret_key: str, bucket_name: str, proxy_url: str
     ) -> None:
         _session = boto3.Session(access_key, secret_key, region_name=region)
-        self.endpoint_url = endpoint_url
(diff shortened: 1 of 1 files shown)


frgfm:
@Nastaliss Would you mind adding a PR description and creating an issue referencing all the PRs on that topic please? :pray:

src/app/services/bucket/s3.py
```diff
@@ -31,6 +31,7 @@ def __init__(
         self, region: str, endpoint_url: str, access_key: str, secret_key: str, bucket_name: str, proxy_url: str
     ) -> None:
         _session = boto3.Session(access_key, secret_key, region_name=region)
+        self.endpoint_url = endpoint_url
```
frgfm, line 34: No need we have 
```python
self._s3.meta.endpoint_url
```

frgfm (changes requested):
No need for an extra instance attribute, I'll edit the PR :+1:

Nastaliss:
[issue](https://github.com/pyronear/pyro-api/issues/300)

Issue #304, opened 2023-10-19
Clarifying data/state tables from event tables


frgfm:
I've been thinking about the relevance of some of the early tables we created. I identify the core purpose of the backend as:
- identifying start & end of wildfires events
- provide live monitoring during the wildfire (access latest images/prediction)

Now think about the tables & columns involved in each stage:
- each device needs to be registered in the API (access or device table) with each location & orientation (potentially it's site)
- when a device detects a wildfire (locally a buffer/confirmation mechanism is possible), using its authentication, it needs to send the image. To simplify, I'd argue that if the device can't send the image, it can be used for training/data collection later, but it's no use for firefighters.
- on the backend, the image is uploaded to S3 & generates a URL for it. It creates an event at the location of the device, using the current timestamp
- for monitoring purposes, in the device table, each device should have last_ping, and last_image (url to S3)

Now since we have a lot of noise between events & alerts, I suggest considering the following table with their columns:
- **groups, accesses, users & sites (rename stations?)**: same as before
- **devices**: id, access_id, login, group_id, current_site_id, last_ping, last_image_bucket_key, lat+lon+elevation+azimuth+angle of view
- **alerts**: id, device_id, image_bucket_key, detection_localization, created_at (the device gives lat+lon+azimuth & cie)
Finally, I think we need to rework the event table and simplify it considering that it only aggregates the alerts from a single device (i.e. if we have 4 devices on a site, we can have 4 events at the same time):
- **events**: id, device_id, type, first_alert, last_alert, created_at (the last_alert gets updated by detections, and type by firefighters)

(we remove the legacy media, installations & notifications tables. and I think we should merge webhooks & recipients)

In order to see more clearly, I can set an APM like PostHog to be able to explore events. That should help us figure out which tables are still relevant afterward.

What do you think?

blenzi:
Thanks for initiating the discussion. Since you are considering a large refactoring of the db, here are some thoughts:

I have the impression that some definitions are historical and not used/needed like ongoing alert.

It is not clear to me what can we achieve with PostHog but I think we are desperately lacking ways to monitor and investigate the system behavior, response and usage. Which devices were active at a given time? How many events / alerts were triggered by period / site / device ? What are the associated images ? Who saw / received a notification or at least acknowledged an event / alert and how long did it take ? Was it a real fire?

[...]

PR #307, opened 2023-11-01
feat: Adds product analytics on most of the routes

frgfm:
- simplifies the storage interactions (bucket/s3 --> storage.py)
- adds a telemetry client
- adds event capture on most routes

Any feedback is welcome!

cf. #304

### DIFF
--- a/src/app/api/endpoints/alerts.py
+++ b/src/app/api/endpoints/alerts.py
@@ -13,14 +13,14 @@ from sqlalchemy import select
 from app.api import crud
 from app.api.crud.authorizations import check_group_read, is_admin_access
 from app.api.crud.groups import get_entity_group_id
-from app.api.deps import get_current_access, get_current_device, get_db
+from app.api.deps import get_current_access, get_current_device, get_current_user, get_db
 from app.api.endpoints.devices import get_device
 from app.api.endpoints.notifications import send_notification
 from app.api.endpoints.recipients import fetch_recipients_for_group
 from app.api.external import post_request
 from app.db import alerts, events, media
 from app.models import Access, AccessType, Alert, Device, Event
-from app.schemas import AlertBase, AlertIn, AlertOut, DeviceOut, NotificationIn, RecipientOut
+from app.schemas import AlertBase, AlertIn, AlertOut, DeviceOut, NotificationIn, RecipientOut, UserRead
 from app.services.telemetry import telemetry_client
 
 router = APIRouter()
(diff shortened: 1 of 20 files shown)


blenzi:
Thanks for this nice addition @frgfm ! I believe the errors in the CI come from 
`from app.services import telemetry_client` (not defined) vs from app.services.telemetry import telemetry_client`.

I would suggest monitoring also the alert notifications, in order to remove that from the db, as you suggested in #304.

src/app/api/endpoints/alerts.py
```diff
@@ -81,23 +70,39 @@ async def create_alert(
...
+    background_tasks: BackgroundTasks,
+    user: Annotated[UserRead, Security(get_current_user, scopes=[AccessType.admin])],
+) -> AlertOut:
+    """
+    Creates a new alert based on the given information and send a notification if it is the first alert of the event
+
+    Below, click on "Schema" for more detailed information about arguments
+    or "Example Value" to get a concrete idea of arguments
+    """
+    telemetry_client.capture(user.id, event="alerts-create")
```
blenzi, line 85: Add device / site / group name ?

src/app/api/endpoints/alerts.py
```diff
@@ -81,23 +70,39 @@ async def create_alert(
...
     device: DeviceOut = Security(get_current_device, scopes=[AccessType.device]),
-):
+) -> AlertOut:
     """
     Creates an alert related to the authentified device, uses its device_id as argument
 
     Below, click on "Schema" for more detailed information about arguments
     or "Example Value" to get a concrete idea of arguments
     """
+    telemetry_client.capture(device.id, event="alerts-create-from-device", properties={"owner_id": device.owner_id})
```
blenzi, line 105: Add device, site and group name instead ?

src/app/api/endpoints/devices.py
```diff
@@ -26,17 +26,19 @@
...
 
 @router.post("/", response_model=DeviceOut, status_code=status.HTTP_201_CREATED, summary="Create a new device")
-async def register_device(payload: AdminDeviceAuth, _=Security(get_current_access, scopes=[AccessType.admin])):
+async def register_device(payload: AdminDeviceAuth, access=Security(get_current_access, scopes=[AccessType.admin])):
     """Creates a new device based on the given information
 
     Below, click on "Schema" for more detailed information about arguments
     or "Example Value" to get a concrete idea of arguments
     """
+    telemetry_client.capture(access.id, event="devices-create")
```
blenzi, line 41: Add device / site / group name ?

blenzi (commented):
Thanks a lot for these additions, @frgfm ! 

I have no experience with PostHog but if I understand well, it will not have access to our db. Therefore, to maximize the possibilities it offers, we probably need some more detailed, human-readable information about the events. This will be required when analyzing individual alerts, statistics per site / group, etc. So I added in a few places the suggestion to include device / site / group name, updated information, etc. What do you think ?

frgfm:
So apart from "updated info" (feels like dumping the payload otherwise), I added everything that was immediately available (not requiring another DB operations)

blenzi (approved):
Thanks @frgfm ! Let's start using it asap and adjust the level of information if needed

PR #311, opened 2024-01-24
docs: Updates copyright year

frgfm:
This PR updates the year in copyright notices in all python files.


MateoLostanlen:
Hi @frgfm , thanks for the PR. I find it a bit absurd that we have to do this every year, isn't there another way of structuring the header?

frgfm:
Unfortunately, the alternative is to remove this altogether but if the range expires, then we have a problem. I've checked online and cannot find an unending year range for copyright notice so far :man_shrugging: 

On the commit history it looks like there is a lot but to be honest, changing that every year is rather easy (find & replace in the entire project)

PR #314, opened 2024-02-09
delete deadlink

fe51:
The full package doc is not available: deadlink. 

Hence, I suggest to delete this part of the doc until there is a new doc available. Maybe we have a proper link ? 

Happy to discuss it !


frgfm (approved):
Thanks for the PR :pray:

PR #317, opened 2024-04-05
fix(client): handle the final slash sensitivity on core routers

frgfm:
- reflects the final slash in the endpoint of base router (cf. https://github.com/pyronear/pyro-api/blob/main/src/app/api/endpoints/sites.py#L22) by modifying the client routes
- make sure the API is redirecting to the trailing slash endpoint in the backend to avoid problems later
- add a test case in the integration test script
- speeds up the dependencies installation in the Docker images

Any feedback is welcome!
cc @MateoLostanlen

### DIFF
--- a/.github/workflows/scripts.yml
+++ b/.github/workflows/scripts.yml
@@ -25,7 +25,12 @@ jobs:
           poetry-version: "1.6.1"
       - name: Resolve dependencies
         run: poetry export -f requirements.txt --without-hashes --output src/app/requirements.txt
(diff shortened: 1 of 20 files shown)


MateoLostanlen:
Thanks @frgfm  for fixing this bug and for adding the test, let me know when you've fixed it :smiley:

frgfm:
Alright, all tests passed, everything is backward compatible! I'll merge this to clean up other things in the mean time

PR #320, opened 2024-04-24
ci(github): add workflow to push image to Dockerhub

RonanMorgan:
II have created a test env which will be launched every night so it would be good to update our image in dockerhub regularly


.github/workflows/build-push-image.yml
```diff
@@ -0,0 +1,38 @@
+name: Build and Push Docker Image
+
+on:
+  pull_request:
+    types: [closed]
+    branches:
+      - main
```
frgfm, line 7: I think we should remove that: otherwise when a PR is closed but not merged, the workflow runs (and let's imagine it's because there are issues with it, we'll be in trouble)

frgfm (commented):
Thanks for the PR! Only a slight issue on the action trigger I think :)

frgfm (approved):
Cheers!

frgfm:
Oops, I just remember we already had this! The workflow trigger was not changed properly recently, but I pushed a fix a few hours ago #360 

Images are now pushed again to the hub so I think we can close this PR :)
Feel free to reopen if I missed something!

PR #322, opened 2024-05-02
refactor(models): implement the new data model

frgfm:
Following up on #304, this PR introduces the following modifications:
- data model
UML diagram

- update all routes
- update dependencies, the end to end test, the CI
- simplify docker orchestration by using localstack by default
- noticeable changes: the camera aren't "users" anymore, but an admin can create an unexpiring token for cameras (that means they can use a created token, but don't have credentials to log in). This reduces API calls & problems with creds

The only things missing are:
- [x] Updating the tests
- [x] Updating the client

In the following PRs, this is what will be added:
- wildfire: aggregation of corresponding detection along time periods & specific geographic areas.
- region/scope : limit what users can read depending on their spatial scope
- Webhooks: to enable easily subscribable webhooks for detection creation, and their confirmation.

Any feedback is welcome!


MateoLostanlen:
Hi @frgfm, Love this PR! It will make life easier for us and for pyronear newbies as well !

Just a few things:

To anticipate the use of camera ptz it would be preferable to have a tuple for azimuth in the camera table and to have an azimuth entry (float this time) in the detection table.

Localization is missing from the detection table too.

I'm not sure how you intend to associate a camera list with a user (sdis 7 for example). What do you have planned?

frgfm:
> To anticipate the use of camera ptz it would be preferable to have a tuple for azimuth in the camera table and to have an azimuth entry (float this time) in the detection table.

Could you remind me of how is a camera ptz different and why the tuple?

> Localization is missing from the detection table too.

It's by design: considering that at this stage, we report a detection from a camera, but we don't have more. But perhaps you mean the bounding box on the image itself? If so, I agree that we need to add it. Perhaps in the next PR to keep this one simple? Would a keypoint be enough? (I'm thinking about the usefulness, and the bounding box is linked to the type of model we use, but it's far from comprehensive, a keypoint would perhaps be more robust for the data model?)

> I'm not sure how you intend to associate a camera list with a user (sdis 7 for example). What do you have planned?

cf. the region/scope followup PR I mentioned in the description. Hesitating on the name, but I think "organizations" would work:
- table organization: id, name
- changes to other tables: add a "organization_id" to the user, and camera tables

Then the rest of the logic is easy to implement, and that doesn't complicate things too much

MateoLostanlen:
Basically, ptz cameras can be oriented along 3 axes. In our case, both our cameras and those of the Firefighters will patrol N predefined positions. It would therefore be interesting to have these N positions in the form of a tuple (or list of size N). Then, when the cam sends an alert, we'll add the azimuth to the detection table to find out the corresponding position.
For statis cams, we'll have a list / tuple of size 1 that will always be equal to the value sent to the detection table.


Yes, I'm talking about the bounding box, which is called localization in the current table, but I agree that the name can be changed.
I suggest you keep what we have at the moment, a string that allows you to send any form of points (bbox, keypoints etc ...)

Ok understood for the organization

MateoLostanlen:
Yes, don't worry about adding the bbox in a next pr as long as it's present when you deploy it. It's essential information for firefighters to speed up treatment time.

RonanMorgan (approved):

Makefile
```diff
@@ -29,10 +29,11 @@ stop:
 
 # Run tests for the library
 test:
-	poetry export -f requirements.txt --without-hashes --with test --output requirements.txt
-	docker compose -f docker-compose.dev.yml up -d --build --wait
-	docker compose exec -T backend pytest --cov=app
-	docker compose -f docker-compose.dev.yml down
+	-poetry export -f requirements.txt --without-hashes --with test --output requirements.txt
+	-docker compose -f docker-compose.dev.yml up -d --build --wait
+	-sleep 30
```
frgfm, line 34: Isn't the whole point of the --wait and the healthcheck that we don't need to sleep? :smile:

frgfm (commented):
Thanks @RonanMorgan! Not sure about the sleep in the make though? I thought we implemented healthcheck and add the --wait so that we don't need to wait after that

Or do you still see persisting problems?

Makefile
RonanMorgan: I still have many issues with the database initialization by alembic  if I don't add the sleep :/

RonanMorgan: it's not consistent, it happens one time out of two

frgfm: really? I think we need to investigate that in another PR, sleep is at best a flaky solution :/

src/app/crud/base.py
```diff
@@ -50,7 +50,7 @@ async def get(self, entry_id: int, strict: bool = False) -> Union[ModelType, Non
 
     async def get_by(self, field_name: str, val: Union[str, int], strict: bool = False) -> Union[ModelType, None]:
         statement: Select = select(self.model).where(getattr(self.model, field_name) == val)
-        results = await self.session.execute(statement=statement)
+        results = await self.session.exec(statement=statement)  # type: ignore
```
frgfm, line 53: Also let's no ignore all typing issues and be specific if we ignore any
Not sure why it was worth changing this line, since we're now ignore mypy on this line? Perhaps safer to address this in a more specific PR (if the only CI problem is with mypy temporarily, I think it's quite acceptable)

Makefile
```diff
@@ -29,10 +29,11 @@ stop:
 
 # Run tests for the library
 test:
-	poetry export -f requirements.txt --without-hashes --with test --output requirements.txt
-	docker compose -f docker-compose.dev.yml up -d --build --wait
-	docker compose exec -T backend pytest --cov=app
-	docker compose -f docker-compose.dev.yml down
+	-poetry export -f requirements.txt --without-hashes --with test --output requirements.txt
```
frgfm, line 32: btw, why are there leading hyphens here?

RonanMorgan: leading hyphens are really useful to launch the "docker compose down" even if there are any issue in the commands before

Makefile
RonanMorgan: yes or we can push a solution and find a better one later :) 
since it seems that there is not so much volunteer currently we can revert if you want but it seems to me that it's better to have something which works even if it not super clean

src/app/crud/base.py
RonanMorgan: oh yes sorry for that I forgot th precise the mypy error to ignore

Makefile
frgfm: My debugging mind feels like I'd actually like to have people with problems so we can find the source. But I agree that I didn't expect you to have any, I consistently get everything up & running from the get go :sweat_smile:

Feel free to open a PR after that to fix that bug (I imagine you have it on the other commands as well then?)

Makefile
frgfm: Oh thanks for teaching me, I felt like I didn't something there!
Mind opening a PR after that one to adjust this on all relevant make commands?

RonanMorgan: yes ok no pb !

frgfm:
cf #304

PR #323, opened 2024-05-03
Save media in a "site" bucket

RonanMorgan:
Pyronear

This is a first step to start this refactor : gathering images per site / installation.


frgfm:
Thanks for the PR :)
The description mostly comes from the illustration, so am I understanding this clearly: we create one bucket per entity (organization now I guess?)

RonanMorgan:
well the schema is a bit outdated : I did some calculation with OVH Cold Archive vs OVH S3 and for now it would cose only few euros to keep years of data

but yes I would like to have one bucket per organization :)

PR #324, opened 2024-05-03
Add start_date, end_date & pagination parameters

RonanMorgan:
In order to facilitate data science management I have refactored the endpoint "fetch_media" : I have added a start_date, an end_date to gather images during this time period and a pagination parameter in order to get images 10 by ten

=> still need to adapt pyro-client and to add test in the client part

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,7 @@ test:
 	poetry export -f requirements.txt --without-hashes --with dev --output src/app/requirements.txt
 	docker compose -f docker-compose.test.yml up -d --build
 	docker compose exec localstack awslocal s3 mb s3://sample-bucket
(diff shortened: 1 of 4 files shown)


RonanMorgan:
some thoughts / questions : 
=> this PR is useless, I should I have implemented the pagination code in the client part, not in the API part, no ? 
=> Can I add the boto dependency in the client ?
=> instead of using fetch_media I should have used get_media_url

frgfm:
I think the pagination idea is great for all the "fetch" routes, perhaps we should revisit this for all endpoints. Some rebasing is needed first though :sweat_smile:

RonanMorgan:
Yes I will go back to this pagination after the end of the big refactor :) 
I think I will create a brand new PR

PR #332, opened 2024-06-17
New Data Model : Add an Organization table  

RonanMorgan:
This table will be used to keep information about the owner of one or many cameras.

It will also be useful for resources access management (if you have access to the Site, you have access to all the resources linked to this site)

Pyronear UML


src/app/api/api_v1/endpoints/detections.py
```diff
@@ -26,10 +26,14 @@
...
     azimuth: float = Form(..., gt=0, lt=360, description="angle between north and direction in degrees"),
     file: UploadFile = File(..., alias="file"),
+    cameras: CameraCRUD = Depends(get_camera_crud),
     detections: DetectionCRUD = Depends(get_detection_crud),
     token_payload: TokenPayload = Security(get_jwt, scopes=[Role.CAMERA]),
 ) -> Detection:
     telemetry_client.capture(f"camera|{token_payload.sub}", event="detections-create")
+    camera = await cameras.get(token_payload.sub, strict=True)
+    if camera is not None and token_payload.organization_id != camera.organization_id:
+        raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail="Access forbidden.")
```
frgfm, line 36: Ah sorry, that might not have been clear: at this step, the token has already been verified as being a camera token. Because the camera is creating the detection, I don't think there is a need for the extra verification (i.e. the camera will stamp the detection with its org ID, or not even necessary since the detection has the cam ID)

In short, this verif can only failed if the token was forged (secret key hacked)

src/app/models.py
```diff
@@ -25,8 +25,15 @@ class Role(str, Enum):
     USER: str = "user"
 
 
+class OrganizationType(str, Enum):
+    SDIS: str = "sdis"
+    PARTICULIER: str = "particulier"
+    ADMIN: str = "admin"
```
frgfm, line 31: Curious which purpose you see in the org type? (vs having empty / non-empty org ID for the user & camera)

frgfm (commented):
Thanks for the PR!
Just a few high-level questions:
- whether admin should below to an org
- what you should be able to access when you're not in the org of the specific resource

The simple version in my mind is that org = ID + name, and we add organization_id to users & cameras table. Either we can create a pyronear org or consider that a resource without an org ID is an admin. What do you think?

RonanMorgan:
@frgfm about the pyronear org : it seems dangerous to me to say that an empty org has the admin right, it would be easy to imagine a bug conducting to the creation of an empty org that's why I don't really like this solution (but I might be wrong)

furthermore we could imagine some development resources only available to pyronear member, in this case we could also have non admin user attached to the pyronear org.

.env.example
```diff
@@ -13,6 +13,7 @@ S3_BUCKET_NAME=bucket
 # Initialization
 SUPERADMIN_LOGIN='pyroadmin'
 SUPERADMIN_PWD='LetsProtectForests!'
+SUPERADMIN_ORGANIZATION='pyronear'
```
frgfm, line 16: SUPERADMIN_ORG to keep it short :sweat_smile:

.github/workflows/builds.yml
```diff
@@ -27,6 +27,7 @@ jobs:
         env:
           SUPERADMIN_LOGIN: dummy_login
           SUPERADMIN_PWD: dummy&P@ssw0rd!
+          SUPERADMIN_ORGANIZATION: dummyorga
```
frgfm, line 30: same here

CONTRIBUTING.md
```diff
@@ -125,6 +125,7 @@ None :)
 - `POSTGRES_PASSWORD`: a password for the PostgreSQL database
 - `SUPERADMIN_LOGIN`: the login of the initial admin user
 - `SUPERADMIN_PWD`: the password of the initial admin user
+- `SUPERADMIN_ORGANIZATION`: the organization of the initial admin user
```
frgfm, line 128: same

Makefile
```diff
@@ -29,25 +29,28 @@ stop:
...
 # Run tests for the Python client
 test-client:
-	poetry export -f requirements.txt --without-hashes --output requirements.txt
-	docker compose -f docker-compose.dev.yml up -d --build --wait
-	cd client && pytest --cov=pyroclient tests/ && cd ..
-	docker compose -f docker-compose.dev.yml down
+	- poetry export -f requirements.txt --without-hashes --output requirements.txt
+	- docker compose -f docker-compose.dev.yml up -d --build --wait
+	- cd client && pytest --cov=pyroclient tests/ && cd ..
+	- docker compose -f docker-compose.dev.yml down
```
frgfm, line 45: Could you add a comment in this file explaining the tab + union please? I didn't know about that, let's try to avoid having someone remove those in the future thinking it was a mistake :)

client/README.md
```diff
@@ -39,8 +39,8 @@ api_client = client.Client(API_URL, USER_LOGIN, USER_PASSWORD)
 
 # List your registered devices
 devices = api_client.get_my_devices().json()
-# List sites accessible in your scope
-sites = api_client.get_sites().json()
+# List organizations accessible in your scope
+organizations = api_client.get_organizations().json()
```
frgfm, line 43: I just realized the README is completely out of date, could you remove the legacy methods?

client/pyroclient/client.py
```diff
@@ -30,6 +30,10 @@
     "detections-label": "/detections/{det_id}/label",
     "detections-fetch": "/detections",
     "detections-url": "/detections/{det_id}/url",
+    #################
+    # DETECTIONS
```
frgfm, line 34: orgs you mean?

client/README.md
frgfm: Oh and you created the method "fetch_organizations", let's pick a side and align both

client/pyroclient/client.py
```diff
@@ -182,3 +186,21 @@ def fetch_detections(self) -> Response:
             headers=self.headers,
             timeout=self.timeout,
         )
+
+    # ORGANIZATIONS
+
+    def fetch_organizations(self) -> Response:
```
frgfm, line 192: cf. comment in the readme

client/tests/conftest.py
```diff
@@ -7,8 +7,11 @@
 API_URL = os.getenv("API_URL", "http://localhost:5050/api/v1/")
 SUPERADMIN_LOGIN = os.getenv("SUPERADMIN_LOGIN", "superadmin_login")
 SUPERADMIN_PWD = os.getenv("SUPERADMIN_PWD", "superadmin_pwd")
+SUPERADMIN_ORGANIZATION = os.getenv("SUPERADMIN_ORGANIZATION", "superadmin_organization")
```
frgfm, line 10: I'd suggest organization --> org everywhere for the env variable to keep it short

docker-compose.dev.yml
```diff
@@ -46,6 +46,7 @@ services:
       - POSTGRES_URL=postgresql+asyncpg://dummy_pg_user:dummy_pg_pwd@db/dummy_pg_db
       - SUPERADMIN_LOGIN=superadmin_login
       - SUPERADMIN_PWD=superadmin_pwd
+      - SUPERADMIN_ORGANIZATION=superadmin_organization
```
frgfm, line 49: same

docker-compose.yml
```diff
@@ -52,6 +52,7 @@ services:
       - POSTGRES_URL=postgresql+asyncpg://${POSTGRES_USER}:${POSTGRES_PASSWORD}@db/${POSTGRES_DB}
       - SUPERADMIN_LOGIN=${SUPERADMIN_LOGIN}
       - SUPERADMIN_PWD=${SUPERADMIN_PWD}
+      - SUPERADMIN_ORGANIZATION=${SUPERADMIN_ORGANIZATION}
```
frgfm, line 55: same

src/app/api/api_v1/endpoints/cameras.py
```diff
@@ -27,6 +27,8 @@ async def register_camera(
     token_payload: TokenPayload = Security(get_jwt, scopes=[UserRole.ADMIN, UserRole.AGENT]),
 ) -> Camera:
     telemetry_client.capture(token_payload.sub, event="cameras-create", properties={"device_login": payload.name})
+    if token_payload.organization_id != payload.organization_id and UserRole.ADMIN not in token_payload.scopes:
```
frgfm, line 30: Perhaps a suggestion:
- make org_ID optional in the payload
- if it's none, defaults to the org ID of the authenticated user
- if it's not none, it forwards it if this is an admin, but throws an error if it's an agent "Your role is not authorized to override the field "organization_id""

src/app/api/api_v1/endpoints/cameras.py
```diff
@@ -46,7 +51,10 @@ async def fetch_cameras(
     token_payload: TokenPayload = Security(get_jwt, scopes=[UserRole.ADMIN, UserRole.AGENT, UserRole.USER]),
 ) -> List[Camera]:
     telemetry_client.capture(token_payload.sub, event="cameras-fetch")
-    return [elt for elt in await cameras.fetch_all()]
+    all_cameras = [elt for elt in await cameras.fetch_all()]
+    if UserRole.ADMIN in token_payload.scopes:
+        return all_cameras
+    return [camera for camera in all_cameras if camera.organization_id == token_payload.organization_id]
```
frgfm, line 57: Not sure but can't we make a single list comprehension?
```python
return [elt for elt in await cameras.fetch_all() if UserRole.ADMIN in token_payload.scopes or elt.organization_id == token_payload.organization_id]
```

src/app/api/api_v1/endpoints/cameras.py
```diff
@@ -55,7 +63,9 @@ async def heartbeat(
     token_payload: TokenPayload = Security(get_jwt, scopes=[Role.CAMERA]),
 ) -> Camera:
     # telemetry_client.capture(f"camera|{token_payload.sub}", event="cameras-heartbeat", properties={"camera_id": camera_id})
-    await cameras.get(token_payload.sub, strict=True)
+    camera = await cameras.get(token_payload.sub, strict=True)
+    if camera is not None and token_payload.organization_id != camera.organization_id:
+        raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail="Access forbidden.")
```
frgfm, line 68: The camera is updating itself by design, so actually we can even remove the previous line. With the new version, I think we'll have a problem: if the camera is None, we're gonna get a problem when updating the camera next.

I suggest that we remove the former & newer lines. If the token is forged we have bigger problems. If it's not, it means it comes from a camera, and that camera would be updating itself. (even if somehow in the future we allow admins to update the org ID of the camera between the moment the token is created and the heartbeat, I would say we still want the camera to be able to update itself)

I don't think we should run any org verification for "self" methods

src/app/api/api_v1/endpoints/cameras.py
```diff
@@ -66,10 +76,9 @@ async def create_camera_token(
     token_payload: TokenPayload = Security(get_jwt, scopes=[UserRole.ADMIN]),
 ) -> Token:
     telemetry_client.capture(token_payload.sub, event="cameras-token", properties={"camera_id": camera_id})
-    # Verify camera
     await cameras.get(camera_id, strict=True)
     # create access token using user user_id/user_scopes
-    token_data = {"sub": str(camera_id), "scopes": ["camera"]}
+    token_data = {"sub": str(camera_id), "scopes": ["camera"], "organization_id": token_payload.organization_id}
```
frgfm, line 81: If I'm not mistaking, in the JWT payload, we want the org ID of the camera, not the authenticated admin

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -64,23 +64,40 @@ async def create_detection(
...
-    return cast(Detection, await detections.get(detection_id, strict=True))
+    detection = cast(Detection, await detections.get(detection_id, strict=True))
+    camera = await cameras.get(detection.camera_id, strict=True)
+    if (
+        camera is not None
+        and token_payload.organization_id != camera.organization_id
+        and UserRole.ADMIN not in token_payload.scopes
+    ):
+        raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail="Access forbidden.")
+    return detection
```
frgfm, line 80: more of a product discussion than backend: I agree that org should prevent cross access on users, and cameras. On detection, I don't know :thinking: 

If we want to do it:
- let's add a `if UserTole.ADMIN in token_payload.scopes` and return the previous line
- the other verif (otherwise we fetch a camera for nothing when this is an admin, considering the camera_id verification will happen during the DB operation either way)

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -64,23 +64,40 @@ async def create_detection(
...
 async def get_detection_url(
     detection_id: int = Path(..., gt=0),
+    cameras: CameraCRUD = Depends(get_camera_crud),
     detections: DetectionCRUD = Depends(get_detection_crud),
     token_payload: TokenPayload = Security(get_jwt, scopes=[UserRole.ADMIN, UserRole.AGENT, UserRole.USER]),
 ) -> DetectionUrl:
     """Resolve the temporary media image URL"""
-    telemetry_client.capture(token_payload.sub, event="detections-url", properties={"detection_id": detection_id})
     # Check in DB
+    telemetry_client.capture(token_payload.sub, event="detections-url", properties={"detection_id": detection_id})
```
frgfm, line 92: I imagine this was a revert but it's better when the comment matches the line right below it

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -64,23 +64,40 @@ async def create_detection(
...
     detection = cast(Detection, await detections.get(detection_id, strict=True))
+    camera = await cameras.get(detection.camera_id, strict=True)
+    if (
+        camera is not None
+        and token_payload.organization_id != camera.organization_id
+        and UserRole.ADMIN not in token_payload.scopes
+    ):
+        raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail="Access forbidden.")
```
frgfm, line 100: same here

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -89,20 +106,38 @@ async def get_detection_url(
...
-    return [elt for elt in await detections.fetch_all()]
+    all_detections = [elt for elt in await detections.fetch_all()]
+    if UserRole.ADMIN in token_payload.scopes:
+        return all_detections
+    filtered_detections = []
+    for detection in all_detections:
+        camera = await cameras.get(detection.camera_id, strict=True)
+        if camera is not None and camera.organization_id == token_payload.organization_id:
+            filtered_detections.append(detection)
+    return filtered_detections
```
frgfm, line 121: Mmmh this is why I would also argue that org shouldn't prevent read on detections. This step is quite heavy on the DB once there are a lot of detections. We're talking about switching from 1 fetch operations on N rows, to 1 fetch + N get. I don't think we can afford this

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -89,20 +106,38 @@ async def get_detection_url(
...
 ) -> Detection:
     telemetry_client.capture(token_payload.sub, event="detections-label", properties={"detection_id": detection_id})
+    detection = cast(Detection, await detections.get(detection_id, strict=True))
+    camera = await cameras.get(detection.camera_id, strict=True)
+    if (
+        camera is not None
+        and token_payload.organization_id != camera.organization_id
+        and UserRole.ADMIN not in token_payload.scopes
+    ):
+        raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail="Access forbidden.")
```
frgfm, line 140: same, let's verify if admin to allow the shortcut

src/app/core/config.py
```diff
@@ -25,6 +25,7 @@ class Settings(BaseSettings):
     # Authentication
     SUPERADMIN_LOGIN: str = os.environ["SUPERADMIN_LOGIN"]
     SUPERADMIN_PWD: str = os.environ["SUPERADMIN_PWD"]
+    SUPERADMIN_ORGANIZATION: str = os.environ["SUPERADMIN_ORGANIZATION"]
```
frgfm, line 28: same

src/app/crud/base.py
```diff
@@ -28,11 +28,11 @@ async def create(self, payload: CreateSchemaType) -> ModelType:
         try:
             self.session.add(entry)
             await self.session.commit()
-        except exc.IntegrityError:
+        except exc.IntegrityError as error:
             await self.session.rollback()
             raise HTTPException(
                 status_code=status.HTTP_409_CONFLICT,
-                detail="An entry with the same index already exists.",
+                detail="An entry with the same index already exists : " + str(error),
```
frgfm, line 35: f-string please :)

src/app/models.py
frgfm: I still think for now, we should remove the type. It serves no purpose for now I believe, so let's keep simpler schemas as long as possible :crossed_fingers:

src/app/schemas/organizations.py
```diff
@@ -0,0 +1,29 @@
...
+__all__ = ["OrganizationCreate", "OrganizationUpdate"]
+
+
+class OrganizationCreate(BaseModel):
+    name: str = Field(
+        ...,
+        min_length=3,
+        max_length=50,
+        description="name of the camera",
+        json_schema_extra={"examples": ["pyro-camera-01"]},
```
frgfm, line 17: I think you forgot to edit the example value

src/app/schemas/organizations.py
```diff
@@ -0,0 +1,29 @@
...
+    type: str
+
+
+class OrganizationUpdate(BaseModel):
+    name: str = Field(
+        ...,
+        min_length=3,
+        max_length=50,
+        description="name of the camera",
+        json_schema_extra={"examples": ["pyro-camera-01"]},
```
frgfm, line 28: same here

src/app/api/api_v1/endpoints/organizations.py
```diff
@@ -0,0 +1,69 @@
...
+router = APIRouter()
+
+
+@router.post("/", status_code=status.HTTP_201_CREATED, summary="Register a new organization")
+async def register_organization(
+    payload: OrganizationCreate,
+    organizations: OrganizationCRUD = Depends(get_organization_crud),
+    token_payload: TokenPayload = Security(get_jwt, scopes=[UserRole.ADMIN]),
+) -> Organization:
+    telemetry_client.capture(token_payload.sub, event="organization-create", properties={"device_login": payload.name})
```
frgfm, line 27: What I usually do is that I add as properties the target when it's a GET/UPDATE/DELETE. I'd suggest removing this property, especially because this isn't a device_login

src/app/api/api_v1/endpoints/organizations.py
```diff
@@ -0,0 +1,69 @@
...
+    organizations: OrganizationCRUD = Depends(get_organization_crud),
+    token_payload: TokenPayload = Security(get_jwt, scopes=[UserRole.ADMIN, UserRole.AGENT, UserRole.USER]),
+) -> Organization:
+    telemetry_client.capture(
+        token_payload.sub, event="organizations-get", properties={"organization_id": organization_id}
+    )
+    organization = cast(Organization, await organizations.get(organization_id, strict=True))
+    if token_payload.organization_id != organization.id and UserRole.ADMIN not in token_payload.scopes:
+        raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail="Access forbidden.")
+    return organization
```
frgfm, line 45: Let's restrict org routes to admins for now, that will simplify things greatly (potentially later if we need, we'll create a get_my_org but I don't think we need much interactions for now)

frgfm (commented):
Thanks for the edits! Lengthy review I'm sorry, but I did spot a few issues here and there + added some discussion questions

Let me know what you think :+1:

src/app/api/api_v1/endpoints/cameras.py
RonanMorgan: I really don't like that org_id could be None, for me it's a security hole

src/app/api/api_v1/endpoints/cameras.py
RonanMorgan: it seems to me that it's more readable in three lines, no ?

src/app/api/api_v1/endpoints/cameras.py
RonanMorgan: ah yes sorry I fixed that in a following PR by mistake

client/README.md
RonanMorgan: Should I create some examples for "camera" (instead of the device bloc) or we can consider that our user should not play with this part of the API, only the engine ?

src/app/api/api_v1/endpoints/cameras.py
RonanMorgan: hum I don't understand the issue, why a camera would be None ?

src/app/api/api_v1/endpoints/detections.py
RonanMorgan: yes I thought about it, but we could implement it differently in our crud with a GROUP BY so i thought that I would do it later.

RonanMorgan: I can do it now

client/README.md
frgfm: The prime users here are the camera for the detection workflow, the agents through the platform. Illustrating both would be ideal 👍

src/app/api/api_v1/endpoints/cameras.py
frgfm: Oh I got confused by the
´´´python
if camera is not None
´´´

The 'strict=True' throws an error in case the camera isn't found (none), so would you mind removing this conditional check?

src/app/api/api_v1/endpoints/detections.py
frgfm: Happy with any sql ops that avoids this heavy load 🙂

src/migrations/versions/2024_06_17_1521-4265426f8438_create_stes_table.py
```diff
@@ -0,0 +1,44 @@
...
+    op.create_table(
+        "organization",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("name", sqlmodel.sql.sqltypes.AutoString(), nullable=False),
+        sa.PrimaryKeyConstraint("id"),
+        sa.UniqueConstraint("name"),
+    )
+
+    # Add the 'organization_id' column to the 'camera' and 'user' tables and create  foreign key constraints
+    op.add_column("camera", sa.Column("organization_id", sa.Integer(), nullable=True))
```
frgfm, line 32: Open question: for the migration to be reversible, how do we fill the org_id value for existing entries?

src/tests/endpoints/test_organizations.py
```diff
@@ -0,0 +1,166 @@
...
+        assert response.json()["detail"] == status_detail
+    if response.status_code // 100 == 2:
+        assert response.json() == pytest.organization_table[expected_idx]
+
+
+@pytest.mark.parametrize(
+    ("user_idx", "status_code", "status_detail", "expected_response"),
+    [
+        (None, 401, "Not authenticated", None),
+        (0, 200, None, pytest.organization_table[0]),
```
frgfm, line 103: Isn't it supposed to be the full list?

frgfm (commented):
Thanks a lot! Feels almost finalized, just a few details 🙂

src/app/crud/base.py
```diff
@@ -28,11 +28,11 @@ async def create(self, payload: CreateSchemaType) -> ModelType:
         try:
             self.session.add(entry)
             await self.session.commit()
-        except exc.IntegrityError:
+        except exc.IntegrityError as error:
             await self.session.rollback()
             raise HTTPException(
                 status_code=status.HTTP_409_CONFLICT,
-                detail="An entry with the same index already exists.",
+                detail=f"An entry with the same index already exists : {error!s}",
```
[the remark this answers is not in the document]
RonanMorgan: hum I wrote str(error), this was an automatic conversion from mypy I think

src/migrations/versions/2024_06_17_1521-4265426f8438_create_stes_table.py
RonanMorgan: good question : do we really want the migration to be reversible ? :)

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -64,45 +64,77 @@ async def create_detection(
...
 ) -> List[Detection]:
     telemetry_client.capture(token_payload.sub, event="detections-fetch")
-    return [elt for elt in await detections.fetch_all()]
+    if UserRole.ADMIN in token_payload.scopes:
+        return [elt for elt in await detections.fetch_all()]
+
+    cameras_list = await cameras.get_all_by("organization_id", token_payload.organization_id, strict=True)
+    camera_ids = [camera.id for camera in cameras_list]
+
+    return await detections.get_in(camera_ids, "camera_id")
```
frgfm, line 118: Perfect for the SQL ops. One feedback though: let's improve those API
- `get_all_by` --> you already have `fetch_all` using the `filter_pair` argument
```python
cameras_list = await cameras.fetch_all(("organization_id", token_payload.organization_id))
```
- `get_in`  --> I suggest extending `fetch_all` instead with a new arg "" to allow in filter, or by extending "filter_pair" to have either `Tuple[str, scalar]` or `Tuple[str, Sequence[scalar]]` type (in second case, do "in" operation rather than "where". To get something like:
```python
return await detections.fetch_all(("camera_id", camera_ids))
```

src/app/crud/base.py
```diff
@@ -58,11 +58,23 @@ async def get_by(self, field_name: str, val: Union[str, int], strict: bool = Fal
...
+        results = await self.session.exec(statement=statement)
+        entries = results.all()
+
+        if strict and not entries:
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND,
+                detail=f"Table {self.model.__name__} has no corresponding entry.",
+            )
+        return entries
+
```
frgfm, line 72: cf previous comment. Apart from the strict (which we can check with `len` here), you have everything in `fetch_all`

src/app/crud/base.py
```diff
@@ -79,7 +91,12 @@ async def update(self, entry_id: int, payload: UpdateSchemaType) -> ModelType:
...
-        statement = delete(self.model).where(self.model.id == entry_id)  # type: ignore[attr-defined]
+        statement = delete(self.model).where(self.model.id == entry_id)
 
         await self.session.exec(statement=statement)  # type: ignore[call-overload]
         await self.session.commit()
+
+    async def get_in(self, list_: List[Any], field_name: str) -> List[ModelType]:
+        statement = select(self.model).where(getattr(self.model, field_name).in_(list_))  # type: ignore[var-annotated]
+        results = await self.session.exec(statement)
+        return results.all()
```
frgfm, line 102: same here, because to adapt the `filter_pair` mechanism of `fetch_all`

src/app/main.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (C) 2020-2024, Pyronear.
+# Copyright (C) 2024, Pyronear.
```
frgfm, line 1: !

src/app/schemas/base.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (C) 2020-2024, Pyronear.
+# Copyright (C) 2024, Pyronear.
```
frgfm, line 1: !

src/app/schemas/cameras.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (C) 2020-2024, Pyronear.
+# Copyright (C) 2024, Pyronear.
```
frgfm, line 1: !

src/app/schemas/detections.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (C) 2020-2024, Pyronear.
+# Copyright (C) 2024, Pyronear.
```
frgfm, line 1: !

src/app/schemas/login.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (C) 2020-2024, Pyronear.
+# Copyright (C) 2024, Pyronear.
```
frgfm, line 1: !

src/app/schemas/users.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (C) 2020-2024, Pyronear.
+# Copyright (C) 2024, Pyronear.
```
frgfm, line 1: !

src/app/api/api_v1/endpoints/detections.py
RonanMorgan: done in the Acknowledged PR

client/pyroclient/exceptions.py
```diff
@@ -1,5 +1,4 @@
 # Copyright (C) 2020-2024, Pyronear.
-
```
frgfm, line 2: Missing the blank line here (CI fail)

frgfm (commented):
Looks like there is only a blank line + a rebase/merge to resolve conflicts and we're good to go :)

frgfm:
There are still conflicts, but it's strange the last commit on main was 18hours ago, while your last push is from 12 hours ago. Are you merging the last version of main?

frgfm (approved):
Thanks a lot!

frgfm:
cf. #304

PR #336, opened 2024-06-26
refactor(organization): switch to one-bucket-per-org policy

RonanMorgan:
It seems to me that it's a bit more secure and the images will be simpler to gather for data scientist if we save all the images of the same organization in one bucket. Furthermore I have added the camera_id in the name of the picture.


client/pyroclient/client.py
```diff
@@ -26,7 +26,7 @@
     #################
     # DETECTIONS
     #################
-    "detections-create": "/detections",
+    "detections-create": "/detections/",
```
frgfm, line 29: this shouldn't be a problem :thinking: 
Could you revert this and try to find the problem? I'm supposed to have solved it in #317

frgfm: Besides this is unrelated to the PR :sweat_smile:

client/tests/conftest.py
```diff
@@ -10,7 +10,7 @@
 SUPERADMIN_ORGANIZATION = os.getenv("SUPERADMIN_ORGANIZATION", "superadmin_organization")
 SUPERADMIN_TOKEN = requests.post(
     urljoin(API_URL, "login/creds"),
-    data={"username": SUPERADMIN_LOGIN, "password": SUPERADMIN_PWD, "organization": SUPERADMIN_ORGANIZATION},
+    data={"username": SUPERADMIN_LOGIN, "password": SUPERADMIN_PWD},
```
frgfm, line 13: I don't follow this modification?

scripts/dbdiagram.txt
```diff
@@ -34,12 +35,33 @@ Table "Camera" as C {
 Table "Detection" as D {
   "id" int [not null]
   "camera_id" int [ref: > C.id, not null]
+  "wildfire_id" int [ref: > W.id, not null]
   "azimuth" float [not null]
   "bucket_key" str [not null]
   "is_wildfire" bool
+  "bbox" boxes [not null]
```
frgfm, line 42: not yet?

scripts/dbdiagram.txt
frgfm, line 1: perhaps let's do those modifications in the correct PRs, otherwise it's a bit confusing

scripts/localstack/setup-s3.sh
```diff
@@ -1,2 +1,4 @@
 #!/usr/bin/env bash
-awslocal s3 mb s3://bucket
+awslocal s3 mb s3://admin
+awslocal s3 mb s3://organization1
+awslocal s3 mb s3://organization2
```
frgfm, line 4: Could you explain a bit?
Initially this was creating the default bucket required. Considering the purpose is the same, if the admin is the only user created, his org needs to be created, which is supposed to be the only bucket created. So I don't understand why we'd create 3

scripts/test_e2e.py
```diff
@@ -84,13 +84,6 @@ def main(args):
 
     cam_auth = {"Authorization": f"Bearer {cam_token}"}
 
-    # payload = {
-    #    "camera_id": cam_id,
-    #    "azimuth": 110,
-    #    "starting_time": "2023-11-07T15:08:19.226673",
-    # }
-    # wildfire_id = api_request("post", f"{args.endpoint}/wildfires/", cam_auth, payload)["id"]
-
```
frgfm, line 93: Agreed but not related to the PR

src/app/crud/crud_wildfire.py
```diff
@@ -1,7 +1,7 @@
-# Copyright (C) 2024, Pyronear.
```
frgfm, line 1: This file would have been created in 2024 so no need to edit that part :)

frgfm (commented):
Thanks for the PR! I greatly agree with the high-level idea, but I'm a bit struggling with the review since a few PRs are intertwined. A few suggestions:
- would be easier if this PR was stacked on top of the org PR, not the wildfire PR (since it's unrelated)
- in my mind, the only thing that should change in our existing routes would be: a bucket is created when an org is created (same for deletion), and we save the image of the detection in the correct bucket (+ URL generation).

For simplicity, I'd also suggest that the superadmin org cannot be edited, it's "admins" by default. So that you can safely initialize the bucket accordingly and no need for dynamic considerations

client/pyroclient/client.py
RonanMorgan: I had a 307 redirection in the log, not really an issue but still

client/tests/conftest.py
RonanMorgan: I thought we need this to create the token but when we create the token the function is searching the user in the db and can take the organization_id from it

scripts/dbdiagram.txt
RonanMorgan: ah yes it's in my todo

scripts/localstack/setup-s3.sh
RonanMorgan: because once you told me that you didn't want automatic creation of the buckets to avoid cost inflation, so in production this operation would stay manual. In test I put it here

src/app/services/storage.py
```diff
@@ -26,70 +26,87 @@ class S3Bucket:
...
-    def __init__(
-        self, region: str, endpoint_url: str, access_key: str, secret_key: str, bucket_name: str, proxy_url: str
-    ) -> None:
+    def __init__(self, region: str, endpoint_url: str, access_key: str, secret_key: str, proxy_url: str) -> None:
         _session = boto3.Session(access_key, secret_key, region_name=region)
         self._s3 = _session.client("s3", endpoint_url=endpoint_url)
         # Ensure S3 is connected
         try:
-            self._s3.head_bucket(Bucket=bucket_name)
+            self._s3.head_bucket(Bucket="admin")
```
frgfm, line 37: I would suggest we make this a hardcoded config variable (just so we make sure that it's uniform across the entire app & controlled by a single source of truth)

Here for instance there is a discrepancy with the initialization script I think :)

src/app/services/storage.py
```diff
@@ -26,70 +26,87 @@ class S3Bucket:
...
         except EndpointConnectionError:
             raise ValueError(f"unable to access endpoint {endpoint_url}")
         except ClientError:
-            raise ValueError(f"unable to access bucket {bucket_name}")
-        logger.info(f"S3 bucket {bucket_name} connected on {endpoint_url}")
-        self.bucket_name = bucket_name
+            raise ValueError("unable to access bucket admin")
```
frgfm, line 41: same here then

frgfm (commented):
Thanks for the PR! high level comments:
- we need to harmonize the bucket name in the initialization script with the API source code. I suggest having a harccoded config variable bearing the bucket name for `src/app`
- I don't think we should create buckets when one doesn't exist for detection creation for instance

Let me know what you think :)

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -50,12 +54,25 @@ async def create_detection(
         # Reset byte position of the file (cf. https://fastapi.tiangolo.com/tutorial/request-files/#uploadfile)
         await file.seek(0)
 
+        # Check if bucket exists, if not, create it
```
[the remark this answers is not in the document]
RonanMorgan: well an error will be thrown if we try to save a file in a non existent bucket so I can just remove this part, non ?

src/app/services/storage.py
RonanMorgan: hum indeed. Do we really want to create a bucket ? it should not even be used

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -26,6 +26,10 @@
 router = APIRouter()
 
 
+def get_bucket_name(organization_id: int) -> str:
+    return f"alert-api-{organization_id!s}"
```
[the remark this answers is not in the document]
RonanMorgan: hum in fact we don't need an admin bucket , and for the other organization it will be created automatically si we should be fine I suppose ?

src/app/api/api_v1/endpoints/organizations.py
```diff
@@ -27,7 +32,12 @@ async def register_organization(
     telemetry_client.capture(
         token_payload.sub, event="organization-create", properties={"organization_name": payload.name}
     )
-    return await organizations.create(payload)
+    organization = await organizations.create(payload)
```
frgfm, line 35: Mmmh I was about to suggest we do this after making sure the S3 ops haven't failed, but we need the org ID for the bucket :thinking:

src/app/api/api_v1/endpoints/organizations.py
frgfm, line 1: Shouldn't we also delete the bucket when the org is deleted? it was the core motivation I think

src/app/services/storage.py
```diff
@@ -26,70 +26,81 @@ class S3Bucket:
...
         return True
 
-    async def delete_file(self, bucket_key: str) -> None:
+    async def delete_file(self, bucket_key: str, bucket_name: str) -> None:
         """Remove bucket file and return whether the deletion succeeded"""
         # https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.delete_object
-        self._s3.delete_object(Bucket=self.bucket_name, Key=bucket_key)
+        self._s3.delete_object(Bucket=bucket_name, Key=bucket_key)
+
+    async def create_bucket(self, bucket_name: str) -> bool:
```
frgfm, line 85: We also need "delete_bucket" :)

frgfm (commented):
Thanks! I added a few comments, and since the base branch isn't main, there are things in those commits which are already on main :sweat_smile:

src/app/api/api_v1/endpoints/organizations.py
RonanMorgan: we can catch Error from the S3 ops and return an error ?

src/app/api/api_v1/endpoints/organizations.py
RonanMorgan: with a try catch also ?

RonanMorgan: btw are we sure we want that ? maybe we will need the images for the ML part @MateoLostanlen

scripts/localstack/setup-s3.sh
```diff
@@ -1,2 +1,4 @@
 #!/usr/bin/env bash
-awslocal s3 mb s3://bucket
+awslocal s3 mb s3://admin
+awslocal s3 mb s3://alert-api-1
+awslocal s3 mb s3://alert-api-2
```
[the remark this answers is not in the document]
frgfm: That could work, until now what I've done:
- let's say the bare docker compose is to run everything locally (assuming you only have your single laptop, or server)
- the dev one is the most temporary, I'll rename it to "test" because it's actually used for tests so we don't preserve any data, everything's wiped clean

So we can keep this for the test, I'll handle that later when I clean compose files

frgfm:
With the recent updates, here is what the PR does:
- switch from "one bucket for all" strategy to "one bucket per organization"
- cleans S3 storage interactions (check later #362)
- add proper storage initialization in db.py
- update tests accordingly

frgfm:
cf. #304

PR #340, opened 2024-07-04
feat(detections): add bboxes column to detection and a route to fetch unacknowledged ones

RonanMorgan:
Add the acknowledged boolean in the table Detection and a new endpoint. It will allow the plateform to know if a Detection has been acknowledged by a User.


src/app/api/api_v1/endpoints/detections.py
```diff
@@ -128,6 +132,40 @@ async def fetch_detections(
...
+    cameras: CameraCRUD = Depends(get_camera_crud),
+    token_payload: TokenPayload = Security(get_jwt, scopes=[UserRole.ADMIN, UserRole.AGENT, UserRole.USER]),
+) -> List[Detection]:
+    telemetry_client.capture(token_payload.sub, event="unacknowledged-fetch")
+
+    try:
+        all_unck_detections = [elt for elt in await detections.fetch_all() if elt.is_wildfire is None]
+        if from_date is not None:
+            all_unck_detections = [detection for detection in all_unck_detections if detection.created_at >= from_date]
+    except Exception as e:  # noqa
```
frgfm, line 148: I usually avoid this: if this throws an error, let's leave it to sentry. It will be much easier to debug

Otherwise let's have very specific exceptions

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -128,6 +132,40 @@ async def fetch_detections(
...
+        logging.error(f"Error fetching detections: {e}")
```
frgfm, line 149: Cf above

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -128,6 +132,40 @@ async def fetch_detections(
...
+        raise HTTPException(status_code=500, detail="Internal server error")
+
+    if UserRole.ADMIN in token_payload.scopes:
+        return all_unck_detections
+
+    filtered_detections = []
+    for detection in all_unck_detections:
+        try:
+            camera = await cameras.get(detection.camera_id, strict=True)
+        except Exception as e:  # noqa
```
frgfm, line 159: Either we let strict=True throw an error or we set strict=False and check whether it's None to avoid the try except

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -128,6 +132,40 @@ async def fetch_detections(
...
+            all_unck_detections = [detection for detection in all_unck_detections if detection.created_at >= from_date]
+    except Exception as e:  # noqa
+        logging.error(f"Error fetching detections: {e}")
+        raise HTTPException(status_code=500, detail="Internal server error")
+
+    if UserRole.ADMIN in token_payload.scopes:
+        return all_unck_detections
+
+    filtered_detections = []
+    for detection in all_unck_detections:
```
frgfm, line 156: Overlap with the other PR right?

src/migrations/versions/2024_06_17_1521-4265426f8438_create_stes_table.py
```diff
@@ -41,4 +41,4 @@ def downgrade() -> None:
     op.drop_constraint("fk_user_orga", "user", type_="foreignkey")
     op.drop_column("camera", "organization_id")
     op.drop_column("user", "organization_idation_id")
-    op.drop_table("detection")
+    op.drop_table("organization")
```
frgfm, line 44: This needs to be on the other PR 😅

client/pyroclient/client.py
```diff
@@ -29,6 +29,7 @@
     "detections-create": "/detections",
     "detections-label": "/detections/{det_id}/label",
     "detections-fetch": "/detections",
+    "detections-fetch-unck": "/detections/unacknowledged/from",
```
frgfm, line 32: From is a bit confusing in the route url: I first wasn't sure whether it was "from camera ID" or something else

Open question about the exact need: 
- if we need the last 24h, let's not leave the option
- otherwise a time range or pagination could be useful?

frgfm (commented):
Thanks! High level question:
@MateoLostanlen @pyronear/front-end what is the minimal necessary information we need about unacknowledged détections?

I think it's a fallback until we have proper notification system. But let's make sure we get the exact thing right:
- do we plan on letting users specify a time range?
- if not, what's the time range? Last 24hours? Longer?
- if there are too many (and there will), how many should we keep and in which order should they be? (The N latest, the N oldest)

src/app/api/api_v1/endpoints/detections.py
RonanMorgan: I don't have sentry in development. In this case I could create a loggger which send to sentry and terminal ?

frgfm: But the point is catching that in prod, isn't it? I don't get what you mean by "in development"
To me, we only want to catch things that we know specifically that can happen. but a wide `Exception` will catch anything, even things we don't know and would like to fix.
Considering the worker restarts in a clean way when it crashes, I prefer this option because we get the appropriate signal (sentry notif with trace & everything). So we can log some information but my problem is more with the `except Exception`

RonanMorgan: I raise the exception after (L140)

When I develop I have always lots of trouble to debug because I don't see error in the log

client/pyroclient/client.py
RonanMorgan: @MateoLostanlen 

I can juste write fromDate ?
(we cannot put nothing because FastAPI think we try do call the detections/ endpoints with the Id=unacknowledged

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -4,9 +4,10 @@
 # See LICENSE or go to <https://www.apache.org/licenses/LICENSE-2.0> for full license details.
 
 import hashlib
+import logging
```
frgfm, line 7: not ussing it

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -19,6 +20,9 @@
 from app.services.storage import s3_bucket
 from app.services.telemetry import telemetry_client
 
+logging.basicConfig(level=logging.ERROR)
+logging.getLogger("uvicorn.error")
+
```
frgfm, line 25: same

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -4,9 +4,10 @@
 # See LICENSE or go to <https://www.apache.org/licenses/LICENSE-2.0> for full license details.
 
 import hashlib
+import logging
 from datetime import datetime
 from mimetypes import guess_extension
-from typing import List, cast
+from typing import List, Optional, cast
```
frgfm, line 10: Oh, with the next version of python, it's been widely discussed in the community, but basically, we have to change:
```python
Optional[TypeABC]
 ```
into:
```python
Union[TypeABC, None]
```

client/pyroclient/client.py
```diff
@@ -3,7 +3,7 @@
 # This program is licensed under the Apache License 2.0.
 # See LICENSE or go to <https://opensource.org/licenses/Apache-2.0> for full license details.
 
-from typing import Dict
+from typing import Dict, Optional
```
frgfm, line 6: Let use `Union[Type, None]` intead (cf. comment later)

client/pyroclient/client.py
```diff
@@ -187,6 +188,24 @@ def fetch_detections(self) -> Response:
             timeout=self.timeout,
         )
 
+    def fetch_unacknowledged_detections(self, from_date: Optional[str] = None) -> Response:
```
frgfm, line 191:
```python
Union[str, None] = None
```

client/pyroclient/client.py
frgfm: Oh good point, how about (not sure about PascalCasing in URL):
`/detections/unlabeled/fromdate`

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -112,10 +116,36 @@ async def fetch_detections(
...
+    cameras_list = await cameras.fetch_all(filter_pair=("organization_id", token_payload.organization_id))
     camera_ids = [camera.id for camera in cameras_list]
 
-    return await detections.get_in(camera_ids, "camera_id")
+    return await detections.fetch_all(in_pair=("camera_id", camera_ids))
+
+
+@router.get("/unacknowledged/from", status_code=status.HTTP_200_OK, summary="Fetch all the unacknowledged detections")
+async def fetch_unacknowledged_detections(
+    from_date: Optional[datetime] = None,
```
frgfm, line 127: Path, payload or Form data, but I'd suggest JSON Payload to get proper pydantic validation & for the sake of uniformity across the routes

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -112,10 +116,36 @@ async def fetch_detections(
...
+    detections: DetectionCRUD = Depends(get_detection_crud),
+    cameras: CameraCRUD = Depends(get_camera_crud),
+    token_payload: TokenPayload = Security(get_jwt, scopes=[UserRole.ADMIN, UserRole.AGENT, UserRole.USER]),
+) -> List[Detection]:
+    telemetry_client.capture(token_payload.sub, event="unacknowledged-fetch")
+
+    all_unck_detections = await detections.fetch_all(filter_pair=("is_wildfire", None))
+    if from_date is not None:
+        all_unck_detections = [detection for detection in all_unck_detections if detection.created_at >= from_date]
```
frgfm, line 136: Better to extend again the fetch_all to have this filter in SQL directly

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -112,10 +116,36 @@ async def fetch_detections(
...
+
+    if UserRole.ADMIN in token_payload.scopes:
+        return all_unck_detections
+
+    cameras_list = await cameras.fetch_all(filter_pair=("organization_id", token_payload.organization_id))
+    camera_ids = [camera.id for camera in cameras_list]
+    all_unck_detections = await detections.fetch_all(
+        filter_pair=("is_wildfire", None), in_pair=("camera_id", camera_ids)
+    )
```
frgfm, line 145: we fetch detections twice if we're not an admin here (cf. line 134)

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -112,10 +116,36 @@ async def fetch_detections(
...
+    if from_date is not None:
+        all_unck_detections = [detection for detection in all_unck_detections if detection.created_at >= from_date]
```
frgfm, line 147: same here

frgfm (commented):
Thanks! I added a few comments!

src/app/api/api_v1/endpoints/detections.py
RonanMorgan: hum Form isn't  equal to Payload ? I see that we are already using Form but I don't see any Payload

src/app/api/api_v1/endpoints/detections.py
RonanMorgan: fixed in my last PR

src/app/api/api_v1/endpoints/detections.py
RonanMorgan: I will add it in my last PR (there is refactor of this endpoint there)

src/app/api/api_v1/endpoints/detections.py
frgfm: I mean we define a Pydantic inherited class with one datetime attribute called "starting_date" or "from_date". So that we'll get payload validation.

Form is what we use here for user login for instance

src/app/api/api_v1/endpoints/detections.py
frgfm: Alright, but please next time, 1 PR = 1 focus. If we put fixes of PR in other PRs, it's tricky to review & we lose what we've tried to have so far: every single commit on "main" works (that's why we squash commit of a PR)

MateoLostanlen:
hello @frgfm ,

We only display alerts for the current day (i.e. less than 24 hours).

Then, after discussion with the firefighters, I've come up with the idea of an archive tab for viewing past alerts

All alerts from yesterday or before and all acknowledge alerts will be present in the archive.

frgfm:
Would you mind resolving the conflicts please? The review tab mixes several PR and things already merged on main :sweat_smile:

RonanMorgan:
done :)

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -112,10 +112,33 @@ async def fetch_detections(
...
+async def fetch_unlabeled_detections(
+    from_date: datetime = Query(),
+    detections: DetectionCRUD = Depends(get_detection_crud),
+    cameras: CameraCRUD = Depends(get_camera_crud),
+    token_payload: TokenPayload = Security(get_jwt, scopes=[UserRole.ADMIN, UserRole.AGENT, UserRole.USER]),
+) -> List[Detection]:
+    telemetry_client.capture(token_payload.sub, event="unlabeled-fetch")
+
+    all_unck_detections = await detections.fetch_all(filter_pair=("is_wildfire", None))
+    all_unck_detections = [detection for detection in all_unck_detections if detection.created_at >= from_date]
```
frgfm, line 131: Can't we adapt again `fetch_all` so that this is done in SQL?
We'd save a lot of RAM & potentially a bit of latency

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -112,10 +112,33 @@ async def fetch_detections(
...
+
+    if UserRole.ADMIN in token_payload.scopes:
+        return all_unck_detections
+
+    cameras_list = await cameras.fetch_all(filter_pair=("organization_id", token_payload.organization_id))
+    camera_ids = [camera.id for camera in cameras_list]
+    all_unck_detections = await detections.fetch_all(
+        filter_pair=("is_wildfire", None), in_pair=("camera_id", camera_ids)
+    )
```
frgfm, line 140: if you're not an admin, fetch_all will be called twice while we only need it once

frgfm (commented):
Thanks! Some duplicate SQL queries, and a potential optim (have the date filter move to SQL operation). But it looks like we're almost there!

src/app/api/api_v1/endpoints/detections.py
RonanMorgan: done

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -112,10 +124,46 @@ async def fetch_detections(
...
+        )
+        cameras_list = await cameras.fetch_all()
+        dict_camera_orgid = {}
+        for camera in cameras_list:
+            dict_camera_orgid[camera.id] = camera.organization_id
+        url_list = [
+            DetectionUrl(url=await s3_bucket.get_public_url(detection.bucket_key))
+            for detection in all_unck_detections_admin
+        ]
+        return (all_unck_detections_admin, url_list)
```
frgfm, line 154: cf earlier, let's stick to list of dict

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -112,10 +124,46 @@ async def fetch_detections(
...
+    camera_ids = [camera.id for camera in cameras_list]
+    all_unck_detections = await detections.fetch_all(
+        filter_pair=("is_wildfire", None),
+        in_pair=("camera_id", camera_ids),
+        inequality_pair=("created_at", ">=", from_date),
+    )
+    url_list = [
+        DetectionUrl(url=await s3_bucket.get_public_url(detection.bucket_key)) for detection in all_unck_detections
+    ]
+    return (all_unck_detections, url_list)
```
frgfm, line 166: same here

src/app/models.py
```diff
@@ -53,6 +53,7 @@ class Detection(SQLModel, table=True):
     camera_id: int = Field(..., foreign_key="camera.id", nullable=False)
     azimuth: float = Field(..., gt=0, lt=360)
     bucket_key: str
+    localization: Union[str, None]
```
frgfm, line 56: cf. earlier discussion, I think we need a better name for that: keypoints or bboxes are much clearer (localization could be the actual geographic location of the wildfire or the camera)

frgfm (commented):
Thanks, I added a few comments!

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -24,12 +26,19 @@
 
 @router.post("/", status_code=status.HTTP_201_CREATED, summary="Register a new wildfire detection")
 async def create_detection(
+    localization: Union[str, None] = Form(None),
```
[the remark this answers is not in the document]
RonanMorgan: are you sure we can ? I don't find anything in the Form documentation and Chatgpt suggest something like that : 

`    localization: Union[constr(regex=r"^\[\[\d+\.\d+,\d+\.\d+,\d+\.\d+,\d+\.\d+,\d+\.\d+\]\]$"), None]`

but I did not succeed to make it work yet

src/app/models.py
RonanMorgan: @MateoLostanlen

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -112,10 +124,46 @@ async def fetch_detections(
...
+            filter_pair=("is_wildfire", None), inequality_pair=("created_at", ">=", from_date)
+        )
+        cameras_list = await cameras.fetch_all()
+        dict_camera_orgid = {}
+        for camera in cameras_list:
+            dict_camera_orgid[camera.id] = camera.organization_id
+        url_list = [
+            DetectionUrl(url=await s3_bucket.get_public_url(detection.bucket_key))
+            for detection in all_unck_detections_admin
+        ]
```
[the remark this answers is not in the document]
RonanMorgan: done

src/app/api/api_v1/endpoints/detections.py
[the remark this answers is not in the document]
frgfm: Yup, if you install fastapi, the lsp completion in your IDE should show the "regex" arg in the constructor 🙂 (I tried on Friday)

frgfm:
cf. #304

PR #342, opened 2024-07-09
Add localization in Detection table

RonanMorgan:
Add the localization string allowing the engine to register the bbox detected during the inference


src/tests/conftest.py
```diff
@@ -90,6 +90,7 @@
         "camera_id": 1,
         "azimuth": 43.7,
         "bucket_key": "my_file",
+        "localization": "xyxy",
```
frgfm, line 93: Mmmh, can we get a regex validation here?
Like format of each coord with X decimals, and comma-separated (or something else).

In fact, I'd like the DB initialization to fail if we put "xyxy" as a value.

frgfm (commented):
Thanks! In models & schemas, can we push a bit further the "localization" string validation? Pydantic has regex, and we can't afford to allow "xyxy" in the database

src/tests/conftest.py
RonanMorgan: I can do it in the endpoint directly to avoid saving bad data

PR #343, opened 2024-07-11
Fix :the admin behavior for bucket requests

RonanMorgan:
If you are connected as an admin, the bucket par defaut is "admin" but the endpoint Detection was using the "alert-api-1" name.

Anyway we should not use any of this two solutions when requesting for an URL, we need to use the Detection_id instead


frgfm:
I'm a bit lost here: would it be possible to directly put those commits in the bucket/org PR?

PR #349, opened 2024-07-17
Old datamodel : add url into fetch_unacknowledged events

RonanMorgan:
=> Send back media_url with Events in order to be more efficient in the platform side

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -17,8 +17,8 @@ services:
       - SQLALCHEMY_SILENCE_UBER_WARNING=1
       - SUPERUSER_LOGIN=dummy_login
       - SUPERUSER_PWD=dummy&P@ssw0rd!
-      - BUCKET_NAME=sample-bucket
-      - S3_ENDPOINT_URL=http://s3.localhost.localstack.cloud:4566
+      - BUCKET_NAME=bucket
+      - S3_ENDPOINT_URL=http://localstack:4566
       - S3_ACCESS_KEY=fake
       - S3_SECRET_KEY=fake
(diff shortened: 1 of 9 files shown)


docker-compose.test.yml
```diff
@@ -51,8 +51,10 @@ services:
     environment:
       - EDGE_PORT=4566
       - SERVICES=s3
+    volumes:
+      - ./scripts/localstack:/etc/localstack/init/ready.d
```
frgfm, line 55: For speed & RAM, might be worth removing this and copying it in the actual docker image instead.

src/tests/routes/test_events.py
```diff
@@ -522,7 +522,7 @@ async def test_fetch_unacknowledged_events(test_app_asyncio, init_test_db, acces
             event_ids = [alert["event_id"] for alert in ALERT_TABLE if alert["device_id"] in devices_group_id]
             events_group_id = [event["id"] for event in EVENT_TABLE if event["id"] in event_ids]
 
-        assert response.json() == [
+        assert response.json()[0] == [
```
frgfm, line 525: I don't get how that test can pass? Should this change be reverted?

frgfm (approved):
Nothing major blocking, so I'm happy to merge! I'll only warn about one thing: the tests haven't been run by the CI so make sure to run all tests locally + the end to end script for integration :)

src/tests/routes/test_events.py
RonanMorgan: this is because I am now sending List(Events) + List(URL) in a tuple but I cannot predict the  URL so I am just testing the Events parts

PR #364, opened 2024-08-30
docs(contributing): update the DB UML

frgfm:
This PR simply updates the UML of the database by adding the bboxes introduced in #340


frgfm:
The CI failure is unrelated (DNS config change at the moment of the test, but this is restored now)

PR #366, opened 2024-09-10
Fix fetch unlabeled detections endpoint

RonanMorgan:
When calling the fetch unlabeled detections endpoint with an ADMIN token, the pictures were not found in the S3 because the organisation_id of ADMIN is = 1, while we would like to get all the Detection of all the organization.

So for ADMIN token we should find the organization_id thanks to the Camera.detection_id

### DIFF
--- a/src/app/api/api_v1/endpoints/detections.py
+++ b/src/app/api/api_v1/endpoints/detections.py
@@ -19,10 +19,13 @@ from fastapi import (
     UploadFile,
     status,
 )
+from sqlmodel import select
+from sqlmodel.ext.asyncio.session import AsyncSession
 
 from app.api.dependencies import dispatch_webhook, get_camera_crud, get_detection_crud, get_jwt, get_webhook_crud
 from app.core.config import settings
 from app.crud import CameraCRUD, DetectionCRUD, WebhookCRUD
+from app.db import get_session
 from app.models import Camera, Detection, Role, UserRole
 from app.schemas.detections import (
     BOXES_PATTERN,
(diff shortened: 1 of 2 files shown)


src/app/api/api_v1/endpoints/detections.py
```diff
@@ -165,19 +165,24 @@ async def fetch_unlabeled_detections(
     def get_url(detection: Detection) -> str:
         return bucket.get_public_url(detection.bucket_key)
 
+    async def get_url_with_bucket(detection: Detection) -> str:
+        camera = cast(Camera, await cameras.get(detection.camera_id, strict=True))
```
frgfm, line 169: Considering the size of the potential result, this could become quite slow. We've had to solve this on the old-production branch. I suggest taking the same approach : https://github.com/pyronear/pyro-api/blob/old-production/src/app/api/endpoints/events.py#L149

(otherwise if there are 150 detections, we'll have to do 150 invidual camera fetch)

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -64,7 +64,7 @@ async def create_detection(
     # guess_extension will return none if this fails
     extension = guess_extension(magic.from_buffer(file.file.read(), mime=True)) or ""
     # Concatenate timestamp & hash
-    bucket_key = f"{token_payload.sub}-{datetime.utcnow().strftime('%Y%m%d%H%M%S')}-{sha_hash[:8]}{extension}"
+    bucket_key = f"{datetime.utcnow().strftime('%Y%m%d%H%M%S')}-{sha_hash[:8]}{extension}"
```
frgfm, line 67: Why the change? This enabled us to easily find pictures by device in the bucket

RonanMorgan: sorry it's a mistake

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -165,19 +165,24 @@ async def fetch_unlabeled_detections(
...
+    async def get_url_with_bucket(detection: Detection) -> str:
+        camera = cast(Camera, await cameras.get(detection.camera_id, strict=True))
+        bucket_ = s3_service.get_bucket(s3_service.resolve_bucket_name(camera.organization_id))
+        return bucket_.get_public_url(detection.bucket_key)
+
     if UserRole.ADMIN in token_payload.scopes:
         all_unck_detections = await detections.fetch_all(
             filter_pair=("is_wildfire", None), inequality_pair=("created_at", ">=", from_date)
         )
+        urls = [await get_url_with_bucket(detection) for detection in all_unck_detections]
```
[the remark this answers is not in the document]
RonanMorgan: hum I have lots of issue with the "quality" 
If I keep 
src/app/api/api_v1/endpoints/detections.py:191: error: Incompatible types in assignment (expression has type
"Generator[str, None, None]", variable has type "AsyncGenerator[str, None]")  [assignment]
            urls = (get_url(detection) for detection in all_unck_detections)
and 
src/app/api/api_v1/endpoints/detections.py:193: error: No overload variant of "zip" matches argument types "list[Detection]",
"AsyncGenerator[str, None]"  [call-overload]

It's simpler to keep List comprehension no ?

src/app/api/api_v1/endpoints/detections.py
RonanMorgan: ok I have implemented a pagination parameter in the endpoint (and the fetch_all function) is it ok for you ?

frgfm: On paper yes, but what is implemented on old-production was proofed with platform interactions and extremely fast. Pagination will for sure needed at some point, but for now, for the sake of "let's do what we know works well", I'd suggest taking only adapting that code. If you feel like we should introduce pagination now:
- add a param to change the num of events (https://github.com/pyronear/pyro-api/blob/old-production/src/app/api/endpoints/events.py#L165), but enforce that it cannot be bigger than 15 for instance
- add another param to change the num of consecutive alerts per event for the GIF (https://github.com/pyronear/pyro-api/blob/old-production/src/app/api/endpoints/events.py#L221), but enforce that it cannot be bigger than 10 for instance

Considering this is only for the platform and for now, it fills that need, I think those param and pagination aren't necessary but I'll let you be the judge of that :ok_hand:

src/app/api/api_v1/endpoints/detections.py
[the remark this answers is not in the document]
frgfm: Sure :+1: but I don't get why the traceback mentions a `zip` while there is none

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -64,8 +64,7 @@ async def create_detection(
     # guess_extension will return none if this fails
     extension = guess_extension(magic.from_buffer(file.file.read(), mime=True)) or ""
     # Concatenate timestamp & hash
-    bucket_key = f"{token_payload.sub}-{datetime.utcnow().strftime('%Y%m%d%H%M%S')}-{sha_hash[:8]}{extension}"
-    # Reset byte position of the file (cf. https://fastapi.tiangolo.com/tutorial/request-files/#uploadfile)
+    bucket_key = f"{token_payload.sub}-{datetime.utcnow().strftime('%Y%m%d%H%M%S')}-{sha_hash[:8]}{extension}"  # Reset byte position of the file (cf. https://fastapi.tiangolo.com/tutorial/request-files/#uploadfile)
```
frgfm, line 67: Can you revert the comment position to the next line as previously? (gonna have formatting errors at some point otherwise)

frgfm (commented):
Thanks! I'm pretty sure everything here works but as mentioned in the comment, I have no clue how fast. This route needs to be really low latency and we've iterated on old-production branch to make something that works. Out of safety, I think we should keep that design :ok_hand:

src/app/api/api_v1/endpoints/detections.py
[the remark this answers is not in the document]
RonanMorgan: L193 there is a zip :
`    return [DetectionWithUrl(**detection.model_dump(), url=url) for detection, url in zip(all_unck_detections, urls)]`

PR #368, opened 2024-09-13
Psycopg dependency error

RonanMorgan:
This is fixing the following error : 

> ModuleNotFoundError: No module named 'psycopg2'

It seems that the async driver asyncpg still has a dependency to the psycopg2 driver (for errors definition)

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1045,6 +1045,87 @@ fastapi = ">=0.38.1,<1.0.0"
 prometheus-client = ">=0.8.0,<1.0.0"
 
 [[package]]
+name = "psycopg2-binary"
+version = "2.9.9"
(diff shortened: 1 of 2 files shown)


frgfm:
Thanks for the PR! Seems like the review will be straight forward: can you specify how to reproduce the error though?
I haven't met that error neither the CI, so either we need to update the CI to reproduce that or we don't have the same environment :sweat_smile:

RonanMorgan:
hum ok I think this happened because this part "+asyncpg" was not in the database URL in the production docker-compose so it should be ok now

frgfm:
> hum ok I think this happened because this part "+asyncpg" was not in the database URL in the production docker-compose I found so it should be ok now

Ah gotcha, I'll close the PR then!

PR #371, opened 2024-09-22
feat(cameras): add a route to update the last known image of a camera

frgfm:
- add a "last_image" column in the Camera table
- every time a camera uses it, it uploads an image, stores it in the bucket (and delete the previous one) + update the DB entry
- update the client & corresponding tests
- update the UML & made it live

Here is the updated UML
pyronear_uml
(live link: https://dbdiagram.io/d/Pyronear-UML-665a15d0b65d933879357b58)


frgfm:
Worth noting: @MateoLostanlen we'll have to update the engine. We now have 2 potential high frequency interactions: heartbeat & update image. I suggest reversing the equations to set the adequate frequency (& decorrelate it from the prediction frequency on raspberry pi):
- last image: I'd argue we don't need more often than every 10mins
- heartbeat: this is our way to detect if we have "downtime" on a station. Probably every minute ?

What do you think?

PR #374, opened 2024-09-24
Add field "manual bboxes" in the Detection table

RonanMorgan:
We need this field in order to be able to differentiate the bboxes detected by the algorithm and the bboxes manually annoted by a human


frgfm:
As discussed with @MateoLostanlen , this now falls on the labelling api side, not the alert api. Feel free to reopen if I missed something 👍

PR #382, opened 2024-10-29
Add a prefix to bucket name

RonanMorgan:
Currently we have a conflict between bucket name because in our OVH account we have many environment (production + dev *2). This small refactor prevent this conflicts

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -21,6 +21,7 @@ SERVER_NAME=
 POSTHOG_HOST='https://eu.posthog.com'
 POSTHOG_KEY=
 SUPPORT_EMAIL=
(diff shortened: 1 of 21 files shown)


frgfm (commented):
Thanks! Agree with the fix, but for the sake of limiting the number of env vars, I think we should mutualize this with `SERVER_NAME` and probably renaming it to `INSTANCE_NAME` (or by default using `socket.gethostname()` or something unique to the VM so that we don't have to figure out which prefix isn't being used)

RonanMorgan:
ok ! I just merge with main and push the modification

frgfm (approved):
thanks!

PR #383, opened 2024-10-29
Fix redirect error when calling cameras/

RonanMorgan:
Without  this "/" there was a redirection to the port 80 (which was causing an error since it's not proxied)

According to this : https://github.com/fastapi/fastapi/discussions/9328
There is surely a proper solution if the HTTP headers are set up correctly, but I didn't succeed


frgfm (approved):
thanks for the fix!

PR #399, opened 2025-01-07
fix(detections): allow azimuth = 0

### LINKED ISSUE #395: Azimuth can be zero
Hello guys, small change needed I had an issue with 

```
{"detail":[{"type":"greater_than","loc":["body","azimuth"],"msg":"Input should be greater than 0","input":"0","ctx":{"gt":0.0}}]}
{'detail': [{'type': 'greater_than', 'loc': ['body', 'azimuth'], 'msg': 'Input should be greater than 0', 'input': '0', 'ctx': {'gt': 0.0}}]}
```

But an azimuth can be zero, it must be in [0,360]

RonanMorgan:
Currently there is an error if the azimuth is defined equal to 0

Close #395


MateoLostanlen (approved):
Hi @RonanMorgan , thanks a lot for the fix

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -60,7 +60,7 @@ async def create_detection(
         min_length=2,
         max_length=settings.MAX_BBOX_STR_LENGTH,
     ),
-    azimuth: float = Form(..., gt=0, lt=360, description="angle between north and direction in degrees"),
+    azimuth: float = Form(..., gt=-1, lt=360, description="angle between north and direction in degrees"),
```
frgfm, line 63: Actually this will allow ]-1,360[ which could be problematic. I think you're looking for:

```suggestion
    azimuth: float = Form(..., ge=0, lt=360, description="angle between north and direction in degrees"),
```

src/app/models.py
```diff
@@ -57,7 +57,7 @@ class Detection(SQLModel, table=True):
     __tablename__ = "detections"
     id: int = Field(None, primary_key=True)
     camera_id: int = Field(..., foreign_key="cameras.id", nullable=False)
-    azimuth: float = Field(..., gt=0, lt=360)
+    azimuth: float = Field(..., gt=-1, lt=360)
```
frgfm, line 60: same here

src/app/schemas/detections.py
```diff
@@ -20,7 +20,7 @@ class DetectionLabel(BaseModel):
 class Azimuth(BaseModel):
     azimuth: float = Field(
         ...,
-        gt=0,
+        gt=-1,
```
frgfm, line 23: same

src/migrations/versions/2024_05_30_1200-f84a0ed81bdc_init.py
```diff
@@ -22,7 +22,7 @@
 def upgrade() -> None:
     # ### commands auto generated by Alembic - please adjust! ###
     op.create_table(
-        "camera",
+        "cameras",
```
frgfm, line 25: reminder that we'll need to test extensively the migration script or clean it at some point 😅

frgfm (changes requested):
Thanks for the PR! I added a suggestion because if we merged this modification we'll get ]-1,360[ (floats, not integers), but we want [0,360[

frgfm (approved):
Thanks! Only one concern: I can't figure out why the docker build job fails 🤔 (it's working on main)
Worth checking before merging

frgfm:
Close #395

RonanMorgan:
@frgfm I don't have any issue locally but it seems to me that the backend stay unhealthy longer than 30sec, can I change this parameters ? :

      interval: 10s
      timeout: 3s
      retries: 3

frgfm:
@RonanMorgan yes but if you do, only increase the number of retries please 🙏

RonanMorgan:
ok I just remove the docker-compose;yml file and replaced it by docker-compose.dev;yml (which make sense now since there is a production docker-compose in ansible)

I think it was a variable error since there are default values in this one

frgfm (commented):
Changing the docker env var values is quite alright and removing the somehow duplicate compose, but not too sure about removing alembic and volume binding for the DB & localstack?

This would make the docker compose solely a "test" compose, because there is no data persistence and we don't create migrations. is this a good idea? 🤔  (it fixes the CI, but my question is: won't this reduce the usability of the project?)

RonanMorgan:
well is this used for anything else than test  ? :) 

I have an other docker-compose for any production usage in ansible, I think this should be enough (?)

frgfm:
> I have an other docker-compose for any production usage in ansible, I think this should be enough (?)

Question of perspective, I'm open to both. Until now, we've made sure the repo had something for the dev to "try locally" what production would do, and run tests as well. You think it's not worth it?

RonanMorgan:
I think that's worth it but in this case we need to add Treaefik and test it regularly in a new vm. For me the docker-compose I removed was too far from the production one to be more useful the the .dev. one

frgfm:
Ok, let's first finish this PR topic: why in the first place did we need to change so many files to have the CI work 🤔 
Especially because the CI has had no problem on main for a while. For last few PRs, we didnt have to change anything on docker or the CI config to get it to pass. If we could have that minimal change, that would awesome 👌

frgfm:
I tried my best to rebase the PR but several conflicts were a bit tideous so I created #414 
Closing this one since this is now handled, thanks again!

PR #412, opened 2025-01-16
feat(sequences

frgfm:
This PR follows up on #405, and implements the final steps for the interactions with detections and sequences.

### DIFF
--- a/scripts/test_e2e.py
+++ b/scripts/test_e2e.py
@@ -5,6 +5,7 @@
 
 import argparse
 import time
+from datetime import datetime
 from typing import Any, Dict, Optional
 
 import requests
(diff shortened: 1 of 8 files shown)


MateoLostanlen:
Thanks for the PR @frgfm 

As discussed together I'm ok with the strategy

The code logic looks good to me, I don't know SQL so it's hard to say more without testing :) 

You're missing the route for acknowledge, aren't you?

Good point about teegram, can you open an issue maybe?

I'll let you see about the test that fails

PS: You should remove the test required in 3.9

frgfm:
> You're missing the route for acknowledge, aren't you?

The point was that this would be `PATCH sequences/SEQ_ID/label`, which sets the value for "is_wildfire". I thought this was already agreed but let me know!

> Good point about telegram, can you open an issue maybe?

done

MateoLostanlen:
Ok my bad I missed that, event better I will make the change on the platform !

MateoLostanlen (approved):

PR #458, opened 2025-04-17
feat(sequences): add detection cone in sequence fetching routes

### LINKED ISSUE #432: Computer detection cone for triangulation
Hello API team!

The firefighters have requested a triangulation feature, so we are currently developing it on the platform side. The principle is quite simple: if two stations detect the same fire, we can estimate the location by calculating the overlap of their detection cones.

As a reminder, on the platform, the detection cones are computed in [the following function](https://github.com/pyronear/pyro-platform/blob/0723a9cd7692d3c8f141eebef39b50fedc955f52/app/utils/display.py#L38). Using the camera's field of view, its central azimuth, and the detection bounding box, we can infer the detection center azimuth and its angular opening. 

Until now, calculating this on the platform side wasn’t an issue because we only displayed the cone for a single sequence—the one shown on the screen—for which we had already retrieved all the necessary information.

Now, to perform triangulation, we need to find matching sequences, which requires detection cone information for each sequence. If I do this on the platform side, I would have to load the detections for each sequence to extract a bounding box and compute these parameters, which is inefficient.

I propose moving this calculation to the API side. When a new sequence is created, we simply take the first non-null bounding box and compute the detection cone parameters. I suggest adding a new column, detection_cone_parameter, to the sequence table, storing this value as a tuple (detection_azimuth, detection_opening_angle).

Is this okay for you?

frgfm:
This PR:
- updates GET /sequences/unlabeled/latest & GET /all/fromdate to return detection cone in the response
- the detection cone is azimuth of the centre of the highest confidence bbox + angle in the first detection
- add test case

Close #432


src/app/api/api_v1/endpoints/sequences.py
```diff
@@ -31,6 +32,37 @@ async def verify_org_rights(
         raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail="Access forbidden.")
 
 
+def _resolve_cone(azimuth: float, bboxes_str: str, aov: float) -> Tuple[float, float]:
```
MateoLostanlen, line 35: Hey @frgfm ,
As we discussed, selecting the bbox with the highest score is problematic in the case of fires that last a bit longer. As the fire grows, the bbox gets bigger, which reduces cone's precision 

Two options:

1. The best one in my opinion: we take the highest score, but only among the first 4 (the length of the sequence analyzed by the engine)

2. We simply take the first one

frgfm: This is taking the highest confidence bbox on the first detection of the sequence, not the highest conf bbox in the entire sequence :) Isn't that what we wanted?

MateoLostanlen: oh yes my bad I read it a bit too quickly 😄

MateoLostanlen (approved):

PR #468, opened 2025-05-15
Add slack notification

RonanMorgan:
Give the possibility to activate slack notification into a channel (one channel per organisation) 

=> Add an endpoint in the organizations/ path in order to activate the slack notification by registering the SLACK_HOOK
=> Create a SLACK service, similar to the Telegram service
=> modify the create_detection endpoint in on order to add the notification creation : the bbox is added to the images of the Sequence, a gif is created from the images of the Sequence, the gif is saved into the S3 bucket and the url is sent to slack.


src/app/services/slack.py
```diff
@@ -0,0 +1,52 @@
...
+
+logger = logging.getLogger("uvicorn.error")
+
+__all__ = ["slack_client"]
+
+
+class SlackClient:
+    def __init__(self) -> None:
+        self.is_enabled = True
+        # Do we want a config in settings ?
```
RonanMorgan, line 18: check

MateoLostanlen:
Hey, thanks for the feedback!

I’m not sure I fully understand the issue here. Sure, OpenCV isn’t the lightest dependency, but it’s not exactly heavy either. Unless I’m missing something, we’re not particularly resource-constrained, are we? Do you have a concrete case in mind where this could be a problem?

More broadly, I get the idea of keeping the API minimal, but I think we might be underestimating how valuable it is to quickly visualize an alert. Seeing a fire without a bounding box takes time. And since we’re doing early detection, the fire is often small and not obvious. A still image with a bbox is better, but a GIF is even quicker to interpret. We’re racing against time here

If including OpenCV in the main container is really an issue, could we consider a second container dedicated to the computer vision part, just generating the GIF and returning a URL to the API? That would keep things modular while still providing the best user experience.

Let me know what you think!

frgfm:
Hey there :)

Yup fully get your point, here is some short feedback:
- we're not exactly resource constrained: when I say light I mean in terms of dependencies (additional licenses but also versioning and bugs that comes with it). I've had that type of experience with some dependencies we have here in another project. I'm only mindful of those when I see a way to avoid them without degrading the output too much
- to my previous point, the slack API is generous enough so that image URL are rendered. This avoids downloading bytes from s3 and "uploading them" to slack, it will only be "downloaded" by the end user when Slack renders --> much faster and lower load. Now you're correct that doing so, we can't modify the image (can't add bbox or transform into GIF) but I don't want the Alert API to take that responsibility, it's a slippery road to stretch that much
- your suggestion is on point especially since I wanted to create a small vision API for pyronear

For now my suggestion:
- first only forwarding the URL (no bbox, no GIF). Remember it's for our Slack, not firefighters
- Otherwise I'd suggest including a URL to the platform: when we click it, it opens the platform on this specific alert. That way no one is taking extra load and everything stays lean :)

What do you think?

MateoLostanlen:
Just to clarify: this is actually meant for firefighters on our Slack, the idea is for them to get alerts in real time on their phone or computer

You're the API king haha, so I’ll let you make the call. I trust you, if you think it’s not the right direction long-term, then let's not go for it. Still, that GIF was pretty sexy! I guess we’ll need to kick off that vision API sooner rather than later. I really think there are quite a few interesting use cases we could develop around it

@RonanMorgan sorry about that… I should’ve kicked off the conversation earlier. I made you work for nothing. Would you mind updating the PR to go back to a message with alert info and a URL instead?

frgfm:
Superseeded by #478

PR #475, opened 2025-06-06
feat(cameras): add last image URLs when fetching all cameras

frgfm:
This PR:
- adds "last_image_url" in the response of `GET /api/v1/cameras`
- update the unit tests and the end to end test


MateoLostanlen:
Thanks for clarifying :)

PR #476, opened 2025-06-06
refactor(client): pass down query param from client

frgfm:
This PR solves #450 by passing down query params from client method


MateoLostanlen (approved):
Perfect thanks for this PR !

PR #478, opened 2025-06-11
Rs/add simple slack notification

RonanMorgan:
Give the possibility to activate slack notification into a channel (one channel per organisation)

=> Add an endpoint in the organizations/ path in order to activate the slack notification by registering the SLACK_HOOK
=> Create a SLACK service, similar to the Telegram service
=> modify the create_detection endpoint in on order to add the notification creation


frgfm:
I might be mistaking but it looks like you forgot to remove cv2 and/or pillow + the extra dependencies from Docker

src/Dockerfile
```diff
@@ -10,7 +10,6 @@ ENV PYTHONPATH="/app"
 # Install curl
 RUN apt-get -y update \
     && apt-get -y install curl libmagic1 \
-    && apt-get clean \
```
frgfm, line 13: Can you please revert this please?

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -134,6 +139,18 @@ async def create_detection(
                 org = cast(Organization, await organizations.get(token_payload.organization_id, strict=True))
                 if org.telegram_id:
                     background_tasks.add_task(telegram_client.notify, org.telegram_id, det.model_dump_json())
+
+            if slack_client.is_enabled:
+                org = cast(Organization, await organizations.get(token_payload.organization_id, strict=True))
```
frgfm, line 144: Let's avoid the double call if telegram and slack client are enabled

src/app/api/api_v1/endpoints/organizations.py
```diff
@@ -92,3 +93,23 @@ async def update_telegram_id(
...
+@router.patch(
+    "/slack-hook/{organization_id}", status_code=status.HTTP_200_OK, summary="Update slack token of an organization"
+)
+async def update_slack_hook(
+    payload: SlackHook,
+    organization_id: int = Path(..., gt=0),
+    organizations: OrganizationCRUD = Depends(get_organization_crud),
+    token_payload: TokenPayload = Security(get_jwt, scopes=[UserRole.ADMIN]),
+) -> Organization:
+    telemetry_client.capture(
```
frgfm, line 107: This isn't relevant for product analytics, no need to add the event capture :)

pyproject.toml
```diff
@@ -17,6 +17,7 @@ pydantic = ">=2.0.0,<3.0.0"
 pydantic-settings = ">=2.0.0,<3.0.0"
 requests = "^2.32.0"
 PyJWT = "^2.8.0"
+bcrypt = "<4.0.0"
```
frgfm, line 20: Can you elaborate on this one? (it's already in the passlib extra) did you encounter a bug with the lockfile?

docker-compose.dev.yml
```diff
@@ -55,6 +55,7 @@ services:
       - S3_ACCESS_KEY=fake
       - S3_SECRET_KEY=fake
       - S3_REGION=us-east-1
+      - SLACK_HOOK=${SLACK_HOOK}
```
frgfm, line 58: if I understand this correctly, this is the admin org slack hook?
If so I think it shouldn't be initialized in the DB initialization script + validated in the config
If not, let's remove it?

poetry.lock
frgfm, line 1: boto3 was already on the version you specified, why is the entire lockfile updated? can you please revert? Check my comment in pyproject but if bcrypt is indeed not needed, we don't need to change the lockfile

(please don't bump versions without a reason)

src/app/api/api_v1/endpoints/detections.py
RonanMorgan: hum I don't understand, what is the issue here ?

docker-compose.dev.yml
RonanMorgan: this is used only for testing purpose

src/app/api/api_v1/endpoints/detections.py
RonanMorgan: ok sorry I wrongly understand the first time

frgfm (approved):
Thanks! Small warning from mypy but let's move on

PR #487, opened 2025-07-16
small refactor of the slack notification

RonanMorgan:


MateoLostanlen (approved):
thanks for this update requested by firefighters !

MateoLostanlen:
@RonanMorgan what do you mean by "Should we send the S3 URL for a better viewing in mobile ?" it's already what we do no ?

PR #497, opened 2025-08-20
Modify is wildfire column

RonanMorgan:
1) create an ENUM which will allow to annotate with more precision
2) fix the migration files which were deprecated (Sequeces table inexistant, webhoo table inexistant, slack_notification ect ect)


src/app/models.py
```diff
@@ -27,6 +27,29 @@ class Role(str, Enum):
     USER = "user"
 
 
+class AnnotationType(str, Enum):
```
Chouffe, line 30: Do we want to have different types of smoke: wildfire, industrial, chimney, agriculture, etc or also all the FP types?

src/app/schemas/sequences.py
```diff
@@ -18,7 +18,7 @@ class SequenceUpdate(BaseModel):
 
 
 class SequenceLabel(BaseModel):
-    is_wildfire: bool
+    is_wildfire: AnnotationType
```
Chouffe, line 21: So this will automatically do the validation at the API level of the string that is passed in?

src/migrations/versions/2025_08_20_1647-307a1d6d490d_modify_is_wilfire_column.py
```diff
@@ -0,0 +1,75 @@
...
+    # Create the enum type in the database
+    annotation_type_enum.create(op.get_bind(), checkfirst=True)
+
+    # Use raw SQL with a CASE expression for the conversion
+    op.execute("""
+        ALTER TABLE sequences
+        ALTER COLUMN is_wildfire
+        TYPE annotationtype
+        USING CASE
+            WHEN is_wildfire = TRUE THEN 'WILDFIRE_SMOKE'::annotationtype
```
Chouffe, line 56: Nice! So we migrate all the true values to wildfire smoke by default.

Chouffe:
So this will break the current pyro annotator. Also the platform is likely to break (alertapi). What is the migration path for these? 

An option that I usually take is to never do backward incompatible changes in the DB and give time to the client and other dependencies to adapt to the new db model. In that case, I would probably create an entirely new field (eg. annotation which is the enum you created) and expose this new data from the API, and mark the is_wildfire_alert as deprecated. Then the platform and annotator can adapt by using the new field. When this is done, we can always drop the is_wildfire_alert boolean db field from the pyro api DB model or even leave it like so, we just remove the API params that can CRUD this field.

Let me know what you think.

src/app/schemas/sequences.py
RonanMorgan: yes :)

client/pyroclient/client.py
```diff
@@ -218,12 +218,12 @@ def fetch_detections(self) -> Response:
             timeout=self.timeout,
         )
 
-    def label_sequence(self, sequence_id: int, is_wildfire: bool) -> Response:
+    def label_sequence(self, sequence_id: int, is_wildfire: str) -> Response:
```
Chouffe, line 221: Should we also include the enum in the client? So that it is not a random string here?

fe51:
Hi there,

I understand the need to avoid a breaking change and to maintain consistency with the annotation API.

However, from the users' point of view, we don't want to impose such an extensive annotation on them at this stage. It seems to me that the request for change comes from the fact that on the user/platform side, we want them to distinguish between three categories: fire outbreaks / other smoke / FP 
-> Indeed, today, smoke that is not a fire outbreak is sometimes labelled as FP, sometimes as a fire outbreak, which means that everything has to be reworked afterwards !

Consequently, switching to a long list of possible categories does not solve this problem.

What do you think?
Would it be relevant to have multiple levels of categorisation ?

By switching to an enum, 1 fire, 0 FP and 2 other types of smoke -> we not not break previous DB values right ?

MateoLostanlen:
hi @fe51 the plan is to switch to fire / other smoke / fp for now, we are juste creating this enum to be prepare in case we want to do more

we need to change the type anyway we have a bool for now

@RonanMorgan do you have a migration script to convert previous annotation fire remain fire and fp witch to other  ?

src/migrations/versions/2025_08_20_1647-307a1d6d490d_modify_is_wilfire_column.py
RonanMorgan: @MateoLostanlen : migration is done here

MateoLostanlen:
dust is a complicated cas I do not have the answer

My idea was to to fp into other for now

fp and other smoke are in the enum on the other end

RonanMorgan:
There is a good point here raised by both  @fe51 @Chouffe  : the Enum could be used in many Pyronear's tools.
What I am suggesting : we accept this  PR in prder to let @Chouffe do the modification in the annotator and I do a scd PR sool (today or next week) in order to create a pyro-library (which would be in an other repo) which could be used every where (client, alert-api, annotation-api ect ect)

Chouffe (approved):
LGTM! Thanks!

frgfm (commented):
Thanks for the PR 🙏 
Late to the party but some high level comments:
- in which specific case is the boolean flag not enough and the enum is more helpful for the detection pipeline? (not the computer vision training or dataset curation)
- "is_wildfire" should arguably stay a boolean, otherwise let's name if something else since it no longer corresponds
- if this is for transition I'd recommend keeping the boolean flag and add an extra temporary column if helpful "detection_type" or something that can be an enum, but we'll still have a useful boolean flag for easy filtering

Let me know if I missed a few things!

PR #499, opened 2025-08-28
Add ip_address and livestream_activated columns in camera table

RonanMorgan:
This has been done to avoid the usage of a .json configuration file with the plateforme

@MateoLostanlen I didn't implement a new endpoint yet, maybe the get_camera endpoint would be enough ? But this means you have to manager the "rights" yourself since the endpoint is available for UserRole.ADMIN, UserRole.AGENT, UserRole.USER

1) If we consider the "ip_adress" very sensible I can : 
=>  create a new endpoint which return only the ip_adress
=> don't return the ip_address in the get_camera endpoint

2) While we are on it we should check all the information needed in the configuration.json file of the engine and add them in the camera table

what do you think ?


frgfm:
Thanks for the PR :) 
I may be late to the party again, so pardon the questions: why do we need the IP address in the DB?
We ssh via a VPN so we can't use it, right? And any HTTP request from the camera to the backend will reveal the IP

MateoLostanlen:
Hi @frgfm , sorry, a bit of context. With the arrival of livestream (remote camera control), the system now works [as follows](https://whimsical.com/architecture-heUuPL3jw3UH8UBZXF7WY)

To make this work, we need to know the IP of each Pi (on the VPN) in order to send commands
At the moment this information is stored in a JSON file on the VM, but we agree it would be cleaner to have it in the API

PR #516, opened 2025-09-25
Triangulation inside API

MateoLostanlen:


frgfm:
Thanks!
**Feature-wise**
- I agree this should be done on the backend because this is the only entity that receives all detections + has access to history.
- For this reason, I think we should run the code as a background tasks after a detection is created if some conditions are met :) (more or so, we a camera position is set, we should be able to know if its point of view can intersect any other)

**Product-wise**
I think this is also the time to consider all general cases for the near future.
Events we'll likely have:
- automatic detection from our camera with only a picture + azimuth + range
- user pinpointing a location (seeing the picture and knowing the location, they can confidently say the visible part of the fire is near GPS coordinate XYZ)
Callback/background:
- this triangulation for automatic detection
- modification of the location by the user
We'll have a very solid product + roadmap if we can handle all those

**Code-wise**: can you elaborate on the changes you made to facilitate the review? :)

Issue #528, opened 2025-11-14
Adds new table Poses & creates/updates related endpoints


fe51:
# Pitch
Following discussion in https://github.com/pyronear/pyro-api/issues/504 and on Slack, we agreed on changes to the data model to introduce new table which will enable us to develop new features that are expected.

In this case, we decided to create a “poses” table (a table containing, for line, a related camera, an azimuth at which detection are made) i.e.: one line for a fixed camera and as many lines as there are positions for PTZ cameras.

Hence relation ships between camera detection and sequences  tables will evolve

### About the table `poses`

Below some guideline about the new table (Description is here for the purpose of the issue but not sure if we include it postgresSQL tables)

Do we include a created_at field ? -> To discuss

| Column      | Type      | Constraints                           | Description                          |
|-------------|-----------|----------------------------------------|--------------------------------------|
| id          | INT       | PRIMARY KEY                            | Unique identifier                    |
| camera_id   | INT       | FOREIGN KEY → cameras(id)               | Reference to the related camera in cameras table     |
| azimuth     | FLOAT     | CHECK (azimuth BETWEEN 0 AND 359/360)      | Azimuth angle value                  |
| created_at  | TIMESTAMP | DEFAULT CURRENT_TIMESTAMP              | Record creation timestamp            |

# What has to be done

- Add every thing related to the introduction of a the new table poses in the api
- Update in tables:
     - detections table : 
          - create fk pose_id (if possible to be empty for now, until migration)
          - delete fk camera_id to pose_id -> in another PR with associated migration
          - delete azimuth field (as if it was the one related to the current pose)
     - sequences table : 
          -  create fk pose_id (if possible to be empty for now, until migration)
          - detele fk camera_id to pose_id- > in another PR with associated migration

- API ressource CAMERAS : 
     - Update GET /api/v1/cameras/ -> Fetching all cameras gives a list of camera. For each element of the list, add a "field" `poses` which list a list if azimuth, ex [{id_1: 358, id_3:45}], see full exemple at the en of the issue -> we can discuss about the exact format
     - Update GET /api/v1/cameras/{camera_id} -> For one specific camera, same logic as above 
     -  Update POST /api/v1/cameras/ to include azimuth when registering a new cam -> To discuss ?
- API ressource POSES : 
     - Create POST : /api/v1/poses/ -> create a given pose for a given camera : requested body camera_id and azimuth
[...]

PR #531, opened 2025-12-12
Introduces Poses table and related endpoints udpates

fe51:
This PR to address #528 which is part of the planned changes to the data model

It introduces the following modifications
- Add table poses in models and add the fk to the related tables (currently nullable while the migration is being carried out)
- Add poses schema and updates other related
- Add poses crud
- Add pose crud endpoints (GET, POST, PATCH, DELETE)
- Updates camera endpoints (to reflect cameras poses)
- Updates detections endpoints (to include poses) -> later, in migration, azimuth will be delete, as camera_id fk (even if camera_id can be still returned by api)
 - Update tests
 - Updates db diagram

What is missing (to be discussed of to handle in another PR)
- Alembic migrations
- Client updates -> will be done later
- Client tests (struggled to be properly run locally)

Happy to discuss it !

### DIFF
--- a/client/tests/conftest.py
+++ b/client/tests/conftest.py
@@ -40,6 +40,11 @@ def cam_token():
     response = requests.post(urljoin(API_URL, "cameras"), json=payload, headers=admin_headers, timeout=5)
     assert response.status_code == 201
     cam_id = response.json()["id"]
+    # create a pose related to the cam
+    payload = {"azimuth": 359, "patrol_id": 1, "camera_id": cam_id}
+    response = requests.post(urljoin(API_URL, "poses"), json=payload, headers=admin_headers, timeout=5)
+    assert response.status_code == 201
+
     # Create a cam token
     return requests.post(urljoin(API_URL, f"cameras/{cam_id}/token"), headers=admin_headers, timeout=5).json()[
         "access_token"
--- a/client/tests/test_client.py
+++ b/client/tests/test_client.py
@@ -40,7 +40,7 @@ def test_cam_workflow(cam_token, mock_img):
         cam_client.create_detection(mock_img, 123.2, None)
     with pytest.raises(ValueError, match="bboxes must be a non-empty list of tuples"):
         cam_client.create_detection(mock_img, 123.2, [])
-    response = cam_client.create_detection(mock_img, 123.2, [(0, 0, 1.0, 0.9, 0.5)])
+    response = cam_client.create_detection(mock_img, 123.2, [(0, 0, 1.0, 0.9, 0.5)], pose_id=1)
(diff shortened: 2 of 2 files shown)


src/app/api/api_v1/endpoints/detections.py
```diff
@@ -60,6 +60,7 @@ async def create_detection(
         max_length=settings.MAX_BBOX_STR_LENGTH,
     ),
     azimuth: float = Form(..., ge=0, lt=360, description="angle between north and direction in degrees"),
+    pose_id: int = Form(..., gt=0, description="pose id of the detection"),
```
MateoLostanlen: maybe below we need 

`    poses: PoseCRUD = Depends(get_pose_crud),`

scripts/dbdiagram.txt
```diff
@@ -1,77 +1,71 @@
-Enum "userrole" {
```
MateoLostanlen: why are we deleting this ?

src/app/models.py
```diff
@@ -59,10 +59,19 @@ class Camera(SQLModel, table=True):
...
+    camera_id: int = Field(..., foreign_key="cameras.id", nullable=False)
+    azimuth: float = Field(..., ge=0, lt=360)
+    patrol_id: str | None = Field(default=None, max_length=100)
+
+
 class Detection(SQLModel, table=True):
     __tablename__ = "detections"
     id: int = Field(None, primary_key=True)
     camera_id: int = Field(..., foreign_key="cameras.id", nullable=False)
+    pose_id: int = Field(..., foreign_key="poses.id", nullable=True)
```
MateoLostanlen: maybe safer to allow None like for sequences 


```
    pose_id: int | None = Field(default=None, foreign_key="poses.id", nullable=True)
    sequence_id: int | None = Field(default=None, foreign_key="sequences.id", nullable=True)
```

MateoLostanlen (commented):
Hi @fe51, thnaks for the PR 

few remarks 

out need from .crud_pose import * insrc/app/crud/__init__.py and from .poses import * in src/app/schemas/__init__.py

MateoLostanlen:
We miss a route to get all poses associated to a camera

src/app/models.py
MateoLostanlen: not sure to understand, here is cna't be bull

 pose_id: int = Field(..., foreign_key="poses.id", nullable=True)
 
 I just wanted to mimik the way it's done for sequence_id

MateoLostanlen (changes requested):
Hi please can you update create_detection incleint too ? we need to pass a pose id now

Issue #524, opened 2025-10-17
Adds new table Occlusion Mask & creates CRUD related endpoints


fe51:
# Pitch

Occlusion masks allow users to define areas that generate recurring false positives. (occlusion masks are related to a defined position of a camera) -> https://github.com/pyronear/new-pyro-platform/issues/85 or https://github.com/pyronear/new-pyro-platform/pull/103 give an indication of how this look on GUI

Theses masks are regularly retrieved by onsite devices (raspberry pi) in order not to generate alerts in these areas (which would then be distributed but irrelevant)

To date, (due to a lack of time initially) the implementation of occlusion masks management has been done outside of the API. 
However, this creates additional dependencies, as the current use of write/read on an S3 from the platform or Raspberry Pi is not optimal in terms of rights management, and this is the kind of information that should be managed in the API.
(you can find more context on how it is done currently at the end of the issue)

**Occlusion masks must be managed through pyro-api, and so far, stored in the db :** 

### About the table `occlusion_masks`

bbox field : For the first iteration, we could keep the current input so an bbox as in detection table
> string representation of list of detection localizations, each represented as a tuple of relative coords (max 3 decimals) in order: xmin, ymin, xmax, ymax, conf (see table detections)

Below some guideline about the new table (Description is here for the purpose of the issue but not sure if we include it postgresSQL tables)

Do we include a created_at field ? -> To discuss

| Column      | Type      | Constraints                           | Description                          |
|-------------|-----------|----------------------------------------|--------------------------------------|
| id          | INT       | PRIMARY KEY                            | Unique identifier                    |
| pose_id   | INT       | FOREIGN KEY → poses(id)               | Reference to the related pose in poses table     |
| bbox     | String     | see in detections     | a given occlusion bbox   for a pose          |
| created_at  | TIMESTAMP | DEFAULT CURRENT_TIMESTAMP              | Record creation timestamp            |

# What is need to be done

- As a user, I want to update/set occlusion masks of a position of a camera in my organisation
- As a camera (often raspberry pi managing each camera on a site), I want to read occlusion mask of a given position
- As a admin (in case wanting to define the occlusion areas before setting up access to users, or define sky areas or other areas to avoid poor first impressions.),  I want to update/set occlusion masks for a given a position of a camera.

Suggestion guidelines

- Add every thing related to the introduction of a the new table occlusion masks in the api
- CRUD for API Occlusion Mask Resource -> To be detailed
- Add migration script 
[...]

PR #533, opened 2025-12-15
Introduces Occlusion masks table and related endpoints

fe51:
TO REVIEW AFTER  #531 (introducing poses)

This PR to address #524  which is part of the planned changes to the data model

It introduces the following modifications

- Add table occlusion_masks in model
- Add occlusion_masks schema & crud
- Add occlusion_masks CRUD endpoint
- Updates poses endpoints adding occlusion mask listing related to a pose
- Update tests
- Update db diagram

note
- occlusion mask read endpoint is permissive endpoint on (as long as authentified on the api)

What is missing (to be discussed of to handle in another PR)

- Alembic migrations
- Client updates -> will be done later
- Client tests (struggled to be properly run locally)

Happy to discuss it

### DIFF
--- a/src/app/crud/crud_occlusion_mask.py
+++ b/src/app/crud/crud_occlusion_mask.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2024-2025, Pyronear.
+# Copyright (C) 2025-2026, Pyronear.
 
 # This program is licensed under the Apache License 2.0.
 # See LICENSE or go to <https://www.apache.org/licenses/LICENSE-2.0> for full license details.


src/tests/conftest.py
```diff
@@ -94,10 +94,59 @@
...
+    {
+        "id": 2,
+        "pose_id": 1,
+        "mask": "(0.1,0.1,0.9,0.9,1)",
+        "created_at": datetime.strptime("2025-01-02T00:00:00.000000", dt_format),
+    },
+    {
+        "id": 3,
+        "pose_id": 2,
+        "mask": "(1,0.1,0.9,0.9,0.5)",
```
MateoLostanlen: it might be a good idea to test with more than one mask

MateoLostanlen (approved):
Thanks a lot for this PR, LGTM !

PR #539, opened 2025-12-28
Split detections

MateoLostanlen:
#### Detection model and API changes

* Dropped `Detection.azimuth`
* Made `pose_id` mandatory for all detections

* Creation logic now:

  * Explodes multi bbox payloads into **one Detection per bbox**
  * Stores sibling bounding boxes in `others_bboxes`
  * Ensures all detections reference the **same uploaded media key**

#### Sequencing logic improvements

* Sequences are now looked up using:

  * `camera_id`
  * `pose_id`
  * Recent `last_seen`
* For each new detection:

  * Its bbox is parsed
  * A candidate sequence is reused **only if the last bbox overlaps**
  * Otherwise, a new sequence is created

    * Cone is computed using the **pose azimuth**
    * First bbox initializes the spatial reference
* This prevents merging detections that are spatially distant but temporally close

#### Helpers refactor

* Bbox parsing moved to a shared scope
* Overlap and parsing helpers are always available
* Cone computation consistently relies on pose azimuth

### Why this matters

The main goal of this change is to **separate unrelated detections coming from the same image**.
When a single frame contains both a real fire and a false positive, these detections have nothing in common spatially or semantically and must not end up in the same sequence.

By enforcing one detection per bbox and reusing sequences only when spatial overlap exists, we ensure that fires and false positives evolve independently, even if they originate from the same image and timestamp

### DIFF
--- a/src/app/core/config.py
+++ b/src/app/core/config.py
@@ -49,6 +49,14 @@ class Settings(BaseSettings):
     MAX_BBOX_STR_LENGTH: int = (
         2 + MAX_BOXES_PER_DETECTION * (2 + 5 * (2 + DECIMALS_PER_COORD) + 4 * 2) + (MAX_BOXES_PER_DETECTION - 1) * 2
     )
+    # Single bbox (for Detection.bbox field): max 1 box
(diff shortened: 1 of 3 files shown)


fe51 (approved):
Made commits to fixed some discussed points. 
Should be challenged once you are back, but thanks a lot for this feature !

PR #554, opened 2026-03-18
Fix/camera list resilient s3 image url

fe51:
Fixes #553 "_GET /cameras/ fails entirely when a single camera's S3 image is unavailable_"

This PR introduces the following modifications

- Schema (CameraRead.last_image_url): updated field description to document that null is returned when the image is not yet uploaded or temporarily unavailable in S3
- GET /cameras/: added description on the route documenting the null fallback behavior for last_image_url
- GET /cameras/{camera_id}: same documentation addition as above
- fetch_cameras / get_camera: last_image_url resolution is now wrapped in a try/except — any HTTPException raised by the S3 presign call is caught and the field falls back to null instead of aborting the response


MateoLostanlen (dismissed):
thanks for you fix @fe51

MateoLostanlen (approved):

PR #559, opened 2026-04-07
Add detections_count to sequence and alert responses

ThbltLmr:
Closes #551 

## Summary
- add a shared helper to batch count detections by sequence
- expose `detections_count` on sequence responses returned by sequence endpoints
- expose `detections_count` on sequence payloads returned by alert endpoints
- cover the new field in alert and sequence endpoint tests

## Testing

https://github.com/user-attachments/assets/27edabef-b4d4-460f-af09-61ca1034a77e

## Notes
- I do not have the access rights to the API repo so I had to fork it sorry
- I added the detection count to all the endpoint that were listed in the issue #551, maybe they're not all needed. Let me know if we need to remove the property from some endpoints


src/app/api/api_v1/endpoints/_sequence_counts.py
```diff
@@ -0,0 +1,29 @@
...
+        return {}
+
+    stmt: Any = (
+        select(cast(Any, Detection.sequence_id), func.count(Detection.id))
+        .where(cast(Any, Detection.sequence_id).in_(sequence_ids))
+        .group_by(cast(Any, Detection.sequence_id))
+    )
+    res = await session.exec(stmt)
+    return {
+        int(sequence_id): int(detections_count)
```
fe51: Why the int() wrapping? After a DB fetch, sequence.id/alert.id are already int per the model annotation, this seems redundant. it is about test/style/ruff ?

fe51: Comment here apply in other occurence of int(sequence.id) and int(alert.id)

src/app/services/sequence_counts.py
fe51, line 1: Could this live in services/ rather than endpoints/? 
It's an aggregation query, not a CRUD operation, and services/ would be a more natural home for it (instead of here with endpoints only),   and it will be easier to reuse if other parts of the codebase needs it later (even if it might not)

src/app/api/api_v1/endpoints/alerts.py
```diff
@@ -45,10 +46,16 @@ async def _fetch_sequences_by_alert_ids(session: AsyncSession, alert_ids: List[i
     return mapping
 
 
-def _serialize_alert(alert: Alert, sequences: List[Sequence]) -> AlertReadWithSequences:
+def _serialize_sequence(sequence: Sequence, detection_counts: Dict[int, int]) -> SequenceRead:
```
fe51: _serialize_sequence has two different signatures across alerts.py and sequences.py. Could you align them? The sequences.py version (sequence, detections_count: int) seems simpler — the caller does the dict lookup, the function just serializes.

src/tests/endpoints/test_alerts.py
```diff
@@ -59,6 +64,20 @@ async def _create_alert_with_sequences(
     await session.commit()
     for seq in sequences:
         await session.refresh(seq)
+    for sequence, detections_count in zip(sequences, detections_count_by_sequence, strict=False):
```
fe51:
```suggestion
    for sequence, detections_count in zip(sequences, detections_count_by_sequence, strict=True):
```

strict=False silently ignores a length mismatch between sequences and detections_count_by_sequence. strict=True would catch it immediately if the fixture is ever changed.

fe51 (changes requested):
Hi @ThbltLmr , Thank you a lot for your PR !

I have a few comments, but no big changes ! 

Also, about the consideration of adding detection_counts to each sequences related endpoints, here is my views : 
- frontend actually uses only /alerts/unlabeled/latest and /alerts/all/fromdate and we will need detections count (later or perform pagination and improve front endplayer)
- Elsewhere, we do not require performance, and the query added is no expensive, so lets add it. Could be deleted later if not needed and an issue. 

Thanks again, I think after answering this comments we will be able to merge quite quickly :)

src/app/api/api_v1/endpoints/_sequence_counts.py
ThbltLmr: done in d8b1d5b!

src/app/services/sequence_counts.py
ThbltLmr: done in d59fc46!

src/app/api/api_v1/endpoints/alerts.py
ThbltLmr: done in 65c3734!

src/tests/endpoints/test_alerts.py
ThbltLmr: done in b598e19!

fe51 (approved):
Hi @ThbltLmr, thank you for taking comments into accounts ! All good to merge !

PR #564, opened 2026-04-14
Add active boolean field to Pose model

MateoLostanlen:
Poses represent camera orientations that can change over time. Previously, all poses (including outdated ones)   
  were returned when fetching cameras, which cluttered the frontend with stale data. We still need to keep old
  poses in the database for historical consistency (linked detections, sequences, etc.), so rather than deleting   
  them, we introduce an active boolean flag to soft-filter them.

  - Add active boolean field to the Pose model (defaults to True)
  - Camera endpoints (GET /cameras/{id} and GET /cameras/) now only return active poses
  - Alembic migration sets server_default=true so all existing poses remain active                                 
  - Python client updated to support active in create_pose and update_pose


fe51 (approved):
Thanks a lot for this feature !

PR #565, opened 2026-04-14
Filter out non-trustable cameras by default in fetch endpoint

MateoLostanlen:
- GET /cameras now excludes cameras with is_trustable=False by default                                           
  - Added include_non_trustable boolean query parameter (default false) to opt into returning all cameras
  - Works for both admin (all orgs) and org-scoped users

### DIFF
--- a/client/pyroclient/client.py
+++ b/client/pyroclient/client.py
@@ -110,7 +110,7 @@ class Client:
 
     # CAMERAS
     def fetch_cameras(self) -> Response:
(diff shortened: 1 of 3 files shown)


fe51 (commented):
Thanks for this quick PR !
Tested and ok to me !

Perhaps I’m being a bit too wordy, but I would have liked the endpoint description to say: `By default, it only returns trusted cameras``

After that point discussed I will be ready to approve

PR #566, opened 2026-04-15
Add alert-specific URL and improve Slack notification message

MateoLostanlen:
- Include alert ID in platform URL (e.g. /alert/38187) when available                                                                                                                                             
  - Shorten S3 image URL using Slack mrkdwn link format
  - Fix French accents in notification text                                                                                                                                                                         
  - Fix image alt text placeholder                                                                                                                                                                                  
  - Replace bare platform URL with an action-verb CTA: "Visualiser l'alerte en détail sur la plateforme Pyronear"                                                                                                   
  - Drop Slack image block — presigned/internal S3 URLs triggered invalid_blocks errors                                                                                                                             
  - Make platform URL configurable via new PLATFORM_URL env var (default https://platform.pyronear.org); wired into docker-compose.yml and docker-compose.dev.yml                                                   
  - Remove unused url parameter from slack_client.notify and its caller                                                                                                                                             
  - Log the failing payload alongside Slack error responses to ease future debugging    
  
  
Screenshot 2026-04-15 at 12 27 23


fe51:
Hi @MateoLostanlen, thanks for the PR. 

Few questions before reweing

Have you tested on a dedicated test slack channel ?

Also, could you add a screenshot of how it renders on slack ? 

When I have tested in #547 it was looking like below (unfortunately the image is no more available in my example)

image

fe51 (dismissed):
Hello @MateoLostanlen  ! Thank for this PR and discussion about it ! Good to me

fe51 (commented):
Why do you have the fallback for the platform URL in two places: in the compose file and in the config file? Shouldn't it just be one or the other?

fe51 (dismissed):
Thanks for the udpates, all good now !

PR #567, opened 2026-04-15
feat: add endpoint to update camera is_trustable status

MateoLostanlen:
I did #565 witout adding a route to update the status, here it is ..


fe51 (approved):
Thank you for the quick fix and related tests !

PR #568, opened 2026-04-16
fix: use sequence_azimuth in Slack notifications instead of pose_azimuth

MateoLostanlen:


src/app/services/slack.py
```diff
@@ -51,11 +51,9 @@ def notify(
         except json.JSONDecodeError as e:
             raise ValueError("Invalid JSON format for message_detection") from e
 
-        azimuth = detection_data.get("azimuth")
+        azimuth = detection_data.get("sequence_azimuth")
         if azimuth is None:
             azimuth = detection_data.get("pose_azimuth")
```
fe51: That's not the point of this PR, but this approach to azimuth is misleading.

The slack message always indicates the **detection** azimuth. Hence, we should return the detection azimuth, or report it as unknown. Trying to fill in the value with other data seems misleading to me.

MateoLostanlen: you are right let's drop this

MateoLostanlen: done

fe51 (approved):
LGTM

PR #569, opened 2026-04-16
Adapt camera-proxy to new pyro-camera-api client

MateoLostanlen:
Summary
 
 After upted in [pyro-camera-api client](https://github.com/pyronear/pyro-engine/pull/335)
                                                                                                                                                       
  - Aligns /cameras/{id}/... proxy routes with the refreshed pyro_camera_api_client: renames pos_id → patrol_id on /capture, and adds duration / zoom  
  to the legacy /control/move.                                                                                                                         
  - Exposes the new focused PTZ endpoints (preferred over the overloaded move): goto_preset, start_move, stop_move, move_for_duration, move_by_degrees,
   click_to_move, plus GET /control/speed_tables.                                                                                                      
  - speed on /control/move_by_degrees is optional — when omitted the device auto-picks the best calibrated speed (preferred).
  - Temporarily pins pyro-camera-api-client to the calibrate_reolink_cam branch until those client changes land on develop; switch back before merge.


fe51 (commented):
Hi @MateoLostanlen , I have made small modificaiton as suggestion

- Introduced CameraDirection = Literal["Left", "Right", "Up", "Down"] type alias; applied it to start_move, move_for_duration, and move_by_degrees — invalid directions now return 422 instead of failing downstream
- Added deprecated=True to the /control/move route decorator so it appears struck-through in the OpenAPI docs

Also, I have one question about `/control/stop` and `/control/stop_move`

Do these call different downstream methods that have genuinely different scopes? If yes, the summaries should reflect that:

- stop_move → "Halt the current continuous PTZ move" ? 
- stop_camera → "Stop all camera activity" (or whatever it actually does) ? 
If they turn out to do the same thing, one of them should be removed or at least stop should be deprecated like move.

src/app/api/api_v1/endpoints/camera_proxy.py
```diff
@@ -167,10 +173,112 @@ async def proxy_move(
...
 
 
-@router.post("/{camera_id}/control/stop", status_code=status.HTTP_200_OK, summary="Stop camera movement")
+@router.post(
+    "/{camera_id}/control/click_to_move",
+    status_code=status.HTTP_200_OK,
+    summary="Move toward a normalized image click",
+)
+async def proxy_click_to_move(
+    click_x: float = Query(..., ge=0.0, le=1.0, description="Normalized x coordinate in [0, 1]"),
```
fe51: Would not be more robust to put x y. in the payload/body instead of a param url ?

MateoLostanlen:
Hi @fe51, fair point. Keeping /control/stop_move for symmetry with start_move, and deprecating the legacy /control/move + /control/stop together. They'll be removed in a later cleanup pass on pyro-engine + the API once the focused routes are fully adopted.

src/app/api/api_v1/endpoints/camera_proxy.py
MateoLostanlen: Agreed

fe51 (approved):
Hi @MateoLostanlen thank you for the update ! Lets merge !

PR #570, opened 2026-04-16
ci: publish docker image to dockerhub and ghcr on version tags

MateoLostanlen:


fe51 (approved):
Thanks a lot ! 

@MateoLostanlen  Can you create I think the first tag, I think we can consider being in v1.0.0 ? to cjeck if it works ?

(image name is alert-api)

PR #571, opened 2026-04-17
fix: skip triangulation between sequences from the same pose

MateoLostanlen:
## Summary                                                                                                                                                                                
  - Two sequences from the same `pose_id` share the same camera apex. Any cone                                                                                                              
    intersection between them is degenerate (a wedge along the bisector, not a                                                                                                              
    real smoke location), so they must not be triangulated together.                                                                                                                        
  - `_find_overlapping_pairs` now skips pairs that share a `pose_id`. The check                                                                                                             
    is guarded by `has_pose`, so callers that don't pass `pose_id` keep working.                                                                                                            
  - Added `pose_id` to the records built in `detections.py` and `sequences.py`                                                                                                              
    so the filter actually fires in production.                                                                                                                                             
                                                                                                                                                                                            
  ## Context                                                                                                                                                                                
  Investigated alert 38460 on prod. Camera 66 / pose 105 produced two separate                                                                                                              
  sequences (38540 and 38547) with non-overlapping bboxes, both linked to the                                                                                                               
  same alert via cam 41. The current time-overlap check happened to block direct                                                                                                            
  pairing between the two pose-105 sequences in that case, but the guard was                                                                                                                
  missing: two concurrent sequences on the same pose (e.g. two plumes in                                                                                                                    
  different parts of the frame) would pass the time check and fake-triangulate.

### DIFF
--- a/src/app/api/api_v1/endpoints/detections.py
+++ b/src/app/api/api_v1/endpoints/detections.py
@@ -151,6 +151,7 @@ def _build_overlap_records(
             continue
         records.append({
             "id": int(seq.id),
+            "pose_id": seq.pose_id,
             "lat": float(cam.lat),
             "lon": float(cam.lon),
             "sequence_azimuth": float(seq.sequence_azimuth),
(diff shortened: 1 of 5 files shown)


fe51 (approved):
Thks for the quick fix

PR #574, opened 2026-04-24
chore: replace deprecated datetime.utcnow() with a helper

Acruve15:
## Summary
- Python 3.12 deprecated `datetime.utcnow()` and 3.13 will remove it. Swapped every call site in `src/` for a new helper `app.core.time.utcnow()`.
- Helper returns a naive UTC datetime (`datetime.now(timezone.utc).replace(tzinfo=None)`). Keeps semantics identical to what we had, so no migration is needed right now.
- Also cleaned up the `datetime` imports that became unused after the swap (ruff auto-fix).

## Context
I picked this up as a ramp-up exercise. I haven't touched this repo in a while and wanted a small, mechanical change to get back into the layout. Figured I'd turn it into a real PR since the deprecation is going to bite us on Python 3.13 anyway.

## Why not go fully tz-aware
All our `datetime` columns map to `TIMESTAMP WITHOUT TIME ZONE`, and several handlers (detection sequence-relaxation, 24h windows in sequences/alerts, camera `last_active_at`) compare DB-returned naive values against `utcnow() - timedelta(...)`. Going tz-aware here without a matching Alembic migration would start throwing `can't compare offset-naive and offset-aware datetimes`. Worth doing eventually, but I'd rather keep this PR just about the deprecation.

## Out of scope
- `datetime.utcfromtimestamp` in `tests/test_security.py:43`: same class of deprecation, will do in a follow-up.
- `client/` package untouched (published separately).

## Test plan
- [x] `ruff format --check .` and `ruff check .` clean (one pre-existing `RUF069` in `client/tests/test_client.py` is unrelated)
- [x] `src/tests/test_time.py` passes (3/3) locally
- [ ] `make test` in CI
- [ ] Sanity: login + inspect JWT `exp`, create a detection and verify `created_at` is set

### DIFF
--- a/src/app/api/api_v1/endpoints/alerts.py
+++ b/src/app/api/api_v1/endpoints/alerts.py
@@ -4,7 +4,7 @@
 # See LICENSE or go to <https://www.apache.org/licenses/LICENSE-2.0> for full license details.
 
 
-from datetime import date, datetime, timedelta
+from datetime import date, timedelta
 from typing import Any, Dict, List, Union, cast
 
 from fastapi import APIRouter, Depends, HTTPException, Path, Query, Security, status
@@ -13,6 +13,7 @@ from sqlmodel import delete, func, select
 from sqlmodel.ext.asyncio.session import AsyncSession
 
 from app.api.dependencies import get_alert_crud, get_jwt
+from app.core.time import utcnow
 from app.crud import AlertCRUD
 from app.db import get_session
 from app.models import Alert, AlertSequence, Sequence, UserRole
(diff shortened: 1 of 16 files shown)


Acruve15:
Hey @MateoLostanlen  👋 long time no see :)

Quick note on the red `triage` check: it's not from this PR. It fails on every PR opened from a fork.

`actions/labeler` needs write access on PRs, but GitHub silently downgrades `GITHUB_TOKEN` to read-only when a workflow runs on `pull_request` from a fork. The `pull-requests: write` line in the workflow has no effect in that case.

The usual fix is switching the `triage` job in `.github/workflows/pull_requests.yml` from `pull_request` to `pull_request_target`. Safe pattern here since the labeler only reads config.

Happy to put up a small follow-up PR if easier on your side!

fe51 (approved):
Hi @Acruve15, great to see a PR from you again :) 

This is a clean fix, nothing to nitpick.

Followups: 

- You might be right about naive datetimes that should be properly tz awared ! A migration in a another PR should be considered (after #572)

- About CI failing from fork. Your suggestion seems to be right. What would be reason not to do it ? (As long as CI runs at not cost for us, and PR from forks are not that common)  -> Also, I will upgrade back your rights on the repo as writer so you will be able to open PR not from a fork

Acruve15:
> Hi @Acruve15, great to see a PR from you again :)
>
> This is a clean fix, nothing to nitpick.
😍

> Followups:
> 
> * You might be right about naive datetimes that should be properly tz awared ! A migration in a another PR should be considered (after #572)

Anything blocking to merge PR #572? Happy to review if it helps 😊 

> * About CI failing from fork. Your suggestion seems to be right. What would be reason not to do it ? (As long as CI runs at not cost for us, and PR from forks are not that common) 

No reason not to do it to be honest. I'll fix this in another PR.

---
_sorry, I forget tag you..._

MateoLostanlen:
Hi @Acruve15 just missing a review for #572, I'd love a review :)

PR #575, opened 2026-04-26
feat(detections): accept optional crop image for faster smoke checks

MateoLostanlen:
## Summary
- Add an optional `crop` upload to `POST /detections` so cameras can ship a tight, pre-cropped view of the smoke alongside the full frame. The goal is to give reviewers a faster signal when triaging sequences — they can confirm or dismiss a smoke from the crop without waiting on the full frame.
- Persist the crop reference on `Detection` via a new `crop_bucket_key` column (alembic migration chained onto the current baseline) and surface it as `crop_url` on `GET /detections/{id}/url` and `GET /sequences/{id}/detections`.
- The crop endpoint param is opt-in: existing camera clients keep working unchanged, and `/sequences/{id}/detections` accepts `with_crop=false` to skip the extra presign round-trips when the consumer doesn't need crops.

## Notable changes
- `endpoints/detections.py`: authorize the pose **before** uploading to S3 so a 403 no longer leaves an orphan object in the bucket. Crop upload happens only after authorization passes.
- `services/storage.py`: `upload_file` takes a `key_prefix` so frame and crop with identical bytes (same SHA prefix + same second) don't collide on the same bucket key. Crop keys are prefixed `crop_`.
- `client/pyroclient`: `create_detection(...)` gains a `crop: bytes | None` kwarg; `fetch_sequences_detections(...)` gains `with_crop: bool = True`. Also fixed the camera/pose image uploads to send `image/jpeg` instead of mislabelling JPEG bytes as `image/png`.


src/app/api/api_v1/endpoints/sequences.py
```diff
@@ -115,6 +115,10 @@ async def fetch_sequence_detections(
     sequence_id: int = Path(..., gt=0),
     limit: int = Query(10, description="Maximum number of detections to fetch", ge=1, le=100),
     desc: bool = Query(True, description="Whether to order the detections by created_at in descending order"),
+    with_crop: bool = Query(
+        True,
```
fe51: This suggestion to default to false, otherwise, frontend would need to be updated to not get crop (as long as not used yet)

```suggestion
        False,
```

fe51: Tests might be updated then :) maybe something like 
params={"with_crop": "true"} in test_fetch_sequence_detections_includes_crop_url ?

fe51 (changes requested):
Thanks a lot for this PR, few comments and suggestions

I am wondering if you have forgoten to do S3 cleanup on delete for crop image
  if detection.crop_bucket_key:
        bucket.delete_file(detection.crop_bucket_key)

src/app/api/api_v1/endpoints/sequences.py
MateoLostanlen: done :)

MateoLostanlen:
Thanks @fe51 ! Both addressed:
  
  - S3 cleanup on delete: delete_detection now also removes crop_bucket_key, so no more orphan crop objects
  - with_crop default: flipped to false so the frontend keeps its current behavior until it opts in; updated the test to pass with_crop=true

fe51 (approved):
Thank you for taking feedbacks into account, lets merge

PR #576, opened 2026-04-27
fix: normalize sequence azimuth to [0, 360) range

MateoLostanlen:
- `resolve_cone` could return negative or >=360 cone azimuths when a bbox sat near a frame edge with a low/high camera azimuth, leaving rows with `sequence_azimuth < 0` in the DB. Observed in production on sdis-77, alert 39412 (https://platform.pyronear.org/alert/39412).
- Apply modulo 360 after the existing rounding so the returned cone azimuth always falls in `[0, 360)`, with the upper bound strictly exclusive.
- Adds a parametrized unit test covering the centered, negative-wrap, over-360-wrap, and exact-360 cases.


fe51 (approved):
Hi @MateoLostanlen , thank for this fix introducing modulo 360 !

fe51 (approved):

Issue #521, opened 2025-10-10
Endpoint to export alerts as CSV for a given period


fe51:
## Pitch
As a first iteration on this topic, we want to expose an endpoint that allows exporting alerts (or sequences) over a given time range (as csv)

## What to do 
Create an endpoint that:

- Takes two dates as input (start_date, end_date)
- Returns a CSV file containing the list of alerts within that period

### Data

The CSV should include the following columns from the alert table (excluding orga_id):
- id (int, not null)
- lat (float)
- lon (float)
- started_at (timestamp, not null)
- last_seen_at (timestamp, not null)

### Future improvements

In a second iteration, we will enrich the export with additional columns, such as:
- Related sequence IDs
- Associated camera names for each alert

fe51:
closes by #577

PR #577, opened 2026-04-28
feat: add CSV export endpoint for alerts (#521)

Acruve15:
## Summary
- Adds `GET /alerts/export?from_date=YYYY-MM-DD&to_date=YYYY-MM-DD` returning a CSV streamed download with columns `id, lat, lon, started_at, last_seen_at`.
- Filters alerts by `started_at` within the inclusive UTC window `[from_date 00:00:00, to_date 23:59:59.999999]` and scopes results to the caller's `organization_id`.
- Implements the first iteration of [#521](https://github.com/pyronear/pyro-api/issues/521); the issue's "Future improvements" (sequence ids, camera names) are deliberately left for a follow-up.

## Implementation notes
- Route declared **before** `/{alert_id}` in `alerts.py` so FastAPI doesn't try to parse `export` as an `int` path param.
- Uses stdlib `csv` + `io.StringIO` + `StreamingResponse` (no new dependencies; pandas not needed for this shape).
- Inline `select(Alert).where(...)` mirrors `fetch_alerts_from_date`; nothing added to the CRUD layer.
- DB columns store naive UTC datetimes (per `app.core.time.utcnow`), so the bounds are built without `tzinfo`.
- Returns 422 when `to_date < from_date`, 401 when unauthenticated, 200 with header-only body for an empty range.

## Test plan
- [x] `pytest tests/endpoints/test_alerts.py -v` (all 10 tests, including 6 new export tests, pass)
  - Happy path + ascending order
  - Window narrows correctly (only middle-day alert returned)
  - Org isolation (org 2 alerts absent when called as org 1 user)
  - Empty range returns header-only CSV
  - `to_date < from_date` → 422
  - No token → 401
- [x] `ruff 0.11.9 check` and `ruff 0.11.9 format --check` pass
- [ ] Manually `curl` the new endpoint with a valid token and confirm the CSV opens cleanly

### DIFF
--- a/src/tests/endpoints/test_alerts.py
+++ b/src/tests/endpoints/test_alerts.py
@@ -389,9 +389,9 @@ async def test_alerts_export_happy_path(async_client: AsyncClient, detection_ses
 @pytest.mark.asyncio
 async def test_alerts_export_window_narrows(async_client: AsyncClient, detection_session: AsyncSession):
     base = datetime(2026, 4, 10, 12, 0, 0)
-    a_before = await _create_alert(detection_session, 1, base, base + timedelta(minutes=5))
+    await _create_alert(detection_session, 1, base, base + timedelta(minutes=5))
     a_in = await _create_alert(detection_session, 1, base + timedelta(days=1), base + timedelta(days=1, minutes=5))
-    a_after = await _create_alert(detection_session, 1, base + timedelta(days=2), base + timedelta(days=2, minutes=5))
+    await _create_alert(detection_session, 1, base + timedelta(days=2), base + timedelta(days=2, minutes=5))
 
     auth = pytest.get_token(
         pytest.user_table[0]["id"], pytest.user_table[0]["role"].split(), pytest.user_table[0]["organization_id"]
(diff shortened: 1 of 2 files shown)


src/tests/endpoints/test_alerts.py
```diff
@@ -317,3 +319,165 @@ async def test_triangulation_creates_single_alert(
...
+    resp = await async_client.get(
+        "/alerts/export?from_date=2026-04-11&to_date=2026-04-11",
+        headers=auth,
+    )
+    assert resp.status_code == 200, resp.text
+    _, data_rows = _parse_csv_body(resp.text)
+    returned_ids = {int(r[0]) for r in data_rows}
+    assert returned_ids == {a_in.id}
+    assert a_before.id not in returned_ids
+    assert a_after.id not in returned_ids
```
fe51: As long as testing `assert returned_ids == {a_in.id}` it impllies necessarily that the 2 following line passes right ?

fe51 (dismissed):
Hi Alexis, thanks a lot for this PR and sorry for the review delay ! 

Tested it works, and reviewed the code !
This is clean, and quite efficient (have not tested to extract for long time period but the content is light)

I have juste added one remark/question in tests but non blocking to me !

src/tests/endpoints/test_alerts.py
Acruve15: you are right!!

Acruve15: @fe51 you'll have to reapprove please :)

PR #578, opened 2026-04-28
fix: restore utcnow() in test_detections.py

Acruve15:
## Summary
Seven test functions in `src/tests/endpoints/test_detections.py` still call `datetime.utcnow()`, but the `datetime` import was removed in #574 when those call sites were missed during the cleanup. This breaks `pytest`, `pytest-client`, and `precommit-hooks` on every PR against `main`.

This PR restores the calls by using the already-imported `utcnow()` helper from `app.core.time` (same fix that was applied elsewhere in #574).

## Test plan
- [x] `pytest tests/endpoints/test_detections.py -v` (67 / 67 pass)
- [x] `ruff 0.11.9 check` clean

### DIFF
--- a/src/tests/endpoints/test_detections.py
+++ b/src/tests/endpoints/test_detections.py
@@ -511,7 +511,7 @@ DISTANT_LOCATION = (48.5529, 2.8536)
 @pytest.mark.asyncio
 async def test_filter_candidate_keeps_alert_within_threshold(detection_session: AsyncSession):
     alert_crud = AlertCRUD(detection_session)
-    now = datetime.utcnow()
+    now = utcnow()
     nearby = Alert(
         organization_id=1,
         lat=SMOKE_LOCATION[0] + 0.005,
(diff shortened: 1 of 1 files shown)


Acruve15:
@fe51 @MateoLostanlen Heads up on the codecov check: the failure here is a measurement artifact, not a real coverage drop.

## What's happening?

Our tests.yml workflow uploads coverage from two separate jobs (pytest for backend, pytest-client for client). Since [#574](https://github.com/pyronear/pyro-api/issues/574) merged, the pytest job has been failing on main because of leftover datetime.utcnow() calls in test_detections.py. That means every push to main since then only uploaded the client flag (3 files).

Codecov's flag carryforward kept the old backend numbers around for a while, but the cache eventually expired. Today main's stored coverage is just the 3 client files.

Why this PR looks like a regression
```
main (base):  3 files,   125 lines, 90.40%   (client only)
this PR:     52 files, 2225 lines, 88.75%   (full backend + client)
```
Codecov subtracts these two and reports -1.65%, but they're not measuring the same thing.

## Path forward

I split the datetime.utcnow() cleanup into [#578](https://github.com/pyronear/pyro-api/pull/578) so it can land on its own. Once that merges, main's pytest job will be green again and the next push to main will upload a complete backend report. Codecov's base will then match this PR's scope and the project check should pass automatically (no force-push needed on this PR).

[#578](https://github.com/pyronear/pyro-api/issues/578) itself will also fail codecov/project for the same reason. It would need a one-time admin override to merge. After that the system is self-healing and we don't need a codecov.yml workaround.

MateoLostanlen (approved):
Good catch thanks

PR #582, opened 2026-04-30
feat(alerts): unmatch a sequence from an alert

MateoLostanlen:
- Add `POST /alerts/{alert_id}/sequences/{sequence_id}/unmatch` (admin + agent) that detaches a sequence from an alert and recomputes the alert state; if the sequence becomes orphaned a new alert is created for it. Reject when the alert has a single sequence.
- Reject unmatch when the sequence's camera belongs to a different organization than the alert, to prevent cross-org alert creation.
- Extract `refresh_alert_state` from `endpoints/sequences.py` into `services/alerts.py` so both endpoints share the helper.
- Cover the new endpoint with tests for new-alert creation, no-op when the sequence is already linked elsewhere, single-sequence rejection, missing link, role and cross-org access denial.


fe51 (commented):
Hi @MateoLostanlen, thank you for the PR. 

Could you just check why test are failing (not clear to me) and f we are good I can approve and merge. 

After this PR merged, what is need frontend side ? We shall discuss the design for users in a dedicated issue.

fe51 (approved):
Thanks for fixing the tests. LGTM !

PR #583, opened 2026-05-05
feat: filter alerts/sequences/slack by daily fire risk score

MateoLostanlen:
## Why
Operators report way too many alerts on rainy / low-risk-weather days, when there is essentially no chance of an actual wildfire. The detection model still picks up smoke-like patterns (clouds, mist, exhaust) and they flood Slack and the alerts list, drowning real signal. This PR cross-references the per-camera daily Fire Weather Index from pyro-risk-api and drops low-confidence sequences when the local fire risk is `low` or `very_low` — the regime where false positives dominate.

## Summary
- Pulls a daily FWI class per camera from pyro-risk-api (cached in RAM, refreshed at 04:00 UTC) and uses it to drop low-confidence sequences during low-risk weather. On `low` days, drops sequences with max_conf < 0.45; on `very_low`, drops below 0.6; on `moderate+` (or unknown / risk-api unreachable), behaviour is unchanged.
- Applied to `/alerts/unlabeled/latest`, `/alerts/all/fromdate`, `/sequences/unlabeled/latest`, `/sequences/all/fromdate`, and to Slack notifications at sequence creation. Alerts that lose all their sequences are dropped from the response.
- Filter is pushed into the SQL `WHERE` clause (per-camera `CASE` expression on `Sequence.max_conf`), so `LIMIT/OFFSET` pagination stays exact instead of returning sparse pages.
- The `/all/fromdate` endpoints query the risk-api `/scores/{date}?organization_id=...` per request so the threshold matches the requested day; the daily cache only powers the "today" code paths.
- All four list endpoints accept an optional `?risk_score=<fwi_class>` query param that overrides the per-camera lookup and applies a single class to every sequence — useful as a kill-switch (`risk_score=moderate` disables filtering) or to preview a different risk regime without touching the cache.

## Schema change
Adds a nullable `max_conf` column to `sequences`, maintained at ingest:
- New sequence: computed from the primary bboxes of the bootstrap detections only (sibling boxes in `others_bboxes` represent unrelated detections in the same image and would otherwise inflate the score).
- Existing sequence extension: atomically bumped via portable `CASE WHEN` (`max_conf := GREATEST(max_conf, candidate)` semantics, runs on Postgres + SQLite).
- An Alembic migration adds the column and backfills existing rows from `Detection.bbox` only.
- Note: `max_conf` is monotonic non-decreasing — not recomputed when a detection is deleted/reassigned.

## Config
Adds `RISK_API_URL`, `RISK_API_LOGIN`, `RISK_API_PWD`, `RISK_REFRESH_HOUR_UTC` (default 4) to `.env.example` and `docker-compose.yml`. FWI thresholds live in code as `FWI_MIN_CONF` (dict in `services/risk.py`) covering all 6 EFFIS classes. When `RISK_API_URL` is unset, the feature is fully disabled (fail-open).


src/app/api/api_v1/endpoints/alerts.py
```diff
@@ -31,22 +34,44 @@ def verify_org_rights(organization_id: int, alert: Alert) -> None:
...
+    if seq_filter is not None:
+        seq_stmt = seq_stmt.where(seq_filter)
+    seq_stmt = seq_stmt.order_by(cast(Any, AlertSequence.alert_id), desc(cast(Any, Sequence.last_seen_at)))
     res = await session.exec(seq_stmt)
     for alert_id, sequence in res.all():
         mapping.setdefault(int(alert_id), []).append(sequence)
     return mapping
 
 
+async def _resolve_class_per_camera(
```
fe51: name shloud be more explicit, something like, fire_risk or fwi

```suggestion
async def _resolve_FWI_class_per_camera(
```

src/app/api/api_v1/endpoints/alerts.py
```diff
@@ -113,24 +138,35 @@ async def fetch_alert_sequences(
...
-        .where(Sequence.last_seen_at > utcnow() - timedelta(hours=24))
-        .where(Sequence.is_wildfire.is_(None))  # type: ignore[union-attr]
+    classes = await _resolve_class_per_camera(session, token_payload.organization_id, override_class=risk_score)
+    seq_filter = max_conf_filter_clause(classes)
+
+    seq_match: Any = select(AlertSequence.alert_id).join(
+        Sequence, cast(Any, Sequence.id == AlertSequence.sequence_id)
+    )
+    seq_match = seq_match.where(Sequence.last_seen_at > utcnow() - timedelta(hours=24))
+    seq_match = seq_match.where(Sequence.is_wildfire.is_(None))  # type: ignore[union-attr]
```
fe51: Why not  ? (same approach as before)
Might even be relavant to chain with previous lines from l152
```suggestion
    seq_match = seq_match.where(Sequence.last_seen_at > utcnow() - timedelta(hours=24)).where(Sequence.is_wildfire.is_(None))  # type: ignore[union-attr]
```

src/app/models.py
```diff
@@ -104,6 +104,9 @@ class Sequence(SQLModel, table=True):
     cone_angle: Union[float, None] = Field(None, nullable=True)
     started_at: datetime = Field(..., nullable=False)
     last_seen_at: datetime = Field(..., nullable=False)
+    # Highest detection confidence ever attached to this sequence.
+    # Monotonic: never recomputed downward when detections are deleted/reassigned.
+    max_conf: Union[float, None] = Field(None, nullable=True)
```
fe51: I am ok to add comment for field desc. However, would not it be the opportunity to put this comment as a postgres field desc
```suggestion
    max_conf: Union[float, None] = Field(None, nullable=True,         description="Highest detection confidence ever attached to this sequence. Monotonic: never decremented when detections are deleted or reassigned.",
)
```

fe51 (changes requested):
Thanks a lot for this exploratory PR ! 

A preliminary review following our discussion and before we’ve looked at the test updates. 

These aren’t critical comments at this stage. 

Perhaps the endpoint should make it clear that the alerts retrieved are, by default, filtered based on a risk score?

src/app/api/api_v1/endpoints/alerts.py
MateoLostanlen: done :)

src/app/api/api_v1/endpoints/alerts.py
MateoLostanlen: you are right better to chain , juste updated for readibily

src/app/models.py
MateoLostanlen: all good

src/tests/endpoints/test_risk_filter.py
```diff
@@ -0,0 +1,298 @@
...
+@pytest.mark.asyncio
+async def test_unlabeled_latest_keeps_all_when_class_is_moderate(
+    async_client: AsyncClient, detection_session: AsyncSession, reset_risk_cache
+):
+    camera_id = pytest.camera_table[0]["id"]
+    pose_id = pytest.pose_table[0]["id"]
+    low_seq = await _seed_unlabeled_sequence(detection_session, camera_id, pose_id, max_conf=0.10, minutes_ago=30)
+
+    risk_service._scores = {camera_id: "moderate"}
+
```
fe51, line 108: I might be a bit paranoid about this filter, but I reckon it might be worth introducing tests for ‘moderate’ **and above** right now, so that if we update the rules for those risk levels later, they’ll already have been tested.

Below this suggestion (but not tested)
```suggestion
@pytest.mark.asyncio
@pytest.mark.parametrize("fwi_class", ["moderate", "high", "very_high", "extreme"])
async def test_unlabeled_latest_keeps_all_when_class_is_moderate_or_above(
    fwi_class: str, async_client: AsyncClient, detection_session: AsyncSession, reset_risk_cache
):
    camera_id = pytest.camera_table[0]["id"]
    pose_id = pytest.pose_table[0]["id"]
    low_seq = await _seed_unlabeled_sequence(detection_session, camera_id, pose_id, max_conf=0.10, minutes_ago=30)

    risk_service._scores = {camera_id: fwi_class}

```

fe51 (changes requested):
Thanks for the updates, I might I have missed something, but, some tests might be missing

**/alerts/fromdate has zero tests** —> it got risk_score added same as the other endpoints, but nothing covers it. It also uses the distinct target_date code path in _resolve_class_per_camera, which is the only branch that hits risk_service.get_scores_for_date() —> that async path is not exercised by any test in the file.

**Alert with mixed sequences** —> no test for an alert that has one sequence above threshold and one below. The alert would appear (it matched the seq_match subquery via the passing sequence) but with only the surviving sequence in its payload. Worth having a test to pin that behavior as intentional. What do you think ?

src/tests/endpoints/test_risk_filter.py
MateoLostanlen: done

fe51 (approved):
Thanks for adding new tests

PR #584, opened 2026-05-05
fix: keep alert id stable when labeling a solo sequence FP

MateoLostanlen:
- When a sequence labeled non-wildfire is the only one in its alert, the previous flow detached it, deleted the (now empty) alert, and created an identical replacement — churning the alert id for no reason.
- Now the solo case is detected via a `LIMIT 1` siblings probe and the alert is left untouched.
- Multi-sequence alerts still split out the FP into its own alert so it stops contributing to triangulation.
- Adds an integration test for the stable-id case and a unit test for the early-return branch; updates the existing multi-sequence unit test to mock the new probe.

### DIFF
--- a/src/tests/endpoints/test_sequences.py
+++ b/src/tests/endpoints/test_sequences.py
@@ -448,6 +448,61 @@ async def test_sequence_label_updates_alerts(async_client: AsyncClient, detectio
 
 
 @pytest.mark.asyncio
+async def test_sequence_label_keeps_solo_alert(async_client: AsyncClient, detection_session: AsyncSession):
+    # A sequence alone in its alert should keep that alert as-is when labeled non-wildfire.
+    camera = await detection_session.get(Camera, 1)
+    assert camera is not None
+    now = utcnow()
+    seq = Sequence(
+        camera_id=camera.id,
+        pose_id=None,
+        camera_azimuth=180.0,
+        sequence_azimuth=170.0,
+        cone_angle=5.0,
+        is_wildfire=None,
+        started_at=now - timedelta(seconds=30),
+        last_seen_at=now - timedelta(seconds=10),
+    )
+    detection_session.add(seq)
+    await detection_session.commit()
+    await detection_session.refresh(seq)
+
+    alert = Alert(
+        organization_id=camera.organization_id,
+        lat=1.0,
+        lon=2.0,
+        started_at=seq.started_at,
+        last_seen_at=seq.last_seen_at,
+    )
+    detection_session.add(alert)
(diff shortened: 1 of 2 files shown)


src/tests/endpoints/test_sequences.py
```diff
@@ -447,6 +447,61 @@ async def test_sequence_label_updates_alerts(async_client: AsyncClient, detectio
...
+    resp = await async_client.patch(
+        f"/sequences/{seq.id}/label",
+        json={"is_wildfire": SequenceLabel(is_wildfire="other_smoke").is_wildfire},
+        headers=auth,
+    )
+    assert resp.status_code == 200, resp.text
+
+    alerts_res = await detection_session.exec(select(Alert).execution_options(populate_existing=True))
+    alerts_rows = alerts_res.all()
+    assert [row.id for row in alerts_rows] == [original_alert_id]
```
fe51: The two assertions query the entire Alert and AlertSequence tables, so they implicitly assume no other rows exist in the DB at that point. If test isolation ever weakens (shared fixtures, parallel runs, leftover data), these will fail for the wrong reason.

Scoping both queries to the sequence under test makes them resilient to whatever else is in the DB:


# Fetches only AlertSequence rows for THIS sequence, checks it still points to the original alert



```suggestion
alert_links_res = await detection_session.exec(
    select(AlertSequence.alert_id).where(AlertSequence.sequence_id == seq.id)...
)
assert set(alert_links_res.all()) == {original_alert_id}

```

src/tests/endpoints/test_sequences.py
```diff
@@ -447,6 +447,61 @@ async def test_sequence_label_updates_alerts(async_client: AsyncClient, detectio
...
+
+    mappings_res = await detection_session.exec(
+        select(AlertSequence.alert_id, AlertSequence.sequence_id).execution_options(populate_existing=True)
```
fe51:
```suggestion
        select(AlertSequence.alert_id, AlertSequence.sequence_id)
        .where(AlertSequence.sequence_id == seq.id)
        .execution_options(populate_existing=True)

```

fe51 (commented):
Thanks a lot for the PR @MateoLostanlen ! tested and work on devenv

Just a comment and suggestion, not blocking but wondering of would more relevant. Actually, current isolation is probably fine  but worth tightening since the fix is trivial. Tell me what do you think

MateoLostanlen:
Thanks @fe51 I added your change

fe51 (approved):
All good to merge now !

PR #590, opened 2026-05-12
Migrate Python tooling to UV and Ty

frgfm:
## Summary
- Replace Poetry metadata and `poetry.lock` with a UV workspace and a single root `uv.lock`.
- Split dependencies into UV groups for server, client, quality, tests, docs, and e2e usage.
- Replace Mypy with Ty in local checks, CI, and client quality extras.
- Run Docker and CI dependency installs through `uv sync`/`uv run` instead of Poetry exports.

## Main choices
- Kept the backend as a non-package application with `tool.uv.package = false`; `pyroclient` remains the only publishable package and is included as a workspace member.
- Put backend runtime dependencies in a `server` dependency group rather than root `project.dependencies`, so Docker/CI can install exact slices from the lockfile.
- Added a dependency sync checker that tracks UV/Ruff/Ty/Prek versions across pyproject, Dockerfiles, and workflows.
- Preserved missing-import tolerance by configuring Ty with `unresolved-import = "warn"`, then fixed or locally suppressed migration-specific diagnostics.
- Updated the backend test `AsyncClient` fixture to use `ASGITransport(app=...)`, matching current HTTPX syntax.
- Made client tests use the dev compose superadmin credentials explicitly, so CI job env does not break auth against `docker-compose.dev.yml`.
- Added client auth setup assertions so future login failures show status/body instead of `KeyError: access_token`.
- Made the migration check upgrade to `head` before testing the downgrade/upgrade cycle.

## Validation
- `make venv`
- `uv lock --check`
- `uv sync --locked --all-groups --all-packages --no-install-project`
- `make quality`
- `make test` (`481 passed`)
- `make test-client` (`14 passed`)
- Exact client CI sequence with dummy env present and command-level superadmin override (`14 passed`)
- Bad client credentials now fail as `AssertionError: {"detail":"Invalid credentials."}`, not `KeyError: access_token`
- Migration check command from `.github/workflows/backend.yml` completed with `--- Migration cycle successful ---`
- `make docs-client`
- `make e2e`
- Docker smoke: `docker compose -f docker-compose.dev.yml up -d --build --wait` plus backend `/status` check
- YAML parse check for workflows, Dependabot, and pre-commit config
- `docker compose -f docker-compose.dev.yml config --quiet`

## Deliberately excluded
- `AGENTS.md`
- `CLAUDE.md`
- `docker/minio/setup-s3.sh`
- unrelated untracked workflows not needed for this migration


MateoLostanlen:
Hello @frgfm ,

Thanks for this PR.

I was planning to switch from Poetry to uv, so this is perfect, thank you.

I really like the idea of moving to Ty, and the Astral team is doing great work. Do you have any concerns about the project still being relatively young?

Could you also explain the idea behind this new backend CI workflow, and why not reuse the existing ones?

frgfm:
> I really like the idea of moving to Ty, and the Astral team is doing great work. Do you have any concerns about the project still being relatively young?

None at all, I had the same concerns a while back and since I migrated all my other projects, I feel like Ty is way more mature and serious than mypy (performance wise but also quality wise, remember all those mypy ignore). The nice part is that it's so fast, you can use it as a language server and pre commit hook (if you had done that with mypy, you would be waiting 30sec every time)


> Could you also explain the idea behind this new backend CI workflow, and why not reuse the existing ones?

Yup I need to clear up some CI job redundancies actually!

MateoLostanlen:
Hi @frgfm 

ok thanks let's go then

ask me the review when yo uare done with cleaning 😄

frgfm:
Ready for review!
Short brief:
- 99% of the diff comes from the change of lockfile, no need to be scared
- check pyproject.toml changes first
- then check the diff in the dockerfiles
- the rest is adapting CI jobs, makefile commands and fixing typing checks

Feedback is very much welcome!

src/Dockerfile.test
```diff
@@ -0,0 +1,50 @@
...
+COPY --from=builder /opt/venv /opt/venv
+
+COPY pyproject.toml /app/pyproject.toml
+COPY src/app /app/app
+COPY src/tests /app/tests
+COPY src/migrations /app/migrations
+COPY src/alembic.ini /app/alembic.ini
+
+EXPOSE 8000
+HEALTHCHECK --interval=10s --timeout=3s --retries=5 CMD ["curl", "-f", "http://localhost:8000/health", "--max-time", "3"]
```
MateoLostanlen: our api exposes /status not health

MateoLostanlen (changes requested):
Hi @frgfm .  the approach looks sound

A couple of things I’d fix before merge:

1. `src/Dockerfile.test` healthcheck points to `/health`, but the app exposes `/status`. This image would report unhealthy even when the server is running
2. `client/.conda/meta.yaml` is stale versus `client/pyproject.toml`: the client now requires Python `>=3.10,<4` and `requests>=2.33.0,<3.0.0`, while the conda recipe still advertises Python `>=3.8` and `requests>=2.31.0`
3. Fix the header fleged by CI

frgfm:
>     1. `src/Dockerfile.test` healthcheck points to `/health`, but the app exposes `/status`. This image would report unhealthy even when the server is running

fixed!

>     2. `client/.conda/meta.yaml` is stale versus `client/pyproject.toml`: the client now requires Python `>=3.10,<4` and `requests>=2.33.0,<3.0.0`, while the conda recipe still advertises Python `>=3.8` and `requests>=2.31.0`

not the scope of this PR (my opinion is to remove conda all together, since uv brought first class Pytorch builds, no need for conda anymore)

>     3. Fix the header fleged by CI

fixed!

MateoLostanlen (approved):
Hi ,

Thanks for 1 and 3

for 2 it's kind of the scop you created the issue here by update requests>=2.33.0. But yes agree let's  drop conda in an upcoming pr

PR #592, opened 2026-05-14
feat(sequences): add offset query param to detections endpoint

ThbltLmr:
## Summary
- Adds an `offset` query parameter to the `GET /sequences/{sequence_id}/detections` endpoint, wiring through the existing `fetch_all` offset support so callers can page through all detections of a sequence.

## Test plan
- [x] New tests in `src/tests/endpoints/test_sequences.py` cover the offset behavior
- [x] Verify pagination works end-to-end against a populated sequence

### DIFF
--- a/src/app/api/api_v1/endpoints/sequences.py
+++ b/src/app/api/api_v1/endpoints/sequences.py
@@ -127,6 +127,7 @@ async def fetch_sequence_detections(
     if UserRole.ADMIN not in token_payload.scopes and token_payload.organization_id != camera.organization_id:
         raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail="Access forbidden.")
 
+    # Get the bucket of the camera's organization
     bucket = s3_service.get_bucket(s3_service.resolve_bucket_name(camera.organization_id))
     return [
         DetectionWithUrl(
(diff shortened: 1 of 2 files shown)


fe51 (dismissed):
Hi @ThbltLmr,
The kind of update that makes you wonder why we didn’t do it from the start, it feels so natural to add it ! 

Thanks, Thibault, for sorting it out and adding the necessary tests

src/app/api/api_v1/endpoints/sequences.py
```diff
@@ -126,7 +127,6 @@ async def fetch_sequence_detections(
     if UserRole.ADMIN not in token_payload.scopes and token_payload.organization_id != camera.organization_id:
         raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail="Access forbidden.")
 
-    # Get the bucket of the camera's organization
```
fe51, line 129: Why the comment deletion ?

fe51: @ThbltLmr juste quick question ? But i think we are good to merge after that :)

ThbltLmr: restored the comment in f101b3e!

fe51: thanks !

fe51 (approved):
Thanks again, lets merge !

PR #593, opened 2026-05-15
feat(alerts/export): expand CSV with sequence + camera context (#591)

### LINKED ISSUE #591: Expand alerts CSV export with sequence and camera context
Follow-up to #577.

The current `/alerts/export` endpoint only exposes 5 columns (`id`, `lat`, `lon`, `started_at`, `last_seen_at`), which is too thin for studying past alerts.

We want the export to include every field relevant to alert analysis.

### Proposed columns

**Core**
- `alert_id`
- `started_at_date` (UTC)
- `started_at_time` (UTC)
- `last_seen_at` (ISO)
- `duration_seconds`
- `lat`, `lon` (empty when null)
- `organization_id`

**Sequences** (multi-value cells, `|`-separated, ordered by `sequence.started_at` ASC so positions align across columns)
- `sequence_ids`
- `sequence_count`
- `sequence_labels` (`is_wildfire`: `wildfire` / `other` / `unknown`)
- `sequence_azimuths`
- `pose_ids`

**Cameras** (same ordering as sequences)
- `camera_ids`
- `camera_names`
- `camera_count`

Acruve15:
## Summary

- Expands `GET /alerts/export` from the 5-column alert-only CSV shipped in #577 to a **16-column long format with one row per `(alert, sequence)` pair**, picking up sequence + camera context for analysts who want to slice past alerts in Excel/Sheets without going back to the API.
- Reuses the existing `_fetch_sequences_by_alert_ids` helper; adds a focused `_fetch_camera_names_by_ids` that selects only `(id, name)`. A `_WILDFIRE_LABELS` dict collapses `Sequence.is_wildfire` → `wildfire` / `other` / `unknown`.
- Closes #591. Heads-up for pyronear/new-pyro-platform#220 below.

## Departures from the issue spec

Both decided after looking at how the file actually gets used (Excel/Sheets pivots). Happy to revert either if @MateoLostanlen prefers the original shape.

- **Long format instead of `|`-separated multi-value cells.** The issue proposed `id1|id2|id3` packed cells positionally aligned across `sequence_*` and `camera_*` columns. That shape is fragile (a `|` in a `camera_name` silently breaks parsing) and kills Excel's filter + pivot on `sequence_label` / `camera_name`. One row per `(alert, sequence)` is the spreadsheet-native way to encode a 1:N relation; the dropped `sequence_count` / `camera_count` columns are trivial `COUNTIF`s.
- **Two column renames** to make the meaning explicit:
  - `alert_lat` / `alert_lon` → `alert_triangulated_lat` / `alert_triangulated_lon` (pre-answers the "why is it empty?" question raised in the issue thread).
  - `sequence_azimuth` → `sequence_triangulated_azimuth` (the `Sequence` model also has `camera_azimuth`, the cone center; the prefix removes the ambiguity).

## Schema

Each row joins one alert with one of its sequences. Alert-level cells repeat across the alert's rows; sequences within an alert are sorted ASC by `started_at`.

| Section | Columns |
|---|---|
| Alert | `alert_id`, `alert_started_at_date`, `alert_started_at_time`, `alert_last_seen_at`, `alert_duration_seconds`, `alert_triangulated_lat`, `alert_triangulated_lon`, `organization_id` |
| Sequence | `sequence_id`, `sequence_started_at`, `sequence_last_seen_at`, `sequence_triangulated_azimuth`, `sequence_label`, `pose_id` |
| Camera | `camera_id`, `camera_name` |

`sequence_label` maps `AnnotationType` to a 3-bucket label: `WILDFIRE_SMOKE → wildfire`, `OTHER_SMOKE` / `OTHER → other`, `None → unknown`.

## Tests

Split into two clearly-labeled sections:

**Unit tests** for `_iter_alerts_csv` (sync, in-memory `Alert` / `Sequence`, no DB / HTTP / auth):
- header-only when no alerts
- null lat / lon → empty cells
- one row per sequence, ASC by `started_at`
- wildfire label mapping, **parametrized over all 4** `AnnotationType | None` cases (catches the previously-uncovered `AnnotationType.OTHER` mapping)
- `camera_name` resolved per sequence from `camera_names_by_id`

**Integration tests** for `GET /alerts/export` (route → JWT → DB):
- happy path (content-type, `attachment` disposition, full first-row dict equality)
- date window narrows correctly
- org isolation (non-admin user; cross-org alerts hidden)
- `to_date < from_date` → 422
- no token → 401

## Test plan

- [ ] CI runs `pytest`
- [x] `ruff 0.11.9 check` and `ruff 0.11.9 format --check` pass locally (repo pins 0.11.9 in pre-commit)
- [ ] Manual `curl` against a dev stack with a valid token; open the CSV in Sheets and confirm filter / pivot work on `sequence_label` and `camera_name`


MateoLostanlen (approved):
Hi @Acruve15 , sorry for the late review.

"A | in a camera_name silently breaks parsing" isn't a concern, the names are normalized

That said, for processing with code, one sequence per line is probably simpler. My only worry is that it might be misread by the firefighters. but OK to merge as is, we'll adjust if needed.

Thanks for the PR!

PR #615, opened 2026-06-10
feat: validate sequences with the temporal model before triangulation

MateoLostanlen:
Adds a temporal-model validation gate so a sequence is only triangulated and notified once it's *validated*. Validation is coordinated **through Postgres** (not in-memory state), so it is correct with any number of uvicorn workers.

### Pipeline
- Each detection marks its sequence **due for validation** (`sequences.validation_due_at` — an idempotent enqueue: one queue entry per sequence, FIFO, whichever worker receives the detection).
- A **validation worker** (one loop per uvicorn process, started in the lifespan) claims due sequences with `FOR UPDATE SKIP LOCKED` + a lease and runs the pipeline: **risk pre-gate** → **temporal model** (`POST /predict`, scoped to the union of the sequence's bboxes via `roi_xyxyn`) → persist score → claim → triangulate → notify. The model call never holds a DB session.
- **Frames are read at scoring time**, not at enqueue time: a sequence queued behind a backlog is scored with its freshest frame set, and frames arriving *during* a model call keep the job due so it re-runs.

### Window rules (4–10 distinct frames)
- `< 4` frames: wait (next detection retries). `probability > 0.45` validates.
- `> 10` frames **and already scored**: terminal (`window_exhausted`) — the model had its chances and declined; never resurrected by a later fail-open.
- `> 10` frames **never scored** (backlog, model freshly deployed): score the **last 10** anyway — a backlog delays sequences, it never silently drops them.

### Fail-open (explicit and traced via `sequences.validation_status`)
- Model not configured / unreachable / breaker open → validate on the risk gate alone (`fail_open_unavailable`).
- Queued past `TEMPORAL_VALIDATION_MAX_AGE` (default 5 min) → validate on the risk gate alone (`fail_open_stale`), bounding validation latency under a sustained backlog.
- A successful-but-scoreless response does **not** fail open.

### Circuit breaker
3 consecutive outage-like failures (network/5xx) open the breaker with **exponential backoff** (60s base, ×2 per re-trip, 1h cap, jitter) and a **half-open probe**: the first call after the pause closes the breaker on success or re-opens it immediately on failure. 4xx fail the call (fail-open) without tripping the breaker. No more client-side semaphore: the temporal API serializes inference server-side; with N uvicorn workers at most N calls are in flight (`N × model latency < TEMPORAL_API_TIMEOUT`).

### Reliability
- No duplicate model calls across workers (SKIP LOCKED + lease), the queue survives restarts (it's in the DB), a crashed worker's job is reclaimed when its lease expires, and job errors retry with a 30s backoff. Validation is claimed atomically; a failed alert attachment releases the claim and the worker retries.
- **Auth**: sends `Authorization: Bearer` when `TEMPORAL_API_TOKEN` is set (pyronear/temporal-model#35).

### Schema & settings
- `sequences`: `temporal_model_score`, `temporal_model_version`, `temporal_api_version`, `is_validated` (exposed) + `validation_due_at`, `validation_lease_until`, `validation_status` (internal, excluded from API responses) + partial index on due rows; existing rows backfilled to validated. The version pair (model release + serving image tag, reported by the temporal API since pyronear/temporal-model#48) is written with the score so stored scores stay attributable per model release, enabling offline evaluation against `is_wildfire` annotations across deployments.
- New: `TEMPORAL_API_URL`, `TEMPORAL_API_TOKEN`, `TEMPORAL_MODEL_THRESHOLD`, `TEMPORAL_API_TIMEOUT`, `TEMPORAL_VALIDATION_POLL_SECONDS`, `TEMPORAL_VALIDATION_MAX_AGE`, `TEMPORAL_VALIDATION_LEASE_SECONDS`.

Notes: **every notification channel (webhooks, Telegram, Slack) sits behind the validation gate** — a sequence the model declines never notifies anyone; notifications fire after the atomic validation claim with the sequence's latest detection, with at-least-once delivery (a worker dying mid-job is resumed by a sibling — the post-claim work is crash-safe and the queue survives restarts). Alert attachment is serialized per organization (advisory lock) so concurrent validations of the same event can't duplicate alerts. Expected load (~100–500 sequences/day, ≤7 calls each, ~9s/call) sits well under the model's ~400 calls/h serial throughput; the stale fail-open only matters in extreme correlated bursts.


src/app/services/temporal.py
```diff
@@ -0,0 +1,110 @@
...
+        when the call itself fails (network/HTTP), so callers can fail open.
+
+        Sends ``Authorization: Bearer`` when ``TEMPORAL_API_TOKEN`` is set (the temporal API
+        guards /predict with a shared token); unset matches a server with auth disabled.
+        """
+        host = (settings.TEMPORAL_API_URL or "").rstrip("/")
+        headers = {"Authorization": f"Bearer {settings.TEMPORAL_API_TOKEN}"} if settings.TEMPORAL_API_TOKEN else None
+        try:
+            # Serialize against the model's real throughput so bursts queue instead of timing out.
+            async with self._semaphore:
```
Chouffe: **Re: temporal API queueing under load — backpressure & queue-wait deadline**

The `asyncio.Semaphore(TEMPORAL_API_MAX_CONCURRENCY)` here is a clean way to serialize calls against a serially-inferring model, and serializing the burst instead of timing out is the right instinct. Two properties of the queue are worth making explicit before merge, since they only bite under sustained load:

1. **`TEMPORAL_API_TIMEOUT` doesn't cover the wait.** The timeout is on the `httpx` call, which starts only *after* the semaphore is acquired. Time parked on the semaphore is unbounded, so a task's total latency under a backlog can be far longer than 30s.
2. **The queue has no depth limit and no overall deadline.** Nothing caps how many coroutines park on the semaphore. This is cheap memory-wise (Phase 2 holds no DB session — nicely done), but with a slow model and a steady detection rate the backlog grows, and a sequence can cross `MAX_FRAMES` (10 distinct frames) *while still waiting its turn*. When it finally runs, `_temporal_verdict` hard-drops it `(False, None)` — so under load the gate silently sheds the very sequences it's queued to score, rather than the model declining them.

Neither is necessarily wrong for the expected detection volume — but it hinges on that volume staying below the model's throughput. What's the expected steady-state and burst detection rate vs. the model's per-call latency? If bursts can outrun it, an overall deadline (queue-wait + call) or a bounded queue with explicit fail-open-on-overflow would make the shed behavior intentional rather than emergent.

src/app/services/temporal.py
```diff
@@ -0,0 +1,110 @@
...
+        except (httpx.HTTPError, ValueError) as exc:
+            logger.warning("Temporal API call failed: %r", exc)
+            self._record_failure()
+            raise TemporalUnavailableError(str(exc)) from exc
+        self._record_success()
+        probability = data.get("probability") if isinstance(data, dict) else None
+        return float(probability) if isinstance(probability, (int, float)) else None
+
+
+temporal_service = TemporalModelService()
```
Chouffe, line 165: **Re: per-process semaphore & circuit breaker — deployment assumption**

`temporal_service` is a module-level singleton, so the semaphore and the circuit breaker are per-process. With a single uvicorn worker that's exactly "one call in flight" and one shared breaker — but `--workers N` or multiple replicas behind Traefik multiplies it to `N × TEMPORAL_API_MAX_CONCURRENCY` concurrent calls, with N independent breakers (one process can be paused while another keeps calling). The docstring half-notes this already; worth confirming the production launch command is single-worker, or stating the assumption explicitly so nobody scales it out without revisiting the limit. If multi-worker is ever needed, the concurrency cap and breaker would have to move server-side (or into shared state).

src/app/services/temporal.py
```diff
@@ -0,0 +1,110 @@
...
+    semaphore) is per-process, like the risk cache: with N uvicorn workers the effective
+    concurrency is ``N * TEMPORAL_API_MAX_CONCURRENCY``, so run a single worker for this path
+    or move the limit server-side if the model is strictly serial.
+    """
+
+    MIN_FRAMES: int = 4
+    MAX_FRAMES: int = 10
+    WINDOW: int = 6
+    MAX_CONSECUTIVE_FAILURES: int = 3
+    PAUSE_SECONDS: int = 4 * 3600
```
Chouffe: **Re: breaker pause is a fixed 4h blackout with no half-open probe**

The breaker correctly stops hammering a down API — after `MAX_CONSECUTIVE_FAILURES` it sets `_paused_until = now + PAUSE_SECONDS` (4h) and every call fails open instantly during the window, which protects the queue nicely. The thing worth confirming is the **recovery behavior**: `is_available()` only clears the pause once the full 4h has *elapsed* — there's no half-open probe before then. So if the API recovers a few minutes into the pause, the code keeps skipping it for the remaining ~4h, and every sequence in that window validates on the risk gate alone.

The cost of that blackout is precisely the thing this PR exists to prevent: for up to 4h after a transient blip, false positives the temporal model would have caught flow straight through to triangulation + Slack. A brief outage buys a long stretch of degraded gating. (The pause is also per-process/in-memory, so its real duration is `min(4h, time-to-next-restart)`.)

Is up to 4h of fail-open after a transient failure acceptable for this gate? If not, a couple of standard options:
- **Half-open probe:** during the pause, let *one* call through periodically (e.g. every `PAUSE_SECONDS / k`) instead of blocking the whole window. Recovery is then detected within minutes while a sustained outage still mostly fails open.
- **Shorter base pause + exponential backoff:** start at ~30–60s and grow on repeated re-trips (cap at some max). Transient blips recover fast; a real outage still backs off to a long quiet period.

Either keeps the "don't hammer a dead API" property while shrinking the recovery blind spot. If 4h is a deliberate choice for the expected failure profile, a one-line comment on `PAUSE_SECONDS` saying so would save the next reader the same question.

src/app/api/api_v1/endpoints/detections.py
```diff
@@ -489,43 +644,28 @@ async def create_detection(
...
+                notify_dets.append(det)
 
         created.append(det)
 
+    # Schedule validation/triangulation FIRST: Starlette aborts the chain on the first task
+    # that raises, and a failing notification must not skip the gated triangulation + Slack.
+    for seq_id, latest_det_id in affected_sequences.items():
+        background_tasks.add_task(validate_sequence, seq_id, latest_det_id, token_payload.organization_id)
+
+    # Best-effort notifications for newly created sequences (ungated, as before).
```
Chouffe: **Re: webhooks & Telegram fire on unvalidated sequences while Slack is gated**

Nice job moving Slack behind the validation claim — it fires once, only after the temporal model validates, keyed off `claim_validation`. One asymmetry worth confirming is intentional: when a new sequence is created, **webhooks and Telegram still fire immediately** (this `notify_dets` loop, ungated), whereas **Slack and triangulation now wait** for `validate_sequence`.

So a sequence the temporal model later declines (never reaches `is_validated`) has already emitted its webhook + Telegram alert, but will never produce an alert/Slack notification. The `# (ungated, as before)` comment shows it's deliberate, but the PR description's "only validated sequences are triangulated and notified" reads as covering all channels — it's accurate for Slack, not for webhooks/Telegram.

This is fine if webhooks/Telegram are intended as the raw detection firehose (webhooks feeding a platform that filters downstream, Telegram as an ops channel) and Slack is the curated human-alert channel. It's a problem if Telegram reaches the same responders who'd act on Slack, since the temporal gate is then bypassed for them and false positives still page people.

Two questions:
- Is the webhook/Telegram firehose intentional, or should all three channels sit behind the same gate?
- If intentional, can the PR description be tweaked so "notified" isn't read as covering every channel? If not, moving the webhook/Telegram scheduling into `_run_sequence_validation` alongside `_notify_slack_for_sequence` (keyed off the same `claim_validation` win) is the mechanical fix.

src/migrations/versions/2026_06_10_1000-c5e2f7a8b1d0_add_temporal_validation_to_sequences.py
```diff
@@ -0,0 +1,39 @@
...
+"""
+
+import logging
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "c5e2f7a8b1d0"
```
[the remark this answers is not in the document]
MateoLostanlen: False positive: Alembic reads this module global by introspection to build the revision map — same pattern as every migration in this repo. Resolving.

src/migrations/versions/2026_06_10_1000-c5e2f7a8b1d0_add_temporal_validation_to_sequences.py
```diff
@@ -0,0 +1,39 @@
...
+down_revision: Union[str, None] = "7f1c4d2a9b3e"
```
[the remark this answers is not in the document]
MateoLostanlen: False positive: Alembic reads this module global by introspection to build the revision map — same pattern as every migration in this repo. Resolving.

src/migrations/versions/2026_06_10_1000-c5e2f7a8b1d0_add_temporal_validation_to_sequences.py
```diff
@@ -0,0 +1,39 @@
...
+branch_labels: Union[str, Sequence[str], None] = None
```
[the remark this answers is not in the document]
MateoLostanlen: False positive: Alembic reads this module global by introspection to build the revision map — same pattern as every migration in this repo. Resolving.

src/migrations/versions/2026_06_10_1000-c5e2f7a8b1d0_add_temporal_validation_to_sequences.py
```diff
@@ -0,0 +1,39 @@
...
+depends_on: Union[str, Sequence[str], None] = None
```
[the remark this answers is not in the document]
MateoLostanlen: False positive: Alembic reads this module global by introspection to build the revision map — same pattern as every migration in this repo. Resolving.

src/app/main.py
```diff
@@ -69,17 +70,22 @@ async def _risk_refresh_loop() -> None:
...
     finally:
-        if task is not None:
-            task.cancel()
-            with contextlib.suppress(asyncio.CancelledError):
-                await task
+        for task in (risk_task, validation_task):
+            if task is not None:
+                task.cancel()
+                with contextlib.suppress(asyncio.CancelledError):
+                    await task
```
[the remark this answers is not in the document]
MateoLostanlen: False positive: `await task` after `task.cancel()` is the standard asyncio shutdown pattern — it lets the cancelled task actually unwind before the process exits, with the CancelledError suppressed by the surrounding context manager. Removing it would leak the task at shutdown. Resolving.

src/app/services/temporal.py
MateoLostanlen: Good catch! this pushed a redesign. The queue now lives in Postgres: each detection marks its sequence due (one entry per sequence), a per-process worker claims jobs via FOR UPDATE SKIP LOCKED + lease, and frames are read at scoring time. A backlog delays sequences but never drops them (past-window never-scored sequences score their last 10 frames), and jobs queued past TEMPORAL_VALIDATION_MAX_AGE fail open explicitly, traced as validation_status=fail_open_stale. For scale: ~100–500 sequences/day at ≤7 calls each vs ~400 calls/h of model throughput, the queue should stay empty outside extreme bursts.

src/app/services/temporal.py
MateoLostanlen: You were right to push, production runs 2 uvicorn workers, so the in-memory design was wrong. All load-bearing state moved to the DB (claim + lease → no duplicate calls, queue survives restarts, a dead worker's job is resumed at lease expiry). The model itself is protected server-side (the temporal API serializes inference); the client semaphore is gone. The breaker stays per-process, which is acceptable and documented.

src/app/services/temporal.py
MateoLostanlen: Implemented your suggestion: exponential backoff (60s base, ×2 per re-trip, 1h cap, jitter) with a half-open probe — recovery is detected within ~1 min instead of a 4h blind spot. Also: 4xx responses close the breaker (the API answered, it's reachable) while still failing open the call

src/app/api/api_v1/endpoints/detections.py
MateoLostanlen: Honestly I'd skipped it because we don't use webhooks/Telegram today and meant to align them later, but better to have it clean now. All three channels sit behind the validation gate, fired by the worker after the atomic claim. A declined sequence never notifies anyone. PR description updated.

src/app/services/validation.py
```diff
@@ -0,0 +1,385 @@
...
+        organizations = OrganizationCRUD(session)
+        webhooks = WebhookCRUD(session)
+        sequence_ = cast(Sequence, await sequences.get(sequence_id, strict=True))
+        camera = cast(Camera, await cameras.get(sequence_.camera_id, strict=True))
+        organization_id = camera.organization_id
+        # Two workers validating two sequences of the same event concurrently must not
+        # both create an alert: serialize attachment per organization.
+        async with _organization_alert_lock(organization_id):
+            alert_id = await _attach_sequence_to_alert(sequence_, camera, cameras, sequences, alerts)
+        await _notify_for_sequence(sequence_, camera, organization_id, detections, organizations, webhooks, alert_id)
```
Chouffe: **Re: notifications block the next sequence's scoring**

The validation worker is strictly serial — one job start-to-finish — and `_notify_for_sequence` runs *before* `finish_validation_job` (line 347), awaiting webhooks (each up to 5s via `dispatch_webhook`), Telegram, then Slack, sequentially. So no other sequence can be claimed or scored until all of this sequence's notifications complete. A slow Slack hook or a hanging webhook endpoint directly throttles fire validation for every other camera. This also shrinks job wall-time under the 120s lease, reducing the multi-worker re-claim/double-notify window.

Notifications need neither the model nor the lease, and they're *already* best-effort (each wrapped in try/except, never individually retried), so there's no durability guarantee to protect. Complete the job first, then fire them off the critical path.

**Example.** Make `_notify_for_sequence` own its session (so it can outlive the worker's session), then detach it:

```python
# _notify_for_sequence now takes ids and opens its own session
async def _notify_for_sequence(sequence_id: int, organization_id: int, alert_id: Optional[int]) -> None:
    async with session_factory() as session:
        sequences      = SequenceCRUD(session)
        detections     = DetectionCRUD(session)
        organizations  = OrganizationCRUD(session)
        webhooks       = WebhookCRUD(session)
        sequence_ = await sequences.get(sequence_id)
        camera    = await CameraCRUD(session).get(sequence_.camera_id)
        ...  # webhook / telegram / slack, exactly as today

# in _complete_validated_sequence:
    async with _organization_alert_lock(organization_id):
        alert_id = await _attach_sequence_to_alert(...)
    await sequences.finish_validation_job(sequence_id)        # <- complete FIRST, worker is free

# outside the session block — fire-and-forget, off the critical path:
_pending.add(t := asyncio.create_task(_notify_for_sequence(sequence_id, organization_id, alert_id)))
t.add_done_callback(_pending.discard)   # keep a ref so the task isn't GC'd, surface errors
```

(`_pending: set[asyncio.Task] = set()` at module level.) If you want delivery to survive a crash, the heavier alternative is a second `notification_due_at` marker drained by the same worker — but since these are already best-effort, detaching is the cheap, faithful fix.

src/app/services/validation.py
```diff
@@ -0,0 +1,385 @@
...
+        sequence_ = await SequenceCRUD(session).claim_due_validation(settings.TEMPORAL_VALIDATION_LEASE_SECONDS)
+    if sequence_ is None:
+        return False
+    sequence_id = sequence_.id
+    try:
+        await _process_claimed_sequence(sequence_id)
+    except Exception:
+        logger.exception("Sequence validation failed for sequence %s; will retry", sequence_id)
+        async with session_factory() as session:
+            await SequenceCRUD(session).release_validation_lease(sequence_id, retry_in_seconds=RETRY_DELAY_SECONDS)
```
Chouffe: **Re: errored jobs retry forever with no cap**

On any exception this releases the lease but keeps the job due with a 30s backoff — no attempt counter, no terminal failure state — so a persistently failing job retries every 30s **indefinitely**, logging each time, and (serial worker) steals a slot from healthy jobs on every spin. Note `TEMPORAL_VALIDATION_MAX_AGE` does **not** save you here: `release_validation_lease(retry_in_seconds=...)` pushes `validation_due_at` to `now + 30s` on each error, so the staleness fail-open never fires for an error loop — that bound is on queue *wait*, not retry *attempts*.

Add an attempts cap that lands in a terminal state.

**Example.** One new column + a CRUD method, and extend the enqueue guard:

```python
# migration + model
validation_attempts: int = Field(default=0, nullable=False, exclude=True)

# crud_sequence.py
async def fail_or_retry_validation(self, sequence_id: int, *, max_attempts: int, retry_in_seconds: float) -> None:
    seq = cast(Sequence, await self.get(sequence_id, strict=True))
    attempts = (seq.validation_attempts or 0) + 1
    if attempts >= max_attempts:        # dead-letter: stop retrying
        values = {"validation_attempts": attempts, "validation_status": "failed",
                  "validation_due_at": None, "validation_lease_until": None}
        logger.error("Sequence %s failed validation %d times; giving up", sequence_id, attempts)
    else:                               # transient: back off and retry
        values = {"validation_attempts": attempts, "validation_lease_until": None,
                  "validation_due_at": utcnow() + timedelta(seconds=retry_in_seconds)}
    await self.session.exec(update(Sequence).where(cast(Any, Sequence.id) == sequence_id).values(**values))
    await self.session.commit()
```

```python
# process_next_due_validation — the error path becomes:
except Exception:
    logger.exception("Sequence validation failed for sequence %s", sequence_id)
    async with session_factory() as session:
        await SequenceCRUD(session).fail_or_retry_validation(
            sequence_id, max_attempts=5, retry_in_seconds=RETRY_DELAY_SECONDS,
        )
```

```python
# enqueue_validation — exclude the new terminal state too:
.where(status_col.not_in(("window_exhausted", "failed")))
```

Two small follow-ups: reset `validation_attempts = 0` in `finish_validation_job` so the cap counts *consecutive* failures rather than lifetime; and the fact that this wants a new column is itself a nudge toward a dedicated `validation_jobs` table instead of more fields on `sequences`.

src/app/crud/crud_sequence.py
```diff
@@ -33,3 +36,130 @@ async def bump_max_conf(self, sequence_id: int, candidate: float) -> None:
...
+        sequence already queued is NOT re-queued (one entry per sequence, whichever worker
+        receives the detection). No-op for validated or window-exhausted sequences.
+        """
+        status_col = cast(Any, Sequence.validation_status)
+        due_col = cast(Any, Sequence.validation_due_at)
+        stmt: Any = (
+            update(Sequence)
+            .where(cast(Any, Sequence.id) == sequence_id)
+            .where(cast(Any, Sequence.is_validated).is_(False))
+            .where(status_col.is_distinct_from("window_exhausted"))
```
Chouffe: **Re: `validation_status` values are duplicated as bare strings**

The terminal-state guard in `enqueue_validation` compares against the literal `"window_exhausted"`, while `validation.py` defines `WINDOW_EXHAUSTED = "window_exhausted"` as the constant everything else uses. Same value, two sources of truth, no DB-level `CHECK`/enum — so a typo in either silently breaks the guard and window-exhausted (and, with the poison-cap comment, `failed`) sequences re-enqueue forever. Suggest referencing the shared constants in the query (`.not_in((WINDOW_EXHAUSTED, FAILED))`) and/or a `CHECK (validation_status IN (...))` constraint so the state set is enforced in one place rather than by convention.

Chouffe:
**Architectural note (non-blocking): the queue is a hand-built job runner — worth knowing the off-the-shelf options**

The DB-coordinated validation worker is a solid, correct implementation of the classic Postgres-queue pattern — `FOR UPDATE SKIP LOCKED` + a lease + a partial index, with crash recovery via lease expiry. No notes on the mechanics. This is more of a "where does this go next" thought.

Right now the queue is **validation-specific**: the job state (`validation_due_at`, `validation_lease_until`, `validation_status`) lives as columns on `sequences`. If a second async job type ever shows up (re-scoring, cleanup, decoupled notifications, etc.), that shape doesn't generalize — you'd add another set of columns to another domain table. The generic version is a dedicated table:

```
jobs(id, kind, dedup_key, payload jsonb, status,
     run_after, lease_until, attempts, max_attempts, last_error, created_at)
  -- partial index on (run_after) WHERE status='pending'      -> cheap claim scan
  -- unique index on (kind, dedup_key) WHERE status='pending' -> one active job per target (coalescing)
```

A generic worker then claims by `run_after` and dispatches on `kind` to a registered handler; `payload` carries ids only (state is re-read fresh in the handler, preserving the "score against the freshest frames" property this PR already has). This shape also hands you, for free, the two things the inline comments ask for: `attempts`/`max_attempts`/`dead` *is* the poison-job cap, and a separate `kind="notify"` job moves notifications off the validation critical path.

The punchline: **a generic Postgres job table is exactly what [pgqueuer](https://github.com/janbjorge/pgqueuer) (and [procrastinate](https://github.com/procrastinate-org/procrastinate)) already provide** — async-native, `asyncpg`-based, `SKIP LOCKED` + `LISTEN/NOTIFY` (which also removes the current ~2s poll latency), with retries/backoff, dead-letter, and dedup built in. Same single-dependency footprint we have today — no new infra. So if multiple async job types are on the horizon, adopting one of these is likely less code to own than growing this hand-built runner; if validation stays the only job, what's here is perfectly reasonable to keep.

Not asking to change anything in this PR — just flagging the inflection point so it's a deliberate choice later rather than an accumulation of job-state columns.

src/app/services/validation.py
MateoLostanlen: Implemented as suggested: the job completes first, then notifications fire as a detached task (strong refs kept, errors logged), so a hanging channel no longer delays the next sequence's scoring. Trade-off documented: a crash between completion and delivery loses the notification, which is consistent with the channels being best-effort; the durable artifact is the alert row.

src/app/services/validation.py
MateoLostanlen: Implemented: `validation_attempts` counts consecutive errors (reset on every completed job) and at 5 the job dead-letters as `validation_status='failed'`, terminal and excluded from enqueue. On the `validation_jobs` table nudge: agreed in principle, but at our volume (~100-500 seq/day) columns on `sequences` keep the moving parts minimal; worth revisiting if the job state grows further.

src/app/crud/crud_sequence.py
MateoLostanlen: Done: status values now live in `app.models` as the single source of truth (worker and CRUD guards import the same constants, including a `TERMINAL_VALIDATION_STATUSES` tuple), and the migration adds a `CHECK` constraint so the value set is enforced in the DB.

MateoLostanlen:
> Architectural note (non-blocking): the queue is a hand-built job runner

Agreed on the inflection point, and thanks for naming it explicitly. Keeping the hand-built runner is a deliberate choice for this PR: validation is the only job kind, the volume is tiny (~100-500 sequences/day), and the job state is naturally 1:1 with a sequence, so columns on `sequences` stay the smallest thing that works. The two features the generic table would have given us for free (attempts/dead-letter cap, notifications off the critical path) landed inline in the latest commits.

The rule going forward: the day a second async job kind shows up (re-scoring, cleanup, durable notifications), we adopt pgqueuer or procrastinate instead of growing more job columns on domain tables. The LISTEN/NOTIFY latency win over the 2s poll would come with it. Noted in the team backlog so it stays a deliberate choice.

Chouffe (approved):
LGTM! Awesome work! Let's 🚢 it!

PR #616, opened 2026-06-10
fix(detections): map each crop to its own bbox

MateoLostanlen:
- `POST /detections/` accepted N bboxes (creating N detections) but a single `crop`, stamping the same `crop_bucket_key` onto every detection — so with multiple objects the one crop was misattributed. It now accepts one crop per bbox, validates the crop count matches the bbox count before any S3 upload (422 otherwise), and maps each crop to its bbox by index.
- The `crop` multipart field name is kept, so callers sending a single crop still work; crop keys are now prefixed by bbox index so byte-identical crops stay independently deletable.
- **Breaking (Python client):** `Client.create_detection`'s `crop: bytes` argument is replaced by `crops: list[bytes]` (raises `ValueError` on a length mismatch). The HTTP API stays backward compatible.
- Tests: multi-bbox/multi-crop alignment, crop/bbox count mismatch (422, no orphan upload), and client-side alignment + mismatch cases.

### DIFF
--- a/client/pyroclient/client.py
+++ b/client/pyroclient/client.py
@@ -349,7 +349,7 @@ class Client:
         media: bytes,
         bboxes: List[Tuple[float, float, float, float, float]],
         pose_id: int,
-        crop: bytes | None = None,
+        crops: List[bytes] | None = None,
     ) -> Response:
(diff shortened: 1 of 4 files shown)


fe51 (approved):
LGTM thanks !

PR #624, opened 2026-06-12
feat(detections): accept empty bboxes and keep sequence frames continuous

MateoLostanlen:
Devices had to send a fake placeholder bbox `(0.0,0.0,0.0001,0.0001,0.0)` for frames with no detection, which created phantom sequences (placeholders overlap each other and seed a sequence after 3 frames).

- `bboxes="[]"` is now valid on `POST /detections`: the frame is attached with an empty bbox ("continuity detection") to every sequence of the pose seen within `SEQUENCE_CONTINUITY_SECONDS` (default 120s) that no real bbox of the request matched; with no active sequence, nothing is stored (204). This also keeps a frame in a sequence whose smoke faded while another object on the same frame is still detected (multi-bbox case).
- Continuity rows never refresh `last_seen_at` / `max_conf` and never seed sequences, but re-enqueue temporal validation; spatial matching, notifications and per-sequence counts use the last non-empty bbox only.
- Deleting a detection no longer removes the S3 frame object while sibling rows (multi-bbox, continuity) still reference it.
- `pyroclient.create_detection` accepts an empty bbox list.


src/app/services/sequence_counts.py
```diff
@@ -10,15 +10,18 @@
...
 
 async def get_detection_counts_by_sequence_ids(session: AsyncSession, sequence_ids: List[int]) -> Dict[int, int]:
     if not sequence_ids:
         return {}
 
+    # Continuity rows (empty bbox) carry a frame, not a detection: don't count them.
     stmt: Any = (
         select(cast(Any, Detection.sequence_id), func.count(cast(Any, Detection.id)))
         .where(cast(Any, Detection.sequence_id).in_(sequence_ids))
+        .where(cast(Any, Detection.bbox) != EMPTY_BBOXES)
```
fe51: Why not counting them ? ? 
detections_count under-reports frames, it is likely to break image loading on the front-end ! 

This filter (bbox != EMPTY_BBOXES) makes detections_count exclude continuity rows, but:

- Notifications don't use this count. The notification fix in a9d0c65 only changed the image selection (validation.py → get_latest_with_bbox). Excluding empty bboxes from the count was bundled into the same commit but serves nothing on the backend — > get_detection_counts_by_sequence_ids is consumed only by the API serializers → the frontend.

- The frontend uses detections_count as the number of images/detections to load, and GET /sequences/{id}/detections returns continuity rows too (by design —> see its docstring). So the count is now smaller than the list it's meant to size: pagination stops early and frames get missed.

The count and the list must agree. Suggest dropping the EMPTY_BBOXES filter here so detections_count matches what /detections actually returns. If a "real detections only" count is ever needed, expose it as a separate field rather than overloading this one.

src/app/core/config.py
```diff
@@ -70,6 +70,9 @@ def sqlachmey_uri(cls, v: str) -> str:
     SEQUENCE_RELAXATION_SECONDS: int = int(os.environ.get("SEQUENCE_RELAXATION_SECONDS") or 120 * 60)
     SEQUENCE_MIN_INTERVAL_DETS: int = int(os.environ.get("SEQUENCE_MIN_INTERVAL_DETS") or 3)
     SEQUENCE_MIN_INTERVAL_SECONDS: int = int(os.environ.get("SEQUENCE_MIN_INTERVAL_SECONDS") or 5 * 60)
+    # Window after a sequence's last real detection during which a frame with no matching bbox
+    # is still attached to it (with an empty bbox) to keep the frame timeline continuous.
+    SEQUENCE_CONTINUITY_SECONDS: int = int(os.environ.get("SEQUENCE_CONTINUITY_SECONDS") or 2 * 60)
```
fe51, line 87: This PR adds `SEQUENCE_CONTINUITY_SECONDS` next to two existing windows (`RELAXATION`, `MIN_INTERVAL`) that gate sequence behaviour on different timescales, with no note on how they interact. Two small asks:

**1. Document the interaction.** A short comment on each one's role (opportunity to to it in this PR), and — the part that actually matters — that they're tuned **against the camera's frame interval** (continuity must span a few frames to ever fire; `MIN_INTERVAL_DETS` over `MIN_INTERVAL_SECONDS` implies an expected cadence). 

That's the relationship people get wrong, not the values in isolation.


**2. A light sanity check, right here in `config.py`** via a `@model_validator(mode="after")`:

```python
@model_validator(mode="after")
def _check_sequence_windows(self) -> "Settings":
    for name in ("SEQUENCE_RELAXATION_SECONDS", "SEQUENCE_MIN_INTERVAL_SECONDS", "SEQUENCE_CONTINUITY_SECONDS"):
        if getattr(self, name) <= 0:
            raise ValueError(f"{name} must be > 0")
    if self.SEQUENCE_CONTINUITY_SECONDS > self.SEQUENCE_RELAXATION_SECONDS:
        logger.warning(
            "SEQUENCE_CONTINUITY_SECONDS (%s) > SEQUENCE_RELAXATION_SECONDS (%s): "
            "continuity frames attach past the matchable window.",
            self.SEQUENCE_CONTINUITY_SECONDS, self.SEQUENCE_RELAXATION_SECONDS,
        )
    return self
```
Positivity is a hard `raise` (a zero/negative duration is always a bug); the ordering is a `warning`, not a raise — it still produces valid data, and the windows aren't monotonic by design, so no hard ordering assert.

(if too complexe, juste positivity and SEQUENCE_CONTINUITY_SECONDS < SEQUENCE_RELAXATION_SECONDS thats all)

src/app/models.py
```diff
@@ -83,6 +84,12 @@ class OcclusionMask(SQLModel, table=True):
 
 class Detection(SQLModel, table=True):
     __tablename__ = "detections"
+    __table_args__ = (
```
fe51: `ix_detections_bucket_key` belongs here — the "delete the shared frame object only with its last row" logic is new to this PR and needs it. But `ix_detections_sequence_id_created_at` and `ix_sequences_camera_pose_last_seen` are new indexes over queries that already existed on `main` and are unrelated to the empty-bbox feature. 

Suggest moving those two into a dedicated perf PR: keeps this PR's story tight, and lets the index change (which takes a write-blocking `CREATE INDEX` lock) be reviewed and, if needed, reverted on its own.
what do you think ?

fe51 (changes requested):
Thanks for the PR,

Started to review, few questions related to  some code block that will, either introduce a bug or are not related to the PR.

Happy to discuss it :)

src/app/services/sequence_counts.py
MateoLostanlen: Good catch, you're right: this count only feeds the serializers and must size what GET /sequences/{id}/detections actually returns, continuity rows included. Filter dropped in 25d33d8, with a comment stating that constraint. If a "real detections only" count is ever needed we'll expose it as a separate field

src/app/core/config.py
MateoLostanlen: Done in 2fb73c4: each window now has a comment on its role and the fact that they're all tuned against the camera frame interval, plus the model_validator you suggested (hard raise on non-positive values, warning when CONTINUITY > RELAXATION). Covered by a new test_config.py

src/app/models.py
MateoLostanlen: You're right, let's separate. I actually moved all three out, bucket_key included: after measuring at prod scale, the PR runs fine without them (~10 ms overhead per request vs main). Opened [#663](https://github.com/pyronear/pyro-api/issues/663) with the measurements to add them in a dedicated PR with CREATE INDEX CONCURRENTLY

fe51 (approved):
Hi @MateoLostanlen, thanks for taking feedback into account !

iI think it is okay to merge now (and test it in preprod for a while)

PR #633, opened 2026-07-11
feat(api): let admins fetch alerts and sequences across organizations

MateoLostanlen:
- Alert and sequence list endpoints (`unlabeled/latest`, `all/fromdate`, `alerts/export`) were hard-scoped to the caller's organization; admins now see all organizations, matching the existing behavior of the camera and detection listings.
- The FWI risk-score filter is skipped for admins (it is resolved per organization); the `risk_score` override param is ignored for them and documented as such.
- Non-admin behavior is unchanged and covered by updated risk-filter tests plus new cross-organization tests.

### DIFF
--- a/src/app/api/api_v1/endpoints/alerts.py
+++ b/src/app/api/api_v1/endpoints/alerts.py
@@ -222,11 +222,13 @@ async def export_alerts_csv(
 
     stmt: Any = (
         select(Alert)
-        .where(Alert.organization_id == token_payload.organization_id)
         .where(Alert.started_at >= start_dt)
         .where(Alert.started_at <= end_dt)
         .order_by(Alert.started_at.asc())  # type: ignore[attr-defined]
     )
+    # Admins export every organization's alerts
+    if UserRole.ADMIN not in token_payload.scopes:
+        stmt = stmt.where(Alert.organization_id == token_payload.organization_id)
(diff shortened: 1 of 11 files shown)


Acruve15:
@MateoLostanlen need a hand? 😊

MateoLostanlen:
@Acruve15 yes can you review this ? 😄

Acruve15 (dismissed):
**Non-blocking follow-up: factor the admin check into `TokenPayload.is_admin`**

We repeat `UserRole.ADMIN` in `token_payload.scopes` multiple times across the endpoints layer, and this PR adds a few more. Since `TokenPayload` already owns scopes, a property gives us one spelling everywhere and reads better:

```python
# src/app/schemas/login.py
class TokenPayload(BaseModel):
    sub: int = Field(..., gt=0)
    scopes: List[Role] = Field([], description="scopes of the token")
    organization_id: int = Field(..., gt=0)

    @property
    def is_admin(self) -> bool:
        return Role.ADMIN in self.scopes
```

Your call ;)

MateoLostanlen:
Good idea @Acruve15 , done in aa1ece5: added TokenPayload.is_admin and replaced all the scope checks across the endpoints. Also fixed the asserts flagged by the bot. Can you take another look? 🙏

Acruve15 (approved):
neat 👌

PR #640, opened 2026-07-20
Skip s3 obj existence

### LINKED ISSUE #639: Skip the S3 `head_object` existence check when presigning detection image URLs
# Description

## Problem

When serving detection image URLs (most notably `GET /sequences/{id}/detections`), we
run one synchronous S3 `head_object` request per file before generating the presigned
URL. See `get_public_url` in `src/app/services/storage.py:70-75`, called from
`src/app/api/api_v1/endpoints/sequences.py:98-99`.

Two issues make this expensive:

- **Blocking + serial.** `check_file_existence` is synchronous boto3 running inside an
  `async` endpoint, so each `head_object` blocks the event loop for a full S3
  round-trip, one after another. A sequence with the default `limit=10` does ~10 serial
  head requests; up to 100 with `limit=100` (× crops when `with_crop=true`).
- **It's the only network cost.** Presigned URL generation itself is a local signature
  computation and does no network I/O. The `head_object` is the sole reason these
  endpoints touch S3 at all.

## Order of magnitude

An S3 `head_object` is typically ~20-50 ms. So the existence check alone adds roughly:

- `limit=10` → ~0.2-0.5 s
- `limit=100` → ~2-5 s

...on top of the response, purely to verify files that in practice always exist (no
missing-file case has occurred to date).

## Proposal

Skip the existence check and presign directly. If a file were ever missing, the client
would get a `403`/`404` from S3 when loading the URL, instead of an upfront `404` from
the API. Given the missing-file case has never happened, this looks like a good trade.

This mirrors the existing `with_crop` flag, which already defaults to `false` "to skip
the extra S3 head requests when crops are not needed"
(`src/app/api/api_v1/endpoints/sequences.py:71-74`) — we'd just extend the same
reasoning to the main image.

## Options

- Remove the `head_object` check from `get_public_url` outright => sounds the more logic one
- Add a `verify_exists=False` parameter so we only skip it on the hot paths (sequence
  detections), keeping strict behaviour elsewhere if wanted.

fe51:
Fixes #639  (and will help to implement frontend player with better performances)

# About implementation decisions

## Why not just remove the check globally

The check is not dead weight everywhere. The **camera** endpoints rely on it as a
contract: `last_image` points to a single mutable/rotating image that can legitimately be
gone, so `get_public_url` raising a 404 is caught and turned into `last_image_url: null`
(`cameras.py` `get_camera` / `fetch_cameras`). Removing the check globally breaks that —
and indeed broke `test_get_camera_s3_unavailable_returns_null_url` and
`test_fetch_cameras_s3_unavailable_returns_null_url`.

So there are two real options.

## Option A — change the contract (drop the check everywhere)

Remove the existence check entirely; endpoints always return a URL. If a file is missing,
the **frontend** handles the broken/expired URL (S3 returns 403/404 on load).

- **Pros:** simplest server code; no per-call flag; the frontend already has to handle
  expired presigned URLs anyway.
- **Cons:** changes the camera contract — `last_image_url` would no longer be `null` for a
  missing image, so the frontend must detect the broken image and the camera endpoints +
  their tests must be reworked. Error moves from back to front for every image type.

## **Option B — opt-out flag (`verify_exists`) -> **chosen****

Add `verify_exists: bool = True` to `get_public_url`. Default keeps today's behaviour
(cameras unchanged); hot paths pass `verify_exists=False` (sequence detections + detection
url) to skip the `head_object`.

- **Pros:** breaks the least — camera contract and its tests stay intact; the perf win
  lands exactly where the image always exists; localized, low-risk change.
- **Cons:** one extra parameter to reason about; the "verify or not" decision now lives at
  each call site rather than being uniform.

## Decision

Going with **Option B**: after reflection it is the cleanest — it delivers the perf win on
the hot path without forcing a contract change and frontend rework on cameras. Option A
stays on the table if we later decide the frontend should own missing-image handling
uniformly.

Happy to discuss it !

### DIFF
--- a/src/app/api/api_v1/endpoints/detections.py
+++ b/src/app/api/api_v1/endpoints/detections.py
@@ -647,8 +647,10 @@ async def get_detection_url(
     if UserRole.ADMIN not in token_payload.scopes and token_payload.organization_id != camera.organization_id:
         raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail="Access forbidden.")
     bucket = s3_service.get_bucket(s3_service.resolve_bucket_name(camera.organization_id))
-    crop_url = bucket.get_public_url(detection.crop_bucket_key) if detection.crop_bucket_key else None
-    return DetectionUrl(url=bucket.get_public_url(detection.bucket_key), crop_url=crop_url)
+    crop_url = (
+        bucket.get_public_url(detection.crop_bucket_key, verify_exists=False) if detection.crop_bucket_key else None
+    )
+    return DetectionUrl(url=bucket.get_public_url(detection.bucket_key, verify_exists=False), crop_url=crop_url)
 
 
 @router.get("/", status_code=status.HTTP_200_OK, summary="Fetch all the detections")
--- a/src/app/api/api_v1/endpoints/sequences.py
+++ b/src/app/api/api_v1/endpoints/sequences.py
@@ -95,8 +95,12 @@ async def fetch_sequence_detections(
     return [
         DetectionWithUrl(
             **DetectionRead(**elt.model_dump()).model_dump(),
-            url=bucket.get_public_url(elt.bucket_key),
-            crop_url=(bucket.get_public_url(elt.crop_bucket_key) if with_crop and elt.crop_bucket_key else None),
+            url=bucket.get_public_url(elt.bucket_key, verify_exists=False),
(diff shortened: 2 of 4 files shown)


MateoLostanlen (approved):
LGTM, thanks @fe51 this should speed up our platform !

PR #641, opened 2026-07-21
feat(alerts): add count endpoints for fromdate and unlabeled/latest

### LINKED ISSUE #636: Add alerts count to `/api/v1/alerts/all/fromdate`  and `/api/v1/alerts/unlabeled/latest` response
## Pitch
The `/api/v1/alerts/all/fromdate` endpoint returns the alert data for the current day. 
Similarly `/api/v1/alerts/unlabeled/latest` endpoint returns unlabeled alerts data 

The response does not include the total number of matching alerts : alerts_count

Without the total count, the frontend cannot implement pagination efficiently or determine the total number of available pages.

## Proposed Solution
Update endpoints response to include the total number of alerts that match the request criteria : alerts_count

Happy to discuss it :)

Acruve15:
## What

Adds two dedicated count endpoints so the frontend can paginate the alert lists:

- `GET /api/v1/alerts/all/fromdate/count` (params: `from_date`, `risk_score`)
- `GET /api/v1/alerts/unlabeled/latest/count` (param: `risk_score`)

Each returns `{"count": n}`, the number of alerts matching exactly the same filters as its list counterpart, minus `limit`/`offset`.

## Why

Closes #636. `GET /alerts/all/fromdate` and `GET /alerts/unlabeled/latest` return paginated lists but no total, so the frontend cannot compute the number of pages. Rather than embed a count in the list responses (a breaking shape change), these are separate, non-breaking endpoints. They accept the same filtering params (`from_date`, `risk_score`) so the count always matches what the list will show under those filters.

## How

- Extracted the filter-building for each endpoint into shared `_build_unlabeled_alerts_stmt` / `_build_alerts_from_date_stmt` helpers. The list and count handlers for a given endpoint call the same builder, so the count cannot drift from the list.
- Added a `_serialize_alerts_page` helper that removes the previously duplicated list-materialization tail (query exec, sequence fetch, detection counts, serialize) shared by both list handlers.
- Added an `AlertCount` response schema.

## Testing

- New endpoint tests: happy-path counts, `count == len(list)` consistency under the same params, cross-org isolation, empty date, and unauthenticated (401).
- Added a parity test in `test_risk_filter.py` verifying count and list agree while a `risk_score` filter is active.
- Full suite green (608 passed, 1 skipped); `ruff` and `ty` clean.

### DIFF
--- a/src/app/api/api_v1/endpoints/alerts.py
+++ b/src/app/api/api_v1/endpoints/alerts.py
@@ -7,7 +7,7 @@
 import csv
 import io
 from datetime import date, datetime, time, timedelta
-from typing import Any, Dict, Iterable, Iterator, List, Union, cast
+from typing import Any, Dict, Iterable, Iterator, List, Tuple, Union, cast
 
 from fastapi import APIRouter, Depends, HTTPException, Path, Query, Security, status
 from fastapi.responses import StreamingResponse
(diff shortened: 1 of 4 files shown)


fe51 (approved):
Hi @Acruve15 thanks a lot for the PR ! 

Indeed, your approach with dedicated counts endpoints is relevant 🧠 
Because : 
- The frontend will fetch the count once per filter-set (when from_date/risk_score changes), cache it, then pages freely hitting only the list endpoint. That matches how pagination actually behaves.
- Hence it avoids to recompute the count on every page fetch ( even though the total only changes when the filters change, not when you turn a page).
- unlabeled/counts will be useful elsewhere ;) in https://github.com/pyronear/new-pyro-platform/pull/229 (more efficient that requesting unlabeled? )
- no braking change in api contract/shape


That being said, code is clean & tested in dev env working like a charm, lets go

PR #650, opened 2026-07-22
fix(tests): make validation queue drain deterministic under CI load

MateoLostanlen:
- The claim-until-empty drain loop in test_alerts and test_detections exits as soon as no row is claimable, but a retry backoff, a leftover lease, or a runner clock step can leave a job pending and not claimable, silently skipping a sequence's triangulation (flaky `test_triangulation_creates_single_alert`, e.g. CI run 29149066355 on PR #633).
- Adds a shared `drain_validation_queue()` test helper that backdates pending due times and clears leases before each claim, stopping only once the queue is provably empty, with a deadline and per-sequence diagnostics.
- Test-only change; also adds an explicit "all sequences validated" assert to pinpoint future failures.

### DIFF
--- a/src/tests/conftest.py
+++ b/src/tests/conftest.py
@@ -1,25 +1,29 @@
 import asyncio
 import io
 import os
-from datetime import datetime
-from typing import AsyncGenerator, Dict
+import time
+from datetime import datetime, timedelta
+from typing import Any, AsyncGenerator, Dict, cast
 
 import pytest
 import pytest_asyncio
 import requests
 from botocore.exceptions import ClientError
 from httpx import ASGITransport, AsyncClient
+from sqlalchemy import update
 from sqlalchemy.orm import sessionmaker
-from sqlmodel import SQLModel, text
+from sqlmodel import SQLModel, select, text
 from sqlmodel.ext.asyncio.session import AsyncSession
 
 from app.api.api_v1.endpoints import login, users
 from app.core.config import settings
 from app.core.security import create_access_token
-from app.db import engine
+from app.core.time import utcnow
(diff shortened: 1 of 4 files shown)


MateoLostanlen:
@Acruve15 can you have a look at this when you have time ?

src/tests/conftest.py
```diff
@@ -262,6 +266,42 @@ async def _drain_detached_notifications():
...
+            return
+        if time.monotonic() > deadline:
+            async with session_factory() as session:
+                res = await session.exec(select(Sequence).where(due_col.is_not(None)))
+                pending = [
+                    (seq.id, seq.validation_due_at, seq.validation_lease_until, seq.validation_attempts)
+                    for seq in res.all()
+                ]
+            # The last claim may have emptied the queue right as the deadline passed:
+            # only fail when jobs actually remain, else let the next claim confirm.
```
Acruve15, line 300: A suggestion, since these tests are fail-open, each sequence clears `validation_due_at` in one `process_next_due_validation()` pass, and a retry backoff needs a mid-drain throw that fail-open never produces. So the flake looks like start-of-drain skew (stale lease or clock step), not something recurring each round, which means normalizing once and running the plain loop should work:

```python
async def drain_validation_queue() -> None:
    async with session_factory() as session:
        await session.exec(
            update(Sequence)
            .where(cast(Any, Sequence.validation_due_at).is_not(None))
            .values(validation_due_at=utcnow() - timedelta(seconds=60), validation_lease_until=None)
        )
        await session.commit()
    while await process_next_due_validation():
        pass
```

Acruve15 (dismissed):

PR #657, opened 2026-07-25
feat(api): Add the admin only endpoint to update the user role

### LINKED ISSUE #656: Add an admin-only endpoint to update a user's role.
# Pitch

Currently, there is no endpoint to update a user's role.

We need an admin-only endpoint that allows changing a user's role between `agent` and `user`. This will enable administrators to promote or demote users as needed.

crocogab:
Closes #656

Adds `PATCH /users/{user_id}/role`, restricted to the `admin` scope, to promote/ demote a user between `agent` and `user`.

Roles are encoded in the access token as scopes, and `JWT_EXPIRE_MINUTES` defaults to one year. A role change therefore only applies to tokens minted afterwards, so the user has to log in again for it to take effect. This is documented in the endpoint docs.

`test_update_user_role` covers 401, 403 (agent and user scopes), 403 (self-demotion), 403 (admin target), 422 (`user_id=0`, `role=admin`, empty body), 404 for nominal transitions and no modif case.

### DIFF
--- a/src/app/api/api_v1/endpoints/users.py
+++ b/src/app/api/api_v1/endpoints/users.py
@@ -12,7 +12,7 @@ from app.core.security import hash_password
 from app.crud import UserCRUD
 from app.models import User, UserRole
 from app.schemas.login import TokenPayload
-from app.schemas.users import Cred, CredHash, UserCreate
+from app.schemas.users import Cred, CredHash, RoleUpdate, UserCreate
 from app.services.telemetry import telemetry_client
 
 router = APIRouter()
@@ -83,6 +83,33 @@ async def update_user_password(
     telemetry_client.capture(token_payload.sub, event="user-pwd", properties={"user_id": user_id})
     pwd = hash_password(payload.password)
     return await users.update(user_id, CredHash(hashed_password=pwd))
+
+
+@router.patch("/{user_id}/role", status_code=status.HTTP_200_OK, summary="Updates a user's role")
+async def update_user_role(
+    payload: RoleUpdate,
+    user_id: int = Path(..., gt=0),
+    users: UserCRUD = Depends(get_user_crud),
+    token_payload: TokenPayload = Security(get_jwt, scopes=[UserRole.ADMIN]),
+) -> User:
+    """Promote or demote a user between the `agent` role and the `user` role.
+    Admins are out of scope: neither the requester nor the target user can have their admin role changed here.
+
+    Beware that the role is baked into the access tokens that were already issued, and those are long-lived
+    (see `JWT_EXPIRE_MINUTES`). The new role only applies to tokens minted afterwards, so the user has to log in
+    again for the change to take effect.
+    """
+    telemetry_client.capture(
+        token_payload.sub, event="user-role", properties={"user_id": user_id, "role": payload.role}
+    )
+    if user_id == token_payload.sub:
+        raise HTTPException(status.HTTP_403_FORBIDDEN, "Admins cannot change their own role : it can lead to deadlock")
+
+    user = cast(User, await users.get(user_id, strict=True))
+    if user.role == UserRole.ADMIN:
+        raise HTTPException(status.HTTP_403_FORBIDDEN, "Cannot change an admin's role")
+
+    return await users.update(user_id, payload)
 
 
 @router.delete("/{user_id}", status_code=status.HTTP_200_OK, summary="Delete a user")
--- a/src/app/api/api_v1/endpoints/users.py
+++ b/src/app/api/api_v1/endpoints/users.py
@@ -95,20 +95,20 @@ async def update_user_role(
     """Promote or demote a user between the `agent` role and the `user` role.
     Admins are out of scope: neither the requester nor the target user can have their admin role changed here.
 
-    Beware that the role is baked into the access tokens that were already issued, and those are long-lived
-    (see `JWT_EXPIRE_MINUTES`). The new role only applies to tokens minted afterwards, so the user has to log in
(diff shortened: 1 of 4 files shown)


src/app/api/api_v1/endpoints/users.py
```diff
@@ -85,6 +85,33 @@ async def update_user_password(
...
+    user_id: int = Path(..., gt=0),
+    users: UserCRUD = Depends(get_user_crud),
+    token_payload: TokenPayload = Security(get_jwt, scopes=[UserRole.ADMIN]),
+) -> User:
+    """Promote or demote a user between the `agent` role and the `user` role.
+    Admins are out of scope: neither the requester nor the target user can have their admin role changed here.
+
+    Beware that the role is baked into the access tokens that were already issued, and those are long-lived
+    (see `JWT_EXPIRE_MINUTES`). The new role only applies to tokens minted afterwards, so the user has to log in
+    again for the change to take effect.
```
fe51, line 100: User tokens use `JWT_UNLIMITED` (365 d), not `JWT_EXPIRE_MINUTES` (60, never used for user tokens).
As written, someone would lower `JWT_EXPIRE_MINUTES` expecting a shorter window and nothing would
happen. (Same slip in the PR description — the "one year" figure is right.)

````suggestion
    Beware that the role is baked into the access tokens that were already issued, and those last a year
    (`JWT_UNLIMITED`, see `login_with_creds`). The new role only applies to tokens minted afterwards, so the
    user has to log in again for the change to take effect.
````

File: src/app/api/api_v1/endpoints/users.py [hunk 2]
```diff
@@ -85,6 +85,33 @@ async def update_user_password(
...
+    """
+    telemetry_client.capture(
+        token_payload.sub, event="user-role", properties={"user_id": user_id, "role": payload.role}
+    )
+    if user_id == token_payload.sub:
+        raise HTTPException(status.HTTP_403_FORBIDDEN, "Admins cannot change their own role : it can lead to deadlock")
+
+    user = cast(User, await users.get(user_id, strict=True))
+    if user.role == UserRole.ADMIN:
+        raise HTTPException(status.HTTP_403_FORBIDDEN, "Cannot change an admin's role")
```
fe51, line 110: A 403/404 currently emits a `user-role` event indistinguishable from a real change. Moving it after the guards fixes that. Also folding in a wording nit ("lockout" rather than "deadlock", and the space before the colon).

````suggestion
    if user_id == token_payload.sub:
        raise HTTPException(status.HTTP_403_FORBIDDEN, "Admins cannot change their own role: it would lock them out")

    user = cast(User, await users.get(user_id, strict=True))
    if user.role == UserRole.ADMIN:
        raise HTTPException(status.HTTP_403_FORBIDDEN, "Cannot change an admin's role")

    telemetry_client.capture(
        token_payload.sub, event="user-role", properties={"user_id": user_id, "role": payload.role}
    )
````

File: src/app/crud/crud_user.py [hunk 3]
```diff
@@ -9,12 +9,12 @@
...
 from app.crud.base import BaseCRUD
 from app.models import User
-from app.schemas.users import CredHash
+from app.schemas.users import CredHash, RoleUpdate
 
 __all__ = ["UserCRUD"]
 
 
-class UserCRUD(BaseCRUD[User, User, CredHash]):
+class UserCRUD(BaseCRUD[User, User, Union[CredHash, RoleUpdate]]):
```
fe51, line 17: juste voc
`Union[CredHash, RoleUpdate]` has no runtime effect (`BaseCRUD.update` only calls `model_dump`) and
will need extending with every new patch schema. `BaseModel` would be more stable. Fine as is too.

fe51 (changes requested):
Hi there, nice to meet you and thanks for you PR. 
Nice once, clean scoping, and the test matrix is thorough. Ran the suite locally, all green.

Good calls: separate `/role` sub-path (no ambiguity with the password PATCH), `admin` excluded at
the schema level so it shows up in the OpenAPI enum, and the self-demotion guard actually works
(`TokenPayload.sub` is an `int`, so no `str`/`int` trap). Cross-org behaviour matches #633, and no
`pyroclient` change is needed since it doesn't wrap `/users`.

Three things to fix, all inline: a wrong setting name in the docstring, an endpoint branch no test
reaches, and telemetry firing before the guards.

On the token staleness: fine to ship as documented — the common case is granting rights, and
"log out / log back in" covers it. But a *demotion* being a no-op for up to a year deserves its own
issue rather than just a docstring.

Beyond the inline suggestions, a few tests that would be nice to have (none blocking):

- **The actual effect on scopes** — promote user 3, `POST /login/creds` as `third_login`, then
  `GET /login/validate` and assert `scopes == ["agent"]`. That's what the issue is really about, and
  it documents the staleness caveat in executable form. Same for the demotion direction.
- `GET /users/{id}` after the PATCH, to check persistence independently of the response body.
- `{"role": "camera"}` → 422, and a camera-scoped token → 403. ?

MateoLostanlen:
Hi @crocogab , thanks for your PR. I’m not sure if you’ve seen Félix’s feedback

Do you think you’ll have time to address it? If not, we can take over from where you left off. Just let us know!

crocogab:
Hi sorry for the late response i had a lot of work this week. Im gonna work on it this week :)))
Thank you for all your feedback. I'll take what you've said into account and come up with a new pr

fe51:
Hi @crocogab, great to hear ! 
You can commit you update in this current PR !

crocogab:
Waiting for your review. I followed your recommendations :33

PR #674, opened 2026-08-04
feat(storage): apply CORS policy on organization S3 buckets at creation (#642)

### LINKED ISSUE #642: Configure CORS on organization S3 buckets so the platform's "download all" buttons work
## Objective

The platform frontend has three image-download features (see `AlertImages.tsx`):

| Mode | Method | Needs CORS? |
|------|--------|-------------|
| 1 — download current image | native `<a download>` | No (browser navigation, no XHR) |
| 2 — download all as ZIP | `fetch(detection.url)` → JSZip | **Yes** |
| 3 — download all + bounding boxes | `fetch(detection.url)` → canvas | **Yes** |

Modes 2 and 3 use native `fetch()` against the presigned S3 URLs. Because these are
cross-origin requests to the object storage host, the browser requires the S3 response to
carry an `Access-Control-Allow-Origin` header. Today our buckets are created with **no CORS
configuration** (`S3Service.create_bucket` only calls `create_bucket`, nothing else), so both
buttons fail.

## Proposed change

1. Add a new config setting `S3_CORS_ORIGINS` (comma-separated → `List[str]`) listing the
   allowed frontend origins (prod, dev, localhost dev). Default `["*"]` for local/dev.
2. Extend `S3Service.create_bucket` to apply a CORS policy via `put_bucket_cors` right after
   the bucket is created:
   - `AllowedMethods: ["GET", "HEAD"]` (browser only reads images; uploads are server-side)
   - `AllowedOrigins: settings.S3_CORS_ORIGINS`
   - `AllowedHeaders: ["*"]`, `ExposeHeaders: ["Content-Length", "Content-Type"]`,
     `MaxAgeSeconds: 3000`
3. Add a test (LocalStack) asserting `get_bucket_cors` returns the expected rules after
   creation.

## ⚠️ Deployed environments -> retroactive backfill required

`create_bucket` only runs at **organization creation**, so this change applies to **new orgs
only**. Existing buckets on the cloud provider will keep having no CORS policy. Once the
mechanism is validated, we must run a **one-off retroactive script** that applies
`put_bucket_cors` to every already-existing bucket. Track this as a follow-up step before
considering the issue done in prod.

## [ ] Open point to confirm 

Most S3-compatible object storage providers support the standard CORS API, but behavior
varies. We should verify with a `get_bucket_cors` read against a real bucket that the policy
is actually stored (and that we are hitting the provider's S3-compatible endpoint, not a
legacy Swift-style one).

fe51:
Closes #642 

## What and Why

Apply a CORS policy on organization S3 buckets **at creation**, so the platform's
"download all (ZIP)" and "download all + bounding boxes" buttons work.

These buttons use native `fetch()` on presigned S3 URLs, which is subject to CORS —
unlike the single-image download, which uses `<a download>` and isn't. With no CORS
config on the buckets, the browser blocked those cross-origin `fetch()`es.

## How

`S3Service.create_bucket` now applies a CORS rule right after creating the bucket:

- `AllowedMethods`: `GET`, `HEAD` (browser only reads images; uploads are server-side)
- `AllowedOrigins`: from the new `S3_CORS_ORIGINS` setting
- `AllowedHeaders`: `*`
- `ExposeHeaders`: `Content-Length`, `Content-Type`
- `MaxAgeSeconds`: `3000`

## Config / default

`S3_CORS_ORIGINS` is a **comma-separated** list of allowed frontend origins.
**Deployments must set it to the real frontend URLs (prod/preprod).** The default
`http://localhost:5173` is a fail-closed fallback for local dev only — deliberately
**not** `*`, so a forgotten env var breaks visibly instead of silently allowing any origin.

## Existing buckets

This only applies to **newly created** buckets. I'll backfill the existing prod/preprod
buckets separately with a one-off script after merge.

## Validation

- Verified against real cloud provider object storage service: `put_bucket_cors` is honored and served
  (confirmed the `Access-Control-Allow-Origin` header is returned and the frontend
  download works end-to-end).
- Added a LocalStack test asserting the CORS policy is set on the bucket at creation.

### DIFF
--- a/src/app/core/config.py
+++ b/src/app/core/config.py
@@ -68,6 +68,11 @@ class Settings(BaseSettings):
     S3_ENDPOINT_URL: str = os.environ["S3_ENDPOINT_URL"]
     S3_PROXY_URL: str = os.environ.get("S3_PROXY_URL", "")
     S3_URL_EXPIRATION: int = int(os.environ.get("S3_URL_EXPIRATION") or 24 * 3600)
+    # Comma-separated browser origins allowed to fetch bucket objects cross-origin (the frontend
+    # platform URLs). Applied as a CORS policy at bucket creation so the frontend can fetch()
+    # presigned image URLs (e.g. the "download all" buttons). Deployments MUST set this to the
+    # real frontend origins; the localhost default only serves local dev.
+    S3_CORS_ORIGINS: str = os.environ.get("S3_CORS_ORIGINS", "http://localhost:5173")
 
     # Sequence handling: three windows gate sequence behaviour, each on its own timescale.
     # All three are tuned against the cameras' frame interval (time between two frames of the
--- a/src/app/services/storage.py
+++ b/src/app/services/storage.py
@@ -143,10 +143,34 @@ class S3Service:
                 else {"CreateBucketConfiguration": {"LocationConstraint": self._s3.meta.region_name}}
             )
             self._s3.create_bucket(Bucket=bucket_name, **config_)
+            self._put_bucket_cors(bucket_name)
             return True
         except ClientError as e:
             logger.warning(e)
             return False
+
+    def _put_bucket_cors(self, bucket_name: str) -> None:
+        """Apply the CORS policy so browsers can fetch() presigned URLs cross-origin.
+
+        Allows the frontend origins (settings.S3_CORS_ORIGINS) to GET bucket objects, which the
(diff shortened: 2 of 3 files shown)


fe51:
@MateoLostanlen all good to review ! test we randomly broken on a part of the code non related to the PR (triangulation). re-running it solve the failing test..

MateoLostanlen (approved):
LGTM. Have you already prepared the backfill script for the existing buckets?

PR #685, opened 2026-08-30
fix(storage): make the bucket CORS policy best-effort

### LINKED ISSUE #683: create_bucket() returns False on MinIO: PutBucketCors is NotImplemented
## Context

Found in the local dev environment ([pyro-envdev](https://github.com/pyronear/pyro-envdev)):
`make run` dies on the init script, which cannot create a single organization. Production is
not affected — only the dev/CI stack, where S3 is MinIO.

```
saving orga : sdis-77
{'detail': 'Failed to create bucket'}
AssertionError
```

## Bug

`POST /organizations/` 500s with `{"detail": "Failed to create bucket"}` against the MinIO
pinned in pyro-envdev (`minio/minio:RELEASE.2025-09-07T16-13-09Z-cpuv1`). The bucket *is*
created — the CORS call after it fails and takes the whole operation down, so the org row is
rolled back while the bucket stays behind.

`S3Service.create_bucket()` (`src/app/services/storage.py` ~L190) wraps both calls in one
`try`, so any `ClientError` returns `False`:

```python
self._s3.create_bucket(Bucket=bucket_name, **config_)
self._put_bucket_cors(bucket_name)     # raises on MinIO
return True
except ClientError as e:
    return False
```

MinIO has never implemented per-bucket CORS (it uses the global `MINIO_API_CORS_ALLOW_ORIGIN`
env var). Reproducer:

```python
s3.create_bucket(Bucket="x")
s3.put_bucket_cors(Bucket="x", CORSConfiguration={
    "CORSRules": [{"AllowedOrigins": ["*"], "AllowedMethods": ["GET"]}]})
# ClientError: (NotImplemented) ... PutBucketCors
```

## Fix

CORS is an enhancement on an already-usable bucket; it shouldn't gate the return value:

```python
    self._s3.create_bucket(Bucket=bucket_name, **config_)
except ClientError as e:
    logger.warning(e)
    return False
try:
    self._put_bucket_cors(bucket_name)   # MinIO answers NotImplemented; best-effort
except ClientError as e:
    logger.warning(f"CORS policy not applied on {bucket_name}: {e}")
return True
```

Production (OVH S3) supports `PutBucketCors`, so nothing changes there. On the pyro-envdev
side, `MINIO_API_CORS_ALLOW_ORIGIN` on the minio service restores the browser-side behaviour
the policy was meant to provide.

Same method, related: MinIO raises `BucketAlreadyOwnedByYou` where AWS us-east-1 silently
succeeds, so re-runs against a store that still holds the bucket also return `False`. Worth
treating `BucketAlreadyOwnedByYou` / `BucketAlreadyExists` as success.

## Wider question: which S3 implementation for dev/CI?

Second time our dev S3 layer constrains the API, both times because a vendor closed down
rather than for technical reasons: LocalStack moved behind an account and a paid tier, and
MinIO, which worked perfectly until now, has been narrowing its community edition (console
stripped in 2025, focus on commercial AIStor), and here we hit a real API gap to code around.

Maybe time to evaluate a genuinely open alternative for the dev environment: Garage (AGPL),
SeaweedFS (Apache-2.0), Zenko CloudServer (Apache-2.0), Ceph RGW. Criteria: passes our S3
tests unchanged, implements `PutBucketCors` or documents an equivalent, single container with
no account/licence key, no feature gating between editions.

The fix above is independent and should land first.

fe51:
Closes #683

## What

`S3Service.create_bucket()` wrapped bucket creation and `put_bucket_cors()` in a single `try`/`except ClientError`. MinIO does not implement the `PutBucketCors` API. it answers `NotImplemented`, which boto3 raises as a `ClientError`  so the CORS failure was caught by the handler that returns `False`.

The result, reported in #683: the bucket **was** created, but the call reported failure, the organization transaction rolled back, and the bucket was left orphaned in storage. The API surfaced a misleading `Failed to create bucket`.

This splits the two concerns:

- a `ClientError` on `create_bucket` is still fatal → `return False`
- a `ClientError` on `put_bucket_cors` is logged as a warning → the bucket is reported as created

This is safe for production as S3 service implements `PutBucketCors`, so the production path is unchanged: bucket created, CORS applied, `True` returned.

## Tests

The existing CORS assertion in `test_s3_service` is untouched and still passes: CI runs against **localstack**, which implements `PutBucketCors`, so the policy is still guarded on a compliant backend.

Added `test_s3_service_create_bucket_survives_an_unsupported_cors_api`, which monkeypatches
`_put_bucket_cors` to raise a `NotImplemented` `ClientError` and asserts the bucket is still
created and usable.

### DIFF
--- a/src/app/services/storage.py
+++ b/src/app/services/storage.py
@@ -198,11 +198,16 @@ class S3Service:
                 else {"CreateBucketConfiguration": {"LocationConstraint": self._s3.meta.region_name}}
             )
             self._s3.create_bucket(Bucket=bucket_name, **config_)
-            self._put_bucket_cors(bucket_name)
-            return True
         except ClientError as e:
             logger.warning(e)
             return False
+        # MinIO answers NotImplemented here, so CORS is best-effort: the bucket is usable
+        # without it, only cross-origin fetch() of presigned URLs degrades.
+        try:
+            self._put_bucket_cors(bucket_name)
+        except ClientError as e:
+            logger.warning(f"CORS policy not applied on {bucket_name}: {e}")
+        return True
 
     def _put_bucket_cors(self, bucket_name: str) -> None:
         """Apply the CORS policy so browsers can fetch() presigned URLs cross-origin.
--- a/src/app/services/storage.py
+++ b/src/app/services/storage.py
@@ -201,12 +201,19 @@ class S3Service:
         except ClientError as e:
             logger.warning(e)
             return False
-        # MinIO answers NotImplemented here, so CORS is best-effort: the bucket is usable
-        # without it, only cross-origin fetch() of presigned URLs degrades.
         try:
             self._put_bucket_cors(bucket_name)
         except ClientError as e:
-            logger.warning(f"CORS policy not applied on {bucket_name}: {e}")
+            if e.response.get("Error", {}).get("Code") != "NotImplemented":
+                # A genuine failure (AccessDenied, malformed S3_CORS_ORIGINS) on a backend that
+                # does support the API: stay loud and fail, so the caller rolls the organization
+                # back instead of silently keeping a bucket the frontend cannot fetch() from.
+                logger.error(f"unable to apply the CORS policy on {bucket_name}: {e}")
+                return False
+            # MinIO does not implement PutBucketCors. On that backend alone CORS is best-effort:
+            # the bucket is usable without it, only cross-origin fetch() of presigned urls
+            # degrades, so creation must not fail.
+            logger.warning(f"CORS policy unsupported by the backend, skipped on {bucket_name}: {e}")
         return True
 
     def _put_bucket_cors(self, bucket_name: str) -> None:
--- a/src/tests/services/test_storage.py
+++ b/src/tests/services/test_storage.py
@@ -68,10 +68,22 @@ async def test_s3_service(region, endpoint_url, access_key, secret_key, proxy_ur
             S3Service(region, endpoint_url, access_key, secret_key, proxy_url)
 
 
+@pytest.mark.parametrize(
+    ("error_code", "expected_created"),
+    [
+        # MinIO does not implement PutBucketCors: the bucket is usable without the policy, so
+        # creation must succeed and only cross-origin fetch() degrades.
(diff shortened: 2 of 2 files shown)


MateoLostanlen:
Hey @fe51, The fix is right: splitting bucket creation from CORS is the correct root-cause fix for [#683](https://github.com/pyronear/pyro-api/issues/683). One blocking point:

The new except ClientError catches everything, not just MinIO's NotImplemented. On a backend that does support CORS, a real failure (AccessDenied, bad S3_CORS_ORIGINS) would be silently swallowed: bucket created without CORS, 201 returned, and logger.warning never reaches Sentry (warnings are breadcrumbs, not events). Since _put_bucket_cors is only called at creation, the bucket never gets repaired and "download all" breaks silently in the browser. Please re-raise unless e.response["Error"]["Code"] == "NotImplemented", or at least log at error level

And one optional, while you're in this handler: transport errors (ReadTimeoutError etc. are BotoCoreError, not ClientError) still escape create_bucket and skip the org rollback in register_organization. Pre-existing, so fine as a separate issue, but except (ClientError, BotoCoreError) on the CORS call would cover it

fe51:
Hi @MateoLostanlen ,

**nice catch on the first point, fixed.**

- Non-`NotImplemented` errors now log at error level and `return False`, so OVH behaviour is unchanged: `register_organization` still deletes the organization and raises its 500. Only MinIO's `NotImplemented` gets a pass.

- I went with `return False` rather than re-raise on purpose: the caller is `if not s3_service.create_bucket(...)`, so a raised `ClientError` would skip the `await organizations.delete(organization.id)` rollback and leave an organization with no bucket.

- Checked against the MinIO running in devenv: `PutBucketCors` answers HTTP 501 with `Error.Code == "NotImplemented"`, so the narrow check matches the real response.


**On the `BotoCoreError` point: agreed in principle, but leaving it out here**. It is pre-existing, and it is not a one-line widening: `BotoCoreError` has no `.response`, so it would raise an `AttributeError` on the error-code check inside this handler. Better as its own issue.

MateoLostanlen (approved):
Thanks for the edit @fe51 !

